Issue #105, opened 2014-02-14
Add Google Analytics


barbeau:
As discussed in https://github.com/OneBusAway/onebusaway-iphone/issues/263#issuecomment-35123537, we'd like to add Google Analytics to both the OBA iOS and Android apps to have better information about how they are being used.  As recommended in [GA Best Practices](https://support.google.com/analytics/answer/2587087), we plan to use the same property for both apps:

>*Track different platforms of an app in the same property.*
If you’ve developed the same app for different platforms (i.e., MyApp for Android and MyApp for iOS), track them in the same property. You can then set up [new views](https://support.google.com/analytics/answer/2614741) to organize and compare the performances in your Google Analytics reports. If your iOS and Android apps differ greatly in performance or usage, you might want to track each platform in separate properties.

Tracking data is needed for research projects such as [StopInfo](https://github.com/OneBusAway/onebusaway-android/issues/103).

barbeau:
@cagryInside is currently working on this.  As mentioned earlier, we’d like to capture a metric for how far the user is in real-time from their bus stop when they check the arrival time.  For example, is the user waiting at the bus stop when they check the time, or from a more distant location (e.g., work, coffee shop)?  Does this differ depending on the stop?

To do this in a privacy-sensitive manner, we’re proposing that we capture distance ranges for each time the user checks the real-time info, without capturing the exact lat/long.

Here’s our current range definitions that would be capture for each arrival time check:
* **NEAR** = User is less than 75m (246ft) from the stop
* **MEDIUM** = User is between 75m (246ft) and 200m (656ft) from the stop
* **FAR** = User is greater than 200m (656ft) from the stop

Stop ID is captured for each arrival time check as well.  So, we’d be able to determine the breakdown of near/middle/far access for each stop in the system, but without capturing precise locations for the access.  We’d also make the assumption that locations with an extremely large accuracy uncertainty value (e.g., > 200m) would automatically get tossed into the “FAR” category (or thrown out completely?), no matter the exact location.

Any comments/suggestions?  Any different opinions for what the range definitions should be, and the number of ranges?  The current choice is somewhat arbitrary, so improvements are welcome.

barbeau:
Above question also posted to oba-dev and users lists:
https://groups.google.com/forum/#!topic/onebusaway-users/bWe1nWm_gwI
https://groups.google.com/forum/#!topic/onebusaway-developers/bWe1nWm_gwI

bbodenmiller:
[...]

PR #144, opened 2014-06-18
wip: add google analytics

bbodenmiller:
Initial attempt at implementing Google Analytics as mentioned at https://github.com/OneBusAway/onebusaway-android/issues/105. I haven't been able to work on this as much as I'd like to and I'd still like to compare this more with the iOS implementation to make sure they have similar events reported. Despite some work still being required it'd be great if I could get some feedback on the progress so far.

Implements [Google Analytics SDK v4 for Android](https://developers.google.com/analytics/devguides/collection/android/v4/) via the [Google Play Services SDK](https://developer.android.com/google/play-services/index.html). Bumps min SDK from 8 to 9 since GA was added in Google Play services, Version 4.3 which does not support SDK 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,15 @@ env:
 
 android:
   components:
+    - platform-tools
     - build-tools-19.1.0
-    # For Google Maps API v1
+    - android-$ANDROID_API_LEVEL
+    - sysimg-$ANDROID_API_LEVEL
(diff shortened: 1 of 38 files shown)


barbeau:
Closing this in favor of updated work in https://github.com/OneBusAway/onebusaway-android/pull/207

PR #197, opened 2014-12-04
WIP - Stub out core module

pohh:
This request is to stub out the core functionality required by an android module which would be shared between the various deployed apps.

At present this set is limited to the wear module (to be completed in another pull), and the app.

The workflow I was thinking was as core functionality is required by features in the wear app, they would be moved from the current ```onebusaway-android``` module into the ```core``` module to keep the core module as lightweight as possible.


barbeau:
This PR was automatically closed by Github due to the deletion of the `develop` branch - we're moving back to using `master` as the main branch.

@pohh Is this still something you're interested in working on?  Just trying to get a handle of how this PR might look going forward - whether I need to pull into a branch to preserve the initial work for someone else to look at, or whether you want to rebase on `master` and continue work.

PR #212, opened 2015-02-23
Fix #201 - Add a donate link in preferences

### LINKED ISSUE #201: Create a "Donate" link in the mobile app
This would be for donations to the overall OBA project (not regions), and would direct to this link:
http://onebusaway.org/donate/

Currently the donation account is managed via UW, and used for overall project expenses.

Also ticketed for iOS in https://github.com/OneBusAway/onebusaway-iphone/issues/349.

barbeau:
* Renames "Analytics" preference category to "About", and groups both analytics opt-out and donate preference under this

This adds the link to the master branch.  I'll merge this into develop as well.

### DIFF
--- a/onebusaway-android/src/main/res/values/do_not_translate.xml
+++ b/onebusaway-android/src/main/res/values/do_not_translate.xml
@@ -25,8 +25,11 @@
     <string name="preference_key_auto_select_region">preference_auto_select_region</string>
     <string name="preference_key_experimental_regions">preference_experimental_regions</string>
     <string name="preferences_key_analytics">preference_google_analytics</string>
+    <string name="preferences_key_donate">preference_donate</string>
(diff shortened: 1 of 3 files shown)


barbeau:
@cagryInside Could you please give this a quick review?  Thanks!

cagryInside:
@barbeau I think looks good.

barbeau:
Thanks!

bbodenmiller:
@barbeau what is your thoughts on doing analytics on the number of people that decide to opt out of analytics? I think it is justifiable to report when someone opts-out so we have an idea of how many users are missing.

barbeau:
@bbodenmiller Agreed, it seems reasonable to detect the opt-out itself.  This way we'll get some idea how representative the GA data is, as a percentage of all users.

@cagryInside I'll change the other ticket to include both of these.

PR #216, opened 2015-02-24
Fix  #213 - Google analytics implementations for donate link and opt-out...

### LINKED ISSUE #213: More Google Analytics - donate link & opt-out
Just realized that we should be tracking taps on "Donate" preference.

@cagryInside Would you be able to put this together tomorrow?

My goal is to get a new beta based on the master branch that includes GA out by the end of this week.

cagryInside:
### NOTE:
Google analytics' ```setAppOptOut()``` method removed for enabling/disabling the GA in the app. Because, when we send event before disabling the GA, the event doesn't show up in the GA admin. This issues also posted to stackoverflow:
http://stackoverflow.com/q/28688695/701325
http://stackoverflow.com/q/28701628/701325

Instead of using ```setAppOptOut()```, a workaround solution implemented. Simply before pushing any log to GA, we check if the GA is active in the app.

### DIFF
--- a/onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
+++ b/onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
@@ -103,22 +103,24 @@ public class ObaAnalytics {
      * @param label    label name
      */
     public static void reportEventWithCategory(String category, String action, String label) {
-        Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
-        Tracker tracker2 = Application.get().getTracker(Application.TrackerName.GLOBAL_TRACKER);
-        String obaRegionName = getObaRegionName();
+        if (isAnalyticsActive()) {
+            Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
+            Tracker tracker2 = Application.get().getTracker(Application.TrackerName.GLOBAL_TRACKER);
+            String obaRegionName = getObaRegionName();
 
-        tracker.send(new HitBuilders.EventBuilder()
-                .setCategory(category)
-                .setAction(action)
-                .setLabel(label)
-                .setCustomDimension(1, obaRegionName)
-                .build());
-        tracker2.send(new HitBuilders.EventBuilder()
-                .setCategory(category)
-                .setAction(action)
-                .setLabel(label)
-                .setCustomDimension(1, obaRegionName)
-                .build());
+            tracker.send(new HitBuilders.EventBuilder()
+                    .setCategory(category)
+                    .setAction(action)
+                    .setLabel(label)
+                    .setCustomDimension(1, obaRegionName)
+                    .build());
+            tracker2.send(new HitBuilders.EventBuilder()
+                    .setCategory(category)
+                    .setAction(action)
+                    .setLabel(label)
+                    .setCustomDimension(1, obaRegionName)
+                    .build());
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
```diff
@@ -103,6 +103,35 @@ public String toString() {
...
+
+    /**
+     * Reports events with categories even when GA is disabled.
+     * Helps segmentation in GA admin console.
+     *
+     * @param category category name
+     * @param action   action name
+     * @param label    label name
+     */
+    public static void forceReportEventWithCategory(String category, String action, String label) {
```
barbeau: I'd prefer not to expose a public static method that allows us to bypass the user's preference for not reporting analytics.  It's too easy for someone else who isn't as familiar with the code base to come along and accidentally start sending data via this method without checking the preference.  I'd prefer to have a private method in PreferencesActivity handle this, and to have it well commented so others know that the only time this should be used is when we're disabling GA reporting, and reporting that disabling event via GA.  See also my comment in PreferencesActivity.

onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/PreferencesActivity.java
```diff
@@ -149,7 +162,10 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             }
         } else if (preference.equals(analyticsPref) && newValue instanceof Boolean) {
             Boolean isAnalyticsActive = (Boolean) newValue;
```
barbeau: The SharedPreference value for analyticsPref doesn't persist until `onPreferenceChange()` returns (see [Android docs](http://developer.android.com/reference/android/preference/Preference.OnPreferenceChangeListener.html), so you should be ok doing a synchronous call to ObaAnalytics.reportEventWithCategory() without having to do `forceReportEventWithCategory()`.  See also my comment on `forceReportEventWithCategory()`.  Please test to double-check the Android docs are correct ;).

barbeau:
Thanks @cagryInside!  A few requests in-line.

cagryInside:
@barbeau Thanks for pointing out. I made the changes. 
I used ```onPreferenceChange()``` for reporting GA turning off event. Because in this method, GA is not turned off yet.

```onSharedPreferenceChanged()``` used for reporting GA turning on event.  Because in this method, GA is just turned on, so we can send the event.

File: onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/PreferencesActivity.java [hunk 3]
```diff
@@ -149,7 +161,12 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
-            GoogleAnalytics.getInstance(this).setAppOptOut(!isAnalyticsActive);
+            //Report if the analytics turns off, just before shared preference changed
+            if (!isAnalyticsActive)
+                ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.APP_SETTINGS.toString(),
+                        getString(R.string.analytics_action_edit_general),
+                        getString(R.string.analytics_label_analytic_preference)
+                                + (isAnalyticsActive ? "YES" : "NO"));
```
barbeau: @cagryInside Could you also please add `dispatchLocalHits()` after reporting this event?  I'd like to flush any existing analytics info here, so we don't have analytics continuing to report (e.g., in 30 min) after the user has opted-out.  As we found, I guess this will only work on devices without Google Play Services, but at least its our best effort to be consistent with the user's wishes.  And, if Google Play Services decides to support this in the future, we're already set.

cagryInside: @barbeau I thought the same, since ```dispatchLocalHits()``` doesn't work with google play services. But as you mention, It's nice to have for consistency. I will update shortly.

cagryInside:
@barbeau Added the ```dispatchLocalHits```, fyi

barbeau:
@cagryInside Thanks!  I think that should do it.  I'll wait for the Travis build to finish, and then will merge.

Could you also please do a port of this to the develop branch, including no longer using `setAppOptOut()`?

barbeau:
If you could include the new "Donote" preference too, that would be great.  I haven't ported that to develop yet.

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
```diff
@@ -129,87 +131,88 @@ public static void reportEventWithCategory(String category, String action, Strin
...
+                } else if (distanceInMeters < ObaStopDistance.DISTANCE_7.getDistanceInMeters()) {
+                    stopDistance = ObaStopDistance.DISTANCE_7;
+                } else {
+                    stopDistance = ObaStopDistance.DISTANCE_8;
+                }
+
+                Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
+                Tracker tracker2 = Application.get().getTracker(Application.TrackerName.GLOBAL_TRACKER);
+
+                tracker.send(new HitBuilders.EventBuilder()
```
bbodenmiller: Wouldn't it be better to just use `reportEventWithCategory` here?

cagryInside:
@barbeau Sure, I will do that.

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
```diff
@@ -129,87 +131,88 @@ public static void reportEventWithCategory(String category, String action, Strin
...
      */
     public static void reportActivityStart(Activity activity) {
-        Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
-        tracker.setScreenName(activity.getClass().getSimpleName());
-        tracker.send(new HitBuilders.ScreenViewBuilder().build());
-        Tracker tracker2 = Application.get().getTracker(Application.TrackerName.GLOBAL_TRACKER);
-        tracker2.setScreenName(activity.getClass().getSimpleName());
-        tracker2.send(new HitBuilders.ScreenViewBuilder().build());
+        if (isAnalyticsActive()) {
+            Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
```
bbodenmiller: Wouldn't it be better to have this as a function of `ObaAnalytics`?

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
```diff
@@ -129,87 +131,88 @@ public static void reportEventWithCategory(String category, String action, Strin
...
      */
     public static void reportFragmentStart(Fragment fragment) {
-        Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
-        tracker.setScreenName(fragment.getClass().getSimpleName());
-        tracker.send(new HitBuilders.ScreenViewBuilder().build());
-        Tracker tracker2 = Application.get().getTracker(Application.TrackerName.GLOBAL_TRACKER);
-        tracker2.setScreenName(fragment.getClass().getSimpleName());
-        tracker2.send(new HitBuilders.ScreenViewBuilder().build());
+        if (isAnalyticsActive()) {
+            Tracker tracker = Application.get().getTracker(Application.TrackerName.APP_TRACKER);
```
bbodenmiller: Wouldn't it be better to have this section as a function of `ObaAnalytics`?

bbodenmiller:
In general I also think it looks great just a couple comments about code organization not directly related to this PRs purpose. Additionally have you guys figured out how to make sure any existing analytics data is sent right after they disable analytics rather than waiting for GAs built in dispatch rate to hit?

bbodenmiller:
I see flushing GA is not easy. Is that also why `setAppOptOut` is not used? Assume it prevents the analytics opt out event from being sent?

barbeau:
@bbodenmiller Right, apparently when Google Play Services is on the device (and therefore manages Google Analytics), you can't force flush events to Google's servers.  So, when you call `setAppOptOut`, no additional events are sent (including the event for opting out).  We're no longer using this, and just checking our own boolean variable for whether analytics is active before triggering events.

barbeau:
For posterity, here's the section in the [Google Analytics docs](https://developer.android.com/reference/com/google/android/gms/analytics/GoogleAnalytics.html#dispatchLocalHits()) that explains this:

>`public void dispatchLocalHits ()`

>Dispatches queued hits (view, events, or transactions) to Google Analytics if a network connection is available, and the local dispatching service is in use. This method only works if local dispatching is in use. Local dispatching is only used in the absence of Google Play services on the device. In general, applications should not rely on the ability to dispatch hits manually.

We weren't able to find an equivalent method to dispatch hits when using Google Analytics.

cagryInside:
@bbodenmiller I also opened two threads in stackoverflow about this:
http://stackoverflow.com/q/28688695/701325
http://stackoverflow.com/q/28701628/701325

bbodenmiller:
Sounds good. My only concern with that approach is if a developer doesn't use the functions in `ObaAnalytics` (that properly check the internal setting) events could be sent when they shouldn't be. You'd think `setAppOptOut()` would be intelligent enough to just not send events recorded after it is enabled.

barbeau:
>My only concern with that approach is if a developer doesn't use the functions in ObaAnalytics (that properly check the internal setting) events could be sent when they shouldn't be. 

Right, that's a limitation to the workaround.  We've tried to structure the code so this isn't easily possible when using the current ObaAnalytics class.  We can also change if we find a better workaround in the future.

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
cagryInside: @bbodenmiller I don't really understand what you mean by ```have this as a function of ObaAnalytics```. But if you mean that we should implement ```getTracker(Application.TrackerName.APP_TRACKER)``` in the ```ObaAnalytics``` class:

In the Google analytics documentation, they recommend to manage these functions and trackers in the ```Application``` class. 

````
With the new SDK, developers should manage the trackers themselves. To ensure that the
metrics are not over-counted, it is highly recommended that the tracker be created and 
managed in the Application class.
`````

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
cagryInside: @bbodenmiller Yes I agree, I would be better approach to reduce redundancy.
I will update the pr. @barbeau fyi,

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
bbodenmiller: My bad, you already have this implemented in `ObaAnalytics`... That is what I get for reviewing a PR from a phone.

onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java
bbodenmiller: My bad, you already have this implemented in `ObaAnalytics`.

barbeau:
@cagryInside I see you pushed another commit.  So this should be ready to go now?

cagryInside:
@barbeau Yes, i just add little enhancement:

Used ```reportEventWithCategory()``` in ```trackBusStopDistance``` method.

barbeau:
Awesome, thanks!

Issue #181, opened 2014-10-06
Add Android Wear support


barbeau:
Requested by a user.

For adding support for existing notifications, see:
https://developer.android.com/training/wearables/notifications/index.html

brendandburns:
Was any work done on this?  If not I'd love to pick it up as something to explore.

Any pointers you may have would be welcome.

Thanks!
--brendan

barbeau:
That would be great! @pohh got a proof of concept working - see https://github.com/OneBusAway/onebusaway-android/pull/198,  and https://github.com/OneBusAway/onebusaway-android/pull/197 stubs out a breakdown of the project into  modules to help support this .  The PRs were closed during the merge process of some of the UI work back to the master branch,  so the first order of business would probably be seeing how far this work has diverged from current master branch and possibly rehashing. 

@pohh if you have any wisdom based on your work feel free to chime in.

pohh:
The wear app was functional but after looking at the resulting wear app i believe it was overkill.

Implementing voice search intents and rich notifications should get you a lot of oba functionality on the wear device without needing a wear app.  The previous attempt was a littlw heavyweight.  Id consider rich notifications and voice search intents first.

Ive been meaning to get back to this but started a new job shortly after conpleting first slice of the pull request and havent had a chance to finish up.  Glad someone else is interested in finishing this up!  Im out of country at the moment but if you have any other qs i can do my best to help.

Issue #116, opened 2014-04-15
OneBusAway Glassware - native GDK version


barbeau:
I've been working on a GDK (native) version of OBA Glassware lately, and I now have a working alpha version to share.  I'm opening this issue so we can track implementation for the GDK version separately from the Mirror API version (https://github.com/OneBusAway/onebusaway-application-modules/issues/60).

Code for this is now available via a WIP PR - https://github.com/OneBusAway/onebusaway-android/pull/219.

Options to install:
1. Download the current APK [here](https://dl.dropboxusercontent.com/u/46443835/OneBusAway/Glass/onebusaway-glass-debug-unaligned.apk), and follow [these directions](http://appliedanalog.com/agw/?p=17) to install it on Glass, or...
2. If you have the Android dev environment setup w/ GDK installed - run `gradlew installDebug` from root of project using command-line

OBA Glassware is activated by using voice commands "ok glass", then "show me transit times" (_note: this was changed from the original voice command per direction of Glass Review Team_):

![show_me_transit_times](https://cloud.githubusercontent.com/assets/928045/2746026/01192f70-c73f-11e3-9816-48ce3642a7d8.png)

It will figure out your region (Puget Sound, Tampa, Atlanta, Washington DC, or York Canada):
![finding_your_transit_services](https://cloud.githubusercontent.com/assets/928045/2712310/e9a90800-c4d8-11e3-8af6-e73ff36aa6aa.png)

...find the closest bus stop to your real-time location, and then show you the arrival times for that stop:
![arrivals_uatc6-arrow](https://cloud.githubusercontent.com/assets/928045/2882709/dd884c34-d491-11e3-8be0-47a3cfad728d.png)

Times are refreshed every minute until you dismiss the Activity by swiping down.  You can scroll up/down the list of arrival times by tilting your head up and down.  You can pause/resume scrolling the list by long pressing with two fingers on the touch pad.  Tapping with two fingers will automatically scroll to the top of the list.  The arrow points towards your stop from your current position and orientation (so in the example above, you'd turn to your right a little over 90 degrees to face your stop).

[...]

PR #219, opened 2015-02-24
WIP - Issue #116 - OBA Google Glass GDK

barbeau:
This is somewhat dated now (most recent work mid-2014), but I wanted to open this as a WIP so it doesn't get lost, and is more visible if someone else wants to contribute to the Glass work.  

Options to install an alpha version (this PR contents):
1. Download the current APK [here](https://dl.dropboxusercontent.com/u/46443835/OneBusAway/Glass/onebusaway-glass-debug-unaligned.apk), and follow [these directions](http://appliedanalog.com/agw/?p=17) to install it on Glass, or...
2. If you have the Android dev environment setup w/ GDK installed - run `gradlew installDebug` from root of project using command-line

OBA Glassware is activated by using voice commands "ok glass", then "show me transit times".

show_me_transit_times

It will figure out your region (Puget Sound, Tampa, Atlanta, Washington DC, or York Canada):
finding_your_transit_services

...find the closest bus stop to your real-time location, and then show you the arrival times for that stop:
arrivals_uatc6-arrow

Times are refreshed every minute until you dismiss the Activity by swiping down.  You can scroll up/down the list of arrival times by tilting your head up and down.  You can pause/resume scrolling the list by long pressing with two fingers on the touch pad.  Tapping with two fingers will automatically scroll to the top of the list.  The arrow points towards your stop from your current position and orientation (so in the example above, you'd turn to your right a little over 90 degrees to face your stop).

Long-term goal is to get the GDK version of OBA Android merged back into the main project, hopefully using Gradle build flavors and/or project modules (see also work related to Wear in https://github.com/OneBusAway/onebusaway-android/issues/181).  Main differences between the main OBA app project and this PR is that Glass does not support the Google APIs add-on (including maps and `GeoPoint` object) or Google Play Services, and apparently has poor support for Fragments (hence use of a `ListActivity` instead of `ListFragment`).  Most of the internal infrastructure for API calls and various utilities are the same, though, and could be shared between projects.

I'd like to get feedback from others for what features they'd like to see, so please let me know what features you think would be useful.

See https://github.com/OneBusAway/onebusaway-android/issues/116 for more details.


glass-progress-bar/build.gradle
```diff
@@ -0,0 +1,40 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:0.9.+'
```
cagryInside: @barbeau Should we use gradle version ```1.0.1``` as same with OBA (https://github.com/CUTR-at-USF/onebusaway-android/blob/glass/onebusaway-android/build.gradle)?

onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/GlassArrivalsListActivity.java
```diff
@@ -0,0 +1,719 @@
...
+    public void onTaskFinished(ObaStopsForLocationResponse response) {
+        if (response.getCode() != ObaApi.OBA_OK) {
+            showErrorMessage(getString(UIHelp.getStopErrorString(this, response.getCode())));
+            return;
+        }
+
+        Log.d(TAG, "Found stops.");
+        // Find closest stop
+        ObaStop closestStop;
+        closestStop = UIHelp.getClosestStop(this, response.getStops(), mLastKnownLocation);
```
cagryInside: @barbeau I would use these together: 
```
ObaStop closestStop = UIHelp.getClosestStop(this, response.getStops(), mLastKnownLocation);
````

glass-progress-bar/build.gradle
barbeau: @cagryInside Yes, this (and the Gradle wrapper) likely need to be updated before this will build in the most recent version of Android Studio.

onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/GlassArrivalsListActivity.java
barbeau: Agreed!  Code in this PR isn't my best work :).  It's more of a shortest path to get something working on Glass.

glass-progress-bar/build.gradle
barbeau: I'll push an update for this shortly.

barbeau:
Thanks for the feedback @cagryInside, just pushed a few commits to fix those items.

aaronbrethorst:
Closing. Google Glass has been discontinued.

PR #221, opened 2015-02-25
Fix #220 - Always show "What's New" dialog when upgrading app

### LINKED ISSUE #19: Reminders don't restart after an update
What steps will reproduce the problem?
1. Schedule a reminder
2. Update the app 
3. Wait for the reminder

What is the expected output? What do you see instead?

Android doesn't seem to be restarting the TripService, and so the reminder
alarms are removed. The user has to go and re-save a reminder, add a new
reminder, or reboot their phone.

Figure out a way for Android to run the server post-update, or always run
the service when the user runs the app.

### LINKED ISSUE #220: Auto-show "What's New" dialog is broken
I actually didn't realize that this was supposed to be happening until I was looking at implementing the same feature.  It appears that its not working on new installs because the default version is `0`, and there is an IF statement in HomeActivity checking for `oldVer > 0` before notifying the user.

I'll submit a fix for this shortly.

barbeau:
* This may also Fix #19, since the alarm rescheduling for reminders happens in this same IF block

### DIFF
--- a/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/HomeActivity.java
+++ b/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/HomeActivity.java
@@ -15,6 +15,19 @@
  */
 package com.joulespersecond.seattlebusbot;
 
+import com.actionbarsherlock.view.Menu;
+import com.actionbarsherlock.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


barbeau:
Looks like Travis is hung up - https://twitter.com/traviscistatus

barbeau:
...and Travis is back.

PR #236, opened 2015-02-27
Fix  #233 - StopInfo event replaced with Stop icon click event

### LINKED ISSUE #233: StopInfo event reported for incorrect button press in master branch
Right now, when the user taps on the icon the map, we're reporting a "Loaded StopInfo from X", where X is the region name.  This event on iOS is actually related to the StopInfo project, the icon for which is shown next to the stop name on the ArrivalList Fragment, as discussed in https://github.com/OneBusAway/onebusaway-android/issues/103.

We should remove the "Loaded StopInfo from X" event from the current master branch, and replace it with a new event (not currently reported on iOS) called "Clicked MapStopIcon from X".

We should then report a similar map icon click event on iOS, to be opened in a new issue on OBA iOS.

cagryInside:
StopInfo event was reporting incorect button press. StopInfo event replaced with button click event


barbeau:
Thanks @cagryInside!

PR #239, opened 2015-03-03
add IRC badge

bbodenmiller:
I like Gitter better but it does appear people use the IRC channel.


barbeau:
@bbodenmiller Thanks!  I'm not an active user on that IRC channel, though, so I'd prefer not to put the badge on the readme if I'm not planning on being active there.  I'd like to stick with Gitter for now and see how it works.  If it doesn't, I'll probably revert to IRC channel.

PR #241, opened 2015-03-04
Fix #237 - If custom url reporting disabled when it is null.

### LINKED ISSUE #237: Including custom URLs when reporting events with GA.
We start reporting SHA-1 hashed custom URLs as a secondary dimension as discussed https://github.com/OneBusAway/onebusaway-android/issues/105#issuecomment-71099932.  However, when the application first time installed and run in devices, it also reports "CUSTOM URL" event, even though there is no custom URL set.

cagryInside:
* Api region reporting removed from MapStopIcon event as discussed in https://github.com/OneBusAway/onebusaway-android/issues/233#issuecomment-76745031.
* Agreed to keep OBA Android consistent with OBA Iphone as discussed in https://github.com/OneBusAway/onebusaway-android/issues/237#issuecomment-77021844. So, the following differences added:
    - Turned on Experimental Regions
    - Turned off Experimental Regions
    - Set region manually


onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/PreferencesActivity.java
```diff
@@ -216,9 +216,14 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
 
             // Wait to change the region preference description until the task callback
             //Analytics
-            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
-                    getString(R.string.analytics_action_button_press),
-                    getString(R.string.analytics_label_button_press_experimental) + experimentalServers);
+            if (experimentalServers)
```
barbeau: @cagryInside Could you please add braces`{` to this IF statement?  As a matter of style I prefer to always use them.  Also, you might want to check if you have the "reformat code" checkbox checked in the "Commit Changes" view in Android Studio.  I think the style template should have automatically reformatted this to include braces.

onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/PreferencesActivity.java
```diff
@@ -229,9 +234,14 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
             //Analytics
             boolean autoSelect = settings
                     .getBoolean(getString(R.string.preference_key_auto_select_region), false);
-            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
-                    getString(R.string.analytics_action_button_press),
-                    getString(R.string.analytics_label_button_press_auto) + autoSelect);
+            if (autoSelect)
```
barbeau: Same here - please add braces `{`

barbeau:
Thanks @cagryInside!  Mostly looks good - one request in-line.

cagryInside:
@barbeau thanks. I will make the same changes in develop branch too.

barbeau:
Thanks!

barbeau:
@cagryInside I think this should be good to be merged now?

cagryInside:
@barbeau agreed. Thanks.

barbeau:
Great, thanks @cagryInside!

aaronbrethorst:
Fixed in #1102.

PR #242, opened 2015-03-04
Fix #234 report stop info, Fix # 237 custom url reporting disabled - develop branch

### LINKED ISSUE #232: Change location accuracy threshold for Google Analytics
@cagryInside I was looking at some of the location data from a few Android phones, and I'm thinking it probably makes sense to tighten up the horizontal accuracy threshold prior to a production release.

Currently, [we're using 100m](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java#L43) as the cutoff threshold - meaning that we report analytics for distance to stop if the estimated error radius of the device location is less than 100m, and throw it away if its >= 100m.

I'm thinking it may be better to change this to 50m, so we have a tigher bound/accuracy on the analyitics data.  Looking at a few devices, most of them report accuracy < 50m when its reasonable.  If its more than 50m, it starts getting questionable, especially for our [smallest radii](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/com/joulespersecond/oba/ObaAnalytics.java#L51), which is defined at 50m separation.

Your thoughts?  Any data from your devices?

Good apps to test with are [GPS Benchmark](https://play.google.com/store/apps/details?id=com.gpsbenchmark.android) and [GPSTest](https://play.google.com/store/apps/details?id=com.android.gpstest) (yes, they are both my apps ;) ).

### LINKED ISSUE #234: Report StopInfo event (develop)
We should report the "Loaded StopInfo from X" event, where X is the region name, when the user taps on the icon for which is shown next to the stop name on the ArrivalList Fragment, as discussed in #103.

This was originally incorrectly reported for the event of tapping on the map in the master branch, as discussed in #233.

### LINKED ISSUE #237: Including custom URLs when reporting events with GA.

### LINKED ISSUE #244: Add Google Analytics individually for trip/stop problem reports
In the current Google Analytics implementation, we're reporting a generic "problem report" metric in `ReportProblemFragmentBase`:
https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ReportProblemFragmentBase.java#L117

However, as far as I can tell this won't provide any separate reporting for trip/stop reports, and I'd like to be able to record what reports of each type are submitted.

So, this analytics event should probably be moved to the `sendReport()` method of the subclasses [`ReportStopProblemFragment`](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ReportStopProblemFragment.java#L111) and [`ReportTripProblemFragment`](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ReportTripProblemFragment.java#L157), and a separate label should be included (or another way to distinguish the report types).

@cagryInside Could you please confirm this is correct, and if so submit a quick fix for this?  We should fix this before pushing the next beta based on master, and would need to fix in develop too.

cagryInside:
* Added map icon click event as discussed https://github.com/OneBusAway/onebusaway-android/issues/234#issue-59284031
* Fix #237 custom url reporting disabled
* Fix #244 - specific trip/stop problem reports
* Fix #232 - location accuracy for analytics
* Agreed to keep OBA Android consistent with OBA Iphone as discussed in https://github.com/OneBusAway/onebusaway-android/issues/237#issuecomment-77021844. So, the following differences added:
     * Turned on Experimental Regions
     * Turned off Experimental Regions
     * Set region manually


onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
```diff
@@ -214,11 +214,15 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
             task.execute();
 
             // Wait to change the region preference description until the task callback
-
             //Analytics
-            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
-                    getString(R.string.analytics_action_button_press),
-                    getString(R.string.analytics_label_button_press_experimental) + experimentalServers);
+            if (experimentalServers)
```
barbeau: Same as other PR - please add `{`

onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
```diff
@@ -229,10 +233,14 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
             //Analytics
             boolean autoSelect = settings
                     .getBoolean(getString(R.string.preference_key_auto_select_region), false);
-
-            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
-                    getString(R.string.analytics_action_button_press),
-                    getString(R.string.analytics_label_button_press_auto) + autoSelect);
+            if (autoSelect)
```
barbeau: Same as other PR - please add `{`

barbeau:
@cagryInside are you still rebasing this?  Two commits ended up in this PR.

cagryInside:
@barbeau I finish rebasing. I would like to merge two different commits if it is possible. One for custom url reporting , another for location accuracy issue. I thought, it would be better to have these issues in two different commits, since they are not related to each other.

cagryInside:
@barbeau Also, Fix https://github.com/OneBusAway/onebusaway-android/issues/244 added for develop branch

barbeau:
Thanks @cagryInside!

PR #245, opened 2015-03-05
Fix #244 - Add Google Analytics individually for trip/stop problem reports

### LINKED ISSUE #244: Add Google Analytics individually for trip/stop problem reports

cagryInside:
Also, fixed small problem:
```Clicked Send Feedback Link``` changed to ```Clicked Email Link```.

### DIFF
--- a/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ReportStopProblemFragment.java
+++ b/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ReportStopProblemFragment.java
@@ -28,6 +28,7 @@ import android.widget.Spinner;
 import android.widget.TextView;
 
 import com.actionbarsherlock.app.SherlockFragmentActivity;
+import com.joulespersecond.oba.ObaAnalytics;
 import com.joulespersecond.oba.elements.ObaStop;
 import com.joulespersecond.oba.request.ObaReportProblemWithStopRequest;
 import com.joulespersecond.seattlebusbot.util.LocationHelp;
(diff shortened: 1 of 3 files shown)


barbeau:
@cagryInside Looks good, thanks!  So `Clicked Email Link` matches iOS?

cagryInside:
> So Clicked Email Link matches iOS?

@barbeau Yes.

barbeau:
Great, thanks!

PR #246, opened 2015-03-09
Fix  #243 - Workaround solution for GoogleAnalytics.setDryRun(true) issue in Amazon Fire phones

### LINKED ISSUE #243: GoogleAnalytics.setDryRun(true) doesn't work on Amazon Fire Phone (develop)
We've observed a number of Google Analytics (GA) events reported from Amazon Fire Phones, which is strange because:

1. if `BuildConfig.DEBUG == TRUE` we set `GoogleAnalytics.getInstance(context).setDryRun(true);`, which should disable event reporting on any non-production signed build
2. The currently deployed production-signed beta that includes analytics does not run on a Fire Phone (currently only the develop branch supports the amazon build flavor)
3. There haven't been any production-signed releases of the amazon build flavor on the develop branch

Upon further inspection, it appears that `GoogleAnalytics.getInstance(context).setDryRun(true);` does not have any affect devices without Google Play Services (including the Fire Phone).  In other words, this is likely a bug in the Google Analytics SDK with local dispatching.  [The GA documentation](https://productforums.google.com/forum/#!topic/analytics/1zAmZCu1Bx4) does not state anything about this method not working on non-Google Play Services devices, so we assume it should work on devices like the Fire Phone:

>**public void setDryRun (boolean dryRun)**
>Toggles dry run mode. In dry run mode, the normal code paths are executed locally, but hits are not sent to Google Analytics servers. This is useful for debugging calls to the Google Analytics SDK without polluting recorded data.

Related StackOverflow post:
http://stackoverflow.com/questions/28842499/google-analytics-setdryruntrue-doesnt-work-in-amazon-firephone
Related post on Google GA forum:
https://productforums.google.com/forum/#!topic/analytics/1zAmZCu1Bx4

cagryInside:
GoogleAnalytics.setDryRun(true) doesn't work on Amazon Fire Phones. 

>public void setDryRun (boolean dryRun)
Toggles dry run mode. In dry run mode, the normal code paths are executed locally, but hits are not sent to Google Analytics servers. This is useful for debugging calls to the Google Analytics SDK without polluting recorded data.

Instead of ```setDryRun```, we implemented ```GoogleAnalytics.getInstance(context).setAppOptOut(true)``` method, which disables Google analytics reporting on Amazon fire phone devices.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/io/ObaAnalytics.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/io/ObaAnalytics.java
@@ -46,6 +46,11 @@ public class ObaAnalytics {
     private static final float LOCATION_ACCURACY_THRESHOLD = 50f;
 
     /**
+     * Amazon devices manufacturer name
+     */
+    public static final String mAmazonManufacturer = "amazon";
+
+    /**
      * To measure the distance when the bus stop tapped.
      */
     public enum ObaStopDistance {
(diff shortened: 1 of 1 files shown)


barbeau:
Thanks @cagryInside!

Issue #187, opened 2014-10-17
Improve issue reporting


cagryInside:
Summary:  To provide some background – in Tampa, HART has gotten a large amount of feedback from transit riders via OneBusAway.  However, they lack good infrastructure to triage/manage these issues, both because of the multi-departmental administration of OBA, as well as issues that cross jurisdictions and may not relate to HART infrastructure.  Therefore, cross-departmental and cross-agency communication on issues is very important.
 
Our group at USF is looking to improve this process with better support for open issue reporting and tools to triage/manage issues.  We are also working to on-board another regional agency, PSTA that serves Pinellas County, into OneBusAway Tampa.
 
Our current plan is to:
 
1.       Improve the current help/feedback process within the app – Right now, there are several channels within the app to seek help/report problems.  A) Users can report trip/stop problem reports within the arrivals screen.  However, it is not easily discoverable.  B) Users can send an email C) A Twitter account for the region is accessible.  We’d like to better organize/present this within the app so we direct the user to the best channel to submit their feedback based on the type of feedback.
2.       Direct infrastructure/agency-specific feedback to dedicated issue reporting sites – HART has indicated that it is difficult for them to process email and stop problem reports within the OBA webapp.  Therefore, we’d like to provide the option for a region to direct their infrastructure-specific and general feedback (basically, anything not related to arrival times) via established issue reporting sites (more detail below).
 
For 2, we plan to leverage some of the open issue reporting frameworks that already exist, such as [Open311]( http://www.open311.org/).  There are a number of Open311-compliant reporting sites such as [SeeClicFix.com]( http://seeclickfix.com/) (which [Pinellas County is already using]( http://seeclickfix.com/pinellas_county)) – or, regions could set up their own Open311 server.  Use of these 3rd party issue reporting sites would be option for each region (if you want to use it, provide the URL of the Open311 service you want to use in the OBA Directory.  If not, leave it blank).  This optional component will be similar in nature to the StopInfo project.
 
TODO:
·         Add open311 server URL to OBA Server Directory
·         Validate region before submitting report
·         Implement (or leverage existing) open311 client
·         Investigate/Implement backend integration with open311 service

barbeau:
Another item I'd like to add to this is to provide an option for "Ideas for new features", or some similar wording, that would direct the user to the OneBusAway Android Ideascale page:
http://onebusaway.ideascale.com/a/ideas/top/campaign-filter/byids/campaigns/46167

[...]

PR #249, opened 2015-03-18
WIP - BaseMapFragment enhancements

cagryInside:
We would like to use the BaseMapFragment in other Activities. For example, issue reporting implementation (issue https://github.com/OneBusAway/onebusaway-android/issues/187) requires to show bus stops in map view. However, BaseMapFragment doesn't return the clicked map location.
Hence, the following improvements were added to BaseMapFragment in order to satisfy different needs:
* onFocusChanged listener now returns the map clicked location.
* onFocusChanged notified every time instead of notifying when a user clicks on a bus stop.

PS: This PR was opened for discuss further enhancements on BaseMapFragment. Please ***do not*** merge.

### DIFF
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -146,8 +146,9 @@ public class BaseMapFragment extends SupportMapFragment
          *
          * @param stop   the ObaStop that obtained focus, or null if no stop is in focus
          * @param routes a HashMap of all route display names that serve this stop - key is routeId
+         * @param location the user touch location on the map
          */
-        void onFocusChanged(ObaStop stop, HashMap<String, ObaRoute> routes);
+        void onFocusChanged(ObaStop stop, HashMap<String, ObaRoute> routes, Location location);
     }
 
     @Override
@@ -408,7 +409,7 @@ public class BaseMapFragment extends SupportMapFragment
     //
     final Handler mStopChangedHandler = new Handler();
 
-    public void onFocusChanged(final ObaStop stop, final HashMap<String, ObaRoute> routes) {
+    public void onFocusChanged(final ObaStop stop, final HashMap<String, ObaRoute> routes, final Location location) {
         // Run in a separate thread, to avoid blocking UI for long running events
         mStopChangedHandler.post(new Runnable() {
             public void run() {
@@ -422,7 +423,7 @@ public class BaseMapFragment extends SupportMapFragment
                 }
 
                 // Pass overlay focus event up to listeners for this fragment
-                mOnFocusChangedListener.onFocusChanged(stop, routes);
+                mOnFocusChangedListener.onFocusChanged(stop, routes, location);
             }
         });
     }
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
@@ -15,22 +15,6 @@
  */
 package org.onebusaway.android.map.googlemapsv2;
 
-import com.amazon.geo.mapsv2.AmazonMap;
-import com.amazon.geo.mapsv2.Projection;
-import com.amazon.geo.mapsv2.model.BitmapDescriptor;
-import com.amazon.geo.mapsv2.model.BitmapDescriptorFactory;
-import com.amazon.geo.mapsv2.model.LatLng;
-import com.amazon.geo.mapsv2.model.Marker;
-import com.amazon.geo.mapsv2.model.MarkerOptions;
-
-import org.onebusaway.android.R;
-import org.onebusaway.android.app.Application;
-import org.onebusaway.android.io.ObaAnalytics;
-import org.onebusaway.android.io.elements.ObaReferences;
-import org.onebusaway.android.io.elements.ObaRoute;
-import org.onebusaway.android.io.elements.ObaStop;
-import org.onebusaway.android.util.LocationUtil;
-
 import android.app.Activity;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
@@ -50,6 +34,22 @@ import android.os.SystemClock;
 import android.util.Log;
 import android.view.animation.BounceInterpolator;
 import android.view.animation.Interpolator;
+
+import com.amazon.geo.mapsv2.AmazonMap;
+import com.amazon.geo.mapsv2.Projection;
+import com.amazon.geo.mapsv2.model.BitmapDescriptor;
+import com.amazon.geo.mapsv2.model.BitmapDescriptorFactory;
+import com.amazon.geo.mapsv2.model.LatLng;
+import com.amazon.geo.mapsv2.model.Marker;
+import com.amazon.geo.mapsv2.model.MarkerOptions;
+
+import org.onebusaway.android.R;
+import org.onebusaway.android.app.Application;
+import org.onebusaway.android.io.ObaAnalytics;
+import org.onebusaway.android.io.elements.ObaReferences;
+import org.onebusaway.android.io.elements.ObaRoute;
+import org.onebusaway.android.io.elements.ObaStop;
+import org.onebusaway.android.util.LocationUtil;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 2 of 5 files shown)


barbeau:
Thanks @cagryInside!  I plan to comment shortly.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
```diff
@@ -630,19 +630,20 @@ private void doFocusChange(ObaStop stop) {
...
         Log.d(TAG, "Map clicked");
 
-        // Only notify focus changed the first time the map is clicked away from a stop marker
         if (mMarkerData.getFocus() != null) {
             mMarkerData.removeFocus();
-            // Notify listener
-            mOnFocusChangedListener.onFocusChanged(null, null);
         }
+        // Set map clicked location
+        Location location = LocationUtil.makeLocation(latLng.latitude, latLng.longitude);
```
barbeau: This can be changed to `Location location = MapHelpV2.makeLocation(latLng);`

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
```diff
@@ -608,19 +609,21 @@ private void doFocusChange(ObaStop stop) {
...
-        mOnFocusChangedListener.onFocusChanged(stop, routes);
+        mOnFocusChangedListener.onFocusChanged(stop, routes, stop.getLocation());
     }
 
     @Override
     public void onMapClick(LatLng latLng) {
         Log.d(TAG, "Map clicked");
 
-        // Only notify focus changed the first time the map is clicked away from a stop marker
+
```
barbeau: Extra blank line here

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
```diff
@@ -608,19 +609,21 @@ private void doFocusChange(ObaStop stop) {
...
         if (mMarkerData.getFocus() != null) {
             mMarkerData.removeFocus();
-            // Notify listener
-            mOnFocusChangedListener.onFocusChanged(null, null);
         }
+        // Set map clicked location
+        Location location = LocationUtil.makeLocation(latLng.latitude, latLng.longitude);
```
barbeau: This can be changed to Location location = MapHelpV2.makeLocation(latLng);

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/StopOverlay.java
```diff
@@ -630,19 +630,20 @@ private void doFocusChange(ObaStop stop) {
...
+        // Notify focus changed every time the map is clicked away from a stop marker
```
barbeau: @cagryInside The main change in logic here is that we are now notifying all listeners each time the user taps on the map, instead of only the first time the user taps on the map while a stop is selected.  From a quick look at BaseMapFragment and HomeActivity onFocusChanged() methods I don't think this will be a problem.  Main area at risk would be HomeActivity.onFocusedChanged(), since the initial assumption when I wrote that code was we'd only be moving the MyLocationButton down, hiding the ArrivalListFragment, etc. when a stop was currently selected.  Now, this code is executing each time you tap on the map, even when a stop isn't selected.  I think this should be ok, but please take a close look at it (including stepping through the debugger) to make sure nothing strange is happening there.  I'll take a look myself too.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -586,9 +584,10 @@ private void autoShowWhatsNew() {
...
      * @param stop   the ObaStop that obtained focus, or null if no stop is in focus
      * @param routes a HashMap of all route display names that serve this stop - key is routeId
+     * @param location the user touch location on the map
      */
     @Override
-    public void onFocusChanged(ObaStop stop, HashMap<String, ObaRoute> routes) {
+    public void onFocusChanged(ObaStop stop, HashMap<String, ObaRoute> routes, Location location) {
         // Check to see if we're already focused on this same stop - if so, we shouldn't do anything
         if (mFocusedStopId != null && stop != null &&
                 mFocusedStopId.equalsIgnoreCase(stop.getId())) {
```
barbeau: We should clarify the below comment `// A particular stop lost focus (e.g., user tapped on the map), so hide the panel` to be `// No stop is in focus (e.g., user tapped on the map), so hide the panel`

barbeau:
@cagryInside In general this looks good, a few comments inline.  When you're sure everything is working right, let's merge this.  I'll open another WIP PR shortly to show you what I had in mind for turning the management of the current stop marker in the StopOverlay on/off .

cagryInside:
@barbeau Actually, I have couple more changes that I would like to discuss. When I finalize all changes, I will push them again. thanks

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -738,6 +739,8 @@ public void onLocationChanged(Location l) {
         if (firstFix) {
             // If its our first location (across all BaseMapFragments), move the camera to it
             setMyLocation(true, false);
+            //Notify the onFocusChangeListener when we get the first location
+            onFocusChanged(null,null,l);
```
cagryInside: @barbeau I add this to notify the onFocusChange Listener in the first time we get a location.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -745,4 +748,14 @@ public void onLocationChanged(Location l) {
...
+
+    /**
+     * mLastLocation is defined statically.
+     * If the BaseMapFragment was initialized before, the new  BaseMapFragment doesn't notifies
+     * the onFocusChanged listener (see L:742).
+     * resetLastLocation can clear previous mLastLocation data.
+     */
+    public void resetLastLocation(){
+        mLastLocation = null;
+    }
```
cagryInside: @barbeau I also need to add some method for ```mLastLocation``` as we discussed last week. Since, ```mLastLocation``` defined statically, If I create a new BaseMapFragment it never notifies the  onFocusChanged listener (L:727):
```java
 public void onLocationChanged(Location l) {
boolean firstFix = false;
if (mLastLocation == null) { 
// mLastLocation is static, so this is a first fix across all BaseMapFragments
firstFix = true;
}
mLastLocation = l;
if (mListener != null) {
// Show real-time location on map
mListener.onLocationChanged(l);
}
if (firstFix) {
// If its our first location (across all BaseMapFragments), move the camera to it
setMyLocation(true, false);
//Notify the onFocusChangeListener when we get the first location
onFocusChanged(null,null,l);
}
}
```

Please let me know if you have further comments. Maybe, we can handle this as an option in the constructor.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
barbeau: Per our discussion, let's remove this and pass the location we want the map to center on via an Intent to your Activity (when the user taps on "Send feedback" in the HomeActivity), which will then get passed to the BaseMapFragment.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
barbeau: Same as above - let's changed based on our discussion.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -744,4 +747,14 @@ public void onLocationChanged(Location l) {
...
+
+    /**
+     * mLastLocation is defined statically.
+     * If the BaseMapFragment was initialized before, the new  BaseMapFragment doesn't notifies
+     * the onFocusChanged listener (see L:742).
+     * resetLastLocation can clear previous mLastLocation data.
+     */
+    public void resetLastLocation(){
+        mLastLocation = null;
+    }
```
barbeau: Same as above.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -737,6 +738,8 @@ public void onLocationChanged(Location l) {
         if (firstFix) {
             // If its our first location (across all BaseMapFragments), move the camera to it
             setMyLocation(true, false);
+            //Notify the onFocusChangeListener when we get the first location
+            onFocusChanged(null,null,l);
```
barbeau: Same as above.

cagryInside:
@barbeau I made the final changes as we discussed. I believe it is ready for merge.

barbeau:

PR #250, opened 2015-03-20
WIP - Optional current stop marker in StopOverlay

barbeau:
@cagryInside Here's a sketch of what I had in mind for being able to use the BaseMapFragment in the Send Feedback screen in addition to the HomeActivity.  This allows setting a boolean value to define whether the StopOverlay manages the current stop marker.  If false, then it doesn't add the marker to the map.  See the TODO statement for details on how we can set this via an Intent to BaseMapFragment.

**Please don't merge.**


barbeau:
@cagryInside Per our discussion, I like your simpler solution better which just uses the callback onFocusChange() from BaseMapFragment with the location tapped on the map to show the currently selected location when a stop is not selected.  And, we just allow BaseMapFragment to manage the case when the user taps on the stop.

Issue #254, opened 2015-04-04
Better management of amazon/google build flavors common code


barbeau:
With the addition of support for the Amazon Fire Phone (#158), we now have two platform build flavors in the following packages:

* `src/google/java/org/onebusaway/android/map`
* `src/amazon/java/org/onebusaway/android/map`

With the exception of `ProprietaryMapHelpV2`, all other classes in these two packages are identical, except that they import classes from the amazon/google packages for the respective Maps APIs and reference the Amazon/GoogleMap object.

Here's our current manual process for maintaining these files:

>If changes are made to classes in src/google/java/org/onebusaway/android/map package, the same classes should be copied to src/amazon/java/org/onebusaway/android/map package.  Then, the Google classes should be replaced with equivalent Amazon classes as specified in http://bit.ly/AmazonMapsMigration. NOTE:  When doing this copying, ProprietaryMapHelpV2 should not need to be copied. It's specific to each platform.

We should be able to do better than manually copying/changing files.  I see two approaches:

1. Refactor these classes behind a common interface - I started down this path, but this gets ugly quick, since there is substantial logic in `BaseMapFragment` and `StopOverlay` that relies on the core classes used in the respective Maps APIs.  The result is interfaces for all the Maps API v2 classes, and its not clear to me that this is even possible due to the underlying management of SupportMapFragment, which is native to the Maps API, as well as the callbacks for detecting taps on a Marker, which are generated by the platform and provide a Marker Maps API v2 object.
2. Script the copy/replace logic - we should be able to accomplish this using Gradle and the `filterOnCopy` task - http://gradle.org/docs/current/userguide/userguide_single.html#filterOnCopy.  We would set this up so we'd edit the Google flavor classes, and then the Amazon classes would be auto-generated on the first Gradle build following the edit.  We'd also need to add some warning text of "this class is auto-generated, please don't edit" to the top of the Amazon package classes.

The second option seems like the simpler approach that keeps the code the cleanest, so I'm going to look into it further.

barbeau:
Gradle task for when file contents change:
https://gradle.org/docs/current/dsl/org.gradle.api.tasks.incremental.IncrementalTaskInputs.html

barbeau:
The following works as part of `build.gradle` to copy and filter the classes from the Google to Amazon build flavor:

~~~
/**
 * Copies our Maps API v2-based classes from the Google build flavor to the Amazon build flavor,
[...]

PR #255, opened 2015-04-06
WIP - Fix #187 Issue reporting implementation

### LINKED ISSUE #187: Improve issue reporting

cagryInside:
This is the issue reporting enhancement implementation as described in (https://github.com/OneBusAway/onebusaway-android/issues/187). 

List of enhancements:
* Open311 issue reporting specification (http://wiki.open311.org/GeoReport_v2/) was adopted and implemented for OBA.
* Validate region before submitting report. 
* Integration with Open311-compliant reporting sites such as SeeClicFix.com.
* New issue reporting flow

Send Feedback section  |  Region Validation
:-------------------------:|:-------------------------:
device-2015-04-06-115134 |  device-2015-04-06-115034

Reporting categories |  Infrastructure Issue (There is no open311 issue category for current location)
:-------------------------:|:-------------------------:
device-2015-04-06-115444 | device-2015-04-06-115545

 Infrastructure Issue, when a bus stop selected (There is no open311 issue category for current location) |  Infrastructure Issue (There is open311 issue category for current location)
:-------------------------:|:-------------------------:
device-2015-04-06-115858 | device-2015-04-06-115935

TODO:
* Add production open311 servers URL to OBA Server Directory
* Export Open311 package as external library
* Amazon device integration


onebusaway-android/build.gradle
```diff
@@ -148,4 +151,8 @@ dependencies {
     compile 'org.onebusaway.util:comparators:1.0.0'
     // Only used on Amazon devices via amazon build flavor
     amazonCompile('com.amazon.android:amazon-maps-api:2.0')
+
+    //Apache http connection for images
```
barbeau: @cagryInside Can you use the native Android HttpURLConnection for this instead?  If not, we should include comments here or in code why.  See http://android-developers.blogspot.com/2011/09/androids-http-clients.html for why the HttpURLConnection is preferred to Apache libaries.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
```diff
@@ -0,0 +1,443 @@
...
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
+import android.view.View;
+import android.view.Window;
+import android.widget.ArrayAdapter;
+import android.widget.EditText;
+import android.widget.ImageButton;
+import android.widget.Toast;
+
+import com.google.android.gms.maps.model.Marker;
```
barbeau: @cagryInside We should only import the Maps API v2 objects in the maps package.  So, we need to figure out a way to remove these imports and modify BaseMapFragment so you can add the marker from this class but by passing in non Maps API v2 objects.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /**
+     * Places a marker into the map from given Location
+     *
+     * @param location position for marker
+     */
+    private void updateMarkerPosition(Location location) {
+
+        Marker marker = mMapFragment.getMap().addMarker(new MarkerOptions().
```
barbeau: @cagryInside We should only import the Maps API v2 objects in the maps package (this is why the Amazon build flavor is currently failing on Travis).  You may be able to accomplish this by adding a single method to BaseMapFragment that allows you to add a marker by passing in a lat/long.  This means you'd also need to change the logic here so you're not maintaining a reference to a Marker object.

onebusaway-android/build.gradle
cagryInside: @barbeau thanks for the documentation. I added these libraries to be able to send images. In the city of bloomington open311 implementation (https://github.com/City-of-Bloomington/open311-android), these libraries were used to send images as an attachment with a report. In our application, this implementation worked for test servers, however it didn't worked for SeeClickFix (We already send them an email regarding this issue). 

I will investigate sending images with HttpURLConnection (It should be possible, since, we are sending just byte array). In the worst case, after SeeClickFix fixes their issues and when we test everything, I am planing the export this part as a library project as we discussed earlier.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
cagryInside: @barbeau This issue was on my question list, thanks for pointing out. I think a method in BaseMapFragment sounds like a good solution. I will work on that.

onebusaway-android/build.gradle
barbeau: Ok, sounds good.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
barbeau: Ok, great.  You can just implement this in the Google build flavor maps package.  I'll submit a PR shortly for #254 that will automatically script the copying of any changes to the Amazon build flavor maps package.

cagryInside:
@barbeau Also, I implemented "ideas for new features" option as you mentioned here: https://github.com/OneBusAway/onebusaway-android/issues/187#issuecomment-77400599. I used the idea scale's  official icon, but I am not sure if it's fine for copyright purposes. 
ed7f7e6c-dc53-11e4-9fd1-bae387c3f203

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
cagryInside: Thanks, that's really useful enhancement.

barbeau:
@cagryInside Thanks for adding this!  I saw it in the screenshots you sent earlier.  Let's change the image to a light bulb instead, which is fairly commonly used to represent ideas.  This also avoids any potential copyright issues with Ideascale (and most users probably don't know about the Ideascale logo anyway).

Some possible light bulb images can be found on the NounProject:
https://thenounproject.com/search/?q=light-bulb

Maybe https://thenounproject.com/term/light-bulb/18707/, which is under Creative Commons?  We'd just need to provide attribute somewhere.

You can include a black icon with transparent background in the OBA Android project, and then use the ImageView color filter to change it to the primary theme color.  See https://github.com/OneBusAway/onebusaway-android/blob/develop/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java#L488 for an example of how the Navigation Drawer icons are shaded according to the `navdrawer_icon_tint` color (which is mapped to the primary theme color `theme_primary` in `colors.xml`).

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -738,6 +738,7 @@ public void onLocationChanged(Location l) {
         if (firstFix) {
             // If its our first location (across all BaseMapFragments), move the camera to it
             setMyLocation(true, false);
+            onFocusChanged(null,null,l);
```
barbeau: Just realized that this change is still in this branch.  We changed this model in a previous PR to not using this callback on first GPS fix, so this branch should be rebased on develop to take advantage of the new structure.  You can wait to do this until we merge the other PR for scripting Amazon classes so all the debasing is done in one step.

onebusaway-android/src/main/AndroidManifest.xml
```diff
@@ -133,6 +133,9 @@
         <activity android:name="org.onebusaway.android.ui.RegionsActivity"/>
         <activity android:name="org.onebusaway.android.ui.AgenciesActivity"/>
 
+        <activity android:name=".report.ui.ReportActivity"/>
```
barbeau: Let's list the fully qualified activity name here to match the other activities in the manifest.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
cagryInside: @barbeau Yes, my bad. Probably when I was merging the previous pr I missed this. I think the new PR will help to prevent similar problems in the future.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
```diff
@@ -218,6 +218,17 @@ public String getStopInfoUrl() {
...
 
     @Override
+    public String getOpen311Url() {
+        return "http://test.seeclickfix.com/open311/v2/";
+    }
+
+    @Override
+    public String getOpen311JurisdictionId() {
+        return "Tampa";
+//        return "miamidade.gov";
```
cagryInside: This is hard-coded for test purposes. It needs to be changed.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
cagryInside: @barbeau I rebase the PR https://github.com/OneBusAway/onebusaway-android/issues/254 on top of this branch. I was going to implement this particular marker issue in this PR, but I thought it might be confusing. Hence, to avoid confusion for future use, I opened an issue (https://github.com/OneBusAway/onebusaway-android/issues/260) regarding this issue. I will commit a solution shortly.

barbeau: Sounds good, let's hammer out design in a new PR and issue #260.

cagryInside:
@barbeau 
I have implemented the new marker adding function. Furthermore, I revised the Issue type icons and adopted the similar approach with https://github.com/OneBusAway/onebusaway-android/blob/develop/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java#L488.

device-2015-04-25-145408

barbeau:
Awesome, thanks!  Looks good (the screenshot at least, haven't gotten to the code yet ;) )

One comment - could you please flip the lightbulb upside down, so the socket end is facing down and the glass bulb is facing up?  I think this is the more common representation for ideas.

I like the text bubble icon for app feedback.  Can you think of any other icons for "infrastructure issues"?  Maybe a bus stop sign?  The one we have now is ok, but I wonder if we can find one that is better.  If you find candidates, feel free to just add them in-line in a comment here before you actually add them to the project, just so we can comment on them before you spend time integrating into project.

cagryInside:
@barbeau thanks. 

> One comment - could you please flip the lightbulb upside down, so the socket end is facing down and the glass bulb is facing up? I think this is the more common representation for ideas.

I agree on that, I will fix it.



> I like the text bubble icon for app feedback.

This is similar to the send app feedback icon for Google maps.



> Can you think of any other icons for "infrastructure issues"? Maybe a bus stop sign? The one we have now is ok, but I wonder if we can find one that is better.

Actually, I already try with the current stop sign image, however since I start using  the similar approach with ```NavigationDrawerFragment```, when it changes the color, it looks weird. That's why I would like to chose all icons abstract style. I will keep looking for better icons.

barbeau:
@cagryInside Also - if you run `gradlew connectedCheck` on your local machine with your device plugged in, does it pass?  Travis is failing and It would be good to know if this passes on a local device.

cagryInside:
> @cagryInside Also - if you run gradlew connectedCheck on your local machine with your device plugged in, does it pass? Travis is failing and it would be good to know if this passes on a local device.

@barbeau It passes with my device plugged in, but if I don't plugged in any device, it gives the same error.

cagryInside:
> Can you think of any other icons for "infrastructure issues"?

@barbeau 
Here is couple candidate icons for  "infrastructure issues":

 1 - Maybe something related to a bus: 

ic_launcher

 2 - Or, generic infrastructure icons:

ic_launcher
ic_launcher

 3 - Or, some question and/or answer style icons:

ic_launcher
ic_launcher

barbeau:
Thanks for looking.  I like the current icon better than all these, so
let's just stick with it for now.

PR #259, opened 2015-04-17
Fix #254 - Automate updating of Amazon build flavor classes based on Google build flavor

### LINKED ISSUE #254: Better management of amazon/google build flavors common code

barbeau:
This PR implements scripting of the copy/replace logic to update the classes in the Amazon build flavor maps package (`src/amazon/java/org/onebusaway/android/map`) based on changes to the Google build flavor maps package (`src/google/java/org/onebusaway/android/map`), as a solution to #254.  It makes the following changes to the Amazon classes after copying the Google classes:

* `com.google.android.gms.maps` is changed to `com.amazon.geo.mapsv2`
* `GoogleMap` is changed to `AmazonMap`
* A header is added to the Amazon classes with a "do not manually change" message to indicate that those files should always be automatically generated based on changes to the Google classes.

Gradle copy, filter, and incremental changes features are used to script the process in `build.gradle`.  The only class that isn't automatically copied is `ProprietaryMapHelpV2`, since this class contains code specific to each build flavor.

Expected behavior is that without any changes to the Google build flavor map package classes, in the Gradle console (visible in Android Studio or on command line) towards the top of the build you should see the line:

`:onebusaway-android:copyMapsApiV2Classes UP-TO-DATE`

This means that no copying has taken place and the build is taking place as normal.

However, if a class in the Google build flavor map package is changed (excluding `ProprietaryMapHelpV2`), then you should see:

`:onebusaway-android:copyMapsApiV2Classes`

...which means that the `copyMapsApiV2Classes` task in `build.gradle` is executing.  And, after the build finishes, you should see the changes you made in the Google class reflected in the respective Amazon class.  So, when you go to commit the changes, the identical changes should be seen for the same Amazon and Google classes.

@cagryInside Could you please give this a quick look to make sure its working as intended on your machine?  It seems to work fine for me, but I'm also somewhat sleep-deprived :).  After this is merged to develop, if you rebase your work in #255 on this it should automatically copy your changes to `BaseMapFragment` in the Google folder to the Amazon folder.  You should also probably keep a separate backup of your work in another branch just in case something goes horribly wrong with this patch :).


cagryInside:
> @cagryInside Could you please give this a quick look to make sure its working as intended on your machine? It seems to work fine for me, but I'm also somewhat sleep-deprived :). After this is merged to develop, if you rebase your work in #255 on this it should automatically copy your changes to BaseMapFragment in the Google folder to the Amazon folder. You should also probably keep a separate backup of your work in another branch just in case something goes horribly wrong with this patch :).


@barbeau I look at the code, and IMO it looks good. Furthermore, I checkout this PR, and tested with couple cases on my machine. I haven't found any problems. I would also like to check the changes on Amazon device, however I left the device at CUTR. I will do my final tests with the amazon device tomorrow and let you know, thanks.

cagryInside:
@barbeau I run further tests on Amazon phone. It looks working fine.

cagryInside:
Also, I notice that the license header is missing in ```ProprietaryMapHelpV2``` for both [google](https://github.com/OneBusAway/onebusaway-android/blob/copyMapsFiles/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/ProprietaryMapHelpV2.java)  and [amazon](https://github.com/OneBusAway/onebusaway-android/blob/copyMapsFiles/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/ProprietaryMapHelpV2.java)  versions.

barbeau:
Thanks @cagryInside!  I'll push a commit to develop shortly (outside of this PR) that fixes the license headers on `ProprietaryMapHelpV2`.

barbeau:
Ok, license header is added in https://github.com/OneBusAway/onebusaway-android/commit/96a291cf590cacb05ab9e373f1cac690b009e654.  Thanks @cagryInside for noticing this!

Issue #256, opened 2015-04-06
StopProblem and TripProblem fragments can report issues without selecting a problem category


cagryInside:
It should be required to choose a problem category in order to send a stop or trip problem.
In the ```ReportStopProblemFragment``` and ```ReportTripProblemFragment``` users can currently send issues without selecting a problem category.

![device-2015-04-06-132931](https://cloud.githubusercontent.com/assets/2777974/7008519/85184242-dc61-11e4-9e75-15a30e89960c.png)

cagryInside:
Fixed via PR https://github.com/OneBusAway/onebusaway-android/pull/261

PR #261, opened 2015-04-22
Issue #256 - StopProblem and TripProblem fragments can report issues without selecting a problem category

cagryInside:
A stop and trip problem validation methods were implemented to prevent sending problems without choosing a problem category.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
@@ -15,17 +15,6 @@
  */
 package org.onebusaway.android.ui;
 
-import com.google.android.gms.common.ConnectionResult;
-import com.google.android.gms.common.GooglePlayServicesUtil;
-import com.google.android.gms.common.api.GoogleApiClient;
-
-import org.onebusaway.android.R;
-import org.onebusaway.android.io.ObaAnalytics;
-import org.onebusaway.android.io.ObaApi;
-import org.onebusaway.android.io.request.ObaResponse;
-import org.onebusaway.android.util.LocationUtil;
-import org.onebusaway.android.util.UIHelp;
-
 import android.app.Activity;
 import android.content.Context;
 import android.os.Bundle;
@@ -40,7 +29,19 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.Spinner;
 import android.widget.Toast;
+
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GooglePlayServicesUtil;
+import com.google.android.gms.common.api.GoogleApiClient;
+
+import org.onebusaway.android.R;
+import org.onebusaway.android.io.ObaAnalytics;
+import org.onebusaway.android.io.ObaApi;
+import org.onebusaway.android.io.request.ObaResponse;
+import org.onebusaway.android.util.LocationUtil;
+import org.onebusaway.android.util.UIHelp;
 
 import java.util.concurrent.Callable;
 
@@ -54,6 +55,16 @@ public abstract class ReportProblemFragmentBase extends Fragment
      * GoogleApiClient being used for Location Services
      */
     GoogleApiClient mGoogleApiClient;
+
+    /**
+     * Displays the problem categories
+     */
+    Spinner mCodeView;
+
+    /**
+     * Stores the problem category codes
+     */
+    String[] SPINNER_TO_CODE;
 
     @Override
     public void onAttach(Activity activity) {
@@ -114,6 +125,17 @@ public abstract class ReportProblemFragmentBase extends Fragment
             sendReport();
         }
         return false;
+    }
+
+    /**
+     * Validates report arguments before submitting a problem
+     *
+     * @return false if the report is not valid
+     */
+    protected boolean isReportArgumentsValid() {
+        // Check if a user selects a problem category
+        String code = SPINNER_TO_CODE[mCodeView.getSelectedItemPosition()];
+        return code != null;
     }
 
     protected void sendReport() {
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
@@ -15,13 +15,6 @@
  */
 package org.onebusaway.android.ui;
 
-import org.onebusaway.android.R;
-import org.onebusaway.android.io.ObaAnalytics;
-import org.onebusaway.android.io.elements.ObaStop;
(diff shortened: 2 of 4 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
```diff
@@ -55,6 +55,13 @@
      */
     GoogleApiClient mGoogleApiClient;
 
+    /**
```
barbeau: I wouldn't define this as an abstract method in the base class.  This method isn't used in the base class, so it should just be implemented as a normal method in the subclasses.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
```diff
@@ -110,16 +111,31 @@ public void onSaveInstanceState(Bundle outState) {
     }
 
     @Override
+    protected Boolean isReportArgumentsValid() {
```
barbeau: I believe this can be `boolean`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportTripProblemFragment.java
```diff
@@ -156,14 +157,27 @@ public void onSaveInstanceState(Bundle outState) {
     }
 
     @Override
+    protected Boolean isReportArgumentsValid() {
```
barbeau: I believe this can be `boolean`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
barbeau: I see what you're trying to do, though, in terms of avoiding duplicate code in subclasses.  IMHO, for this to work the method should be implemented in the base class, which means you'd also need to move the declaration of `Spinner mCodeView` and `String[] SPINNER_TO_CODE` to the base class as well, and then just reference/assign them in the subclasses.  Want to give that a shot?

barbeau:
@cagryInside Thanks for fixing this!  A few comments inline for your review.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
cagryInside: @barbeau In addition, I thought this method's implementation can vary in the future. In other words, if someone wants to implement another ```Report*ProblemFragment``` by extending  ```ReportProblemFragmentBase```, they may want to implement different validation method. Hence, I tried to use abstraction on this method. Your solution is going to solve the duplicate code in subclasses, but it might limit the future usage. (Or maybe I am over thinking :))

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
cagryInside: @barbeau agreed. Sorry about that, I had a bad experience about primitive types in the past. I am little bit extra careful about this.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
barbeau: Yeah, I think I'd avoid using the abstract method just for the case of it being used in the future, and just focus on the current use.  If you implement this method as a normal base class method, future subclasses can simply override it to implement different validation logic, if needed.  Or, just leave it in the subclasses.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
barbeau: Np.  My personal preference is to only use objects instead of primitives if we need to check for a null value, otherwise just stick to primitives.

cagryInside:
@barbeau pushed the latest changes based on our discussion.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportProblemFragmentBase.java
```diff
@@ -116,6 +127,17 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return false;
     }
 
+    /**
+     * Validates report arguments before submitting a problem
+     *
+     * @return false if the report is not valid
+     */
+    protected boolean isReportArgumentsValid() {
+        //Check if a user selects a problem category
```
barbeau: @cagryInside Could you please change the comments to the format of `// Check...`?  There should be a space between the double slash and the first word.  I've been guilty of not formatting this correctly in the past, and I've been trying to change to the correct format as I touch the code.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportStopProblemFragment.java
```diff
@@ -115,19 +123,17 @@ protected void sendReport() {
...
-            ObaReportProblemWithStopRequest.LOCATION_WRONG,
-            ObaReportProblemWithStopRequest.ROUTE_OR_TRIP_MISSING,
-            ObaReportProblemWithStopRequest.OTHER
-    };
+        if (isReportArgumentsValid()) {
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.SUBMIT.toString(),
+                    getString(R.string.analytics_action_problem), getString(R.string.analytics_label_report_stop_problem));
+            super.sendReport();
+        } else {
+            //Show error message if report arguments is not valid
```
barbeau: Same comment format.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ReportTripProblemFragment.java
```diff
@@ -161,21 +170,17 @@ protected void sendReport() {
...
         imm.hideSoftInputFromWindow(mUserComment.getWindowToken(), 0);
-        ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.SUBMIT.toString(),
-                getString(R.string.analytics_action_problem), getString(R.string.analytics_label_report_trip_problem));
-        super.sendReport();
+        if (isReportArgumentsValid()) {
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.SUBMIT.toString(),
+                    getString(R.string.analytics_action_problem), getString(R.string.analytics_label_report_trip_problem));
+            super.sendReport();
+        } else {
+            //Show error message if report arguments is not valid
```
barbeau: Same comment format.

barbeau:
@cagryInside Thanks, this looks good.  One minor issue with comment formatting and then I think it will be good to go.

cagryInside:
@barbeau thanks. I just pushed the final version (including the comment formatting).

barbeau:

Issue #260, opened 2015-04-21
Add custom markers to BaseMapFragment


cagryInside:
We have implemented several enhancements (PR https://github.com/OneBusAway/onebusaway-android/pull/249) in ```BaseMapFragment``` to be able to use the  ```BaseMapFragment``` in other activities. However, when we are using Maps API v2 objects (e.g., adding a marker: ```mMapFragment.getMap().addMarker(...)``` ) outside of the maps package, Amazon build flavor is failing (https://github.com/OneBusAway/onebusaway-android/pull/255#discussion_r28558592). 

One possible approach could be to implement a method in the  ```BaseMapFragment``` which adds a marker from given lat/long.

cagryInside:
@barbeau I implemented a method (https://github.com/cagryInside/onebusaway-android/commit/e9a1fa46a6d6ca8cb5cb69e1b60885e18fe688f1#diff-5542c1da09f3ac946281c3f258d88e8cR739) that adds markers to ```BaseMapFragment```. It seems working (I only tested with Google phone), but I will also test this with Amazon phone tomorrow and open a PR.

Furthermore, I would like to add this method to ```ObaMapView``` interface, however I had to return created marker from the method. So, I couldn't add this method as interface. Created marker object is important because, if we want to remove the marker from the map, simply we can call ```marker.remove()```.

cagryInside:
Also, I tested this development with Amazon firephone.

barbeau:
@cagryInside I just opened https://github.com/OneBusAway/onebusaway-android/pull/263, please check it out and see if it meets your needs for use in #255.

cagryInside:
@barbeau Thanks, I checked the implementation. It seems it's going to solve all marker related problems. I will test this with Amazon phone tomorrow, and let you know.

cagryInside:
Closed in favor of PR https://github.com/OneBusAway/onebusaway-android/pull/263

PR #262, opened 2015-04-22
Issue #260 - adding custom markers to BaseMapFragment

cagryInside:
This PR implements a method in the ```BaseMapFragment```. The method enables to add custom markers to the Map.

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -16,6 +16,17 @@
  */
 package org.onebusaway.android.map.googlemapsv2;
 
+import android.content.Context;
+import android.location.Location;
+import android.os.Bundle;
+import android.os.Handler;
+import android.text.TextUtils;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -727,6 +729,20 @@ public void deactivate() {
...
     }
 
+    /**
+     * Adds a marker to the Map
+     *
+     * @param l is the marker's position in the map
+     * @return the added marker object,
+     * the marker object can be used to remove the marker from the map by calling marker.remove()
+     */
+    public Marker addMarker(Location l) {
```
barbeau: @cagryInside This will still cause the `Marker` class to be imported into classes outside the `map` package.  I'll submit a PR shortly with a design that I believe will solve this problem and still allow the addition of markers from classes outside the package.

cagryInside: @barbeau yes, sorry. I couldn't see that when we assign the return result, we need to import the ```Marker```.

barbeau: No worries! :)

barbeau:
Closed in favor of #263

PR #263, opened 2015-04-22
Fix #260 - Add SimpleMarkerOverlay to allow classes outside map package ...

### LINKED ISSUE #260: Add custom markers to BaseMapFragment

barbeau:
...to add/remove generic markers in BaseMapFragment

* This still allows management via the ObaMapView interface as well


barbeau:
@cagryInside Please give this a test on both Google and Amazon build flavors.  I haven't tested it, but it builds fine on both flavors on my machine.  Hopefully this solves the problem for #255.

cagryInside:
@barbeau I have tested this on both Google and Amazon devices. They both work fine without having any problem. Thanks.

barbeau:
Great, thanks!  I'll go ahead and merge so you can rebase #255 and test further.

PR #177, opened 2014-09-19
WIP - Issue #155 - UI re-design

barbeau:
Please don't merge - this is a work-in-progress pull request so others can keep track of progress and provide feedback on the UI re-design (#155), including Maps API v2 implementation (https://github.com/OneBusAway/onebusaway-android/issues/56) and StopInfo project (https://github.com/OneBusAway/onebusaway-android/issues/103).

I'm updating this top comment with progress as we go.

To try out the UI re-design, please [register as an alpha tester](https://github.com/OneBusAway/onebusaway-android/blob/master/BETA_TESTING.md#register-as-an-alpha-tester)!

**Known issues:**
* See bugs under [public v2.0.x release milestone](https://github.com/OneBusAway/onebusaway-android/milestones/Public%20v2.0.x%20release)

**TODO:**
* See enhancement issues under [public v2.0.x release milestone](https://github.com/OneBusAway/onebusaway-android/milestones/Public%20v2.0.x%20release)

Feedback is appreciated!  Screenshots are below.

image

@caitbonnar APK to install and test StopInfo is linked above in this PR.

PR #266, opened 2015-05-11
Fix #265 - Hide soft keyboard when leaving Search fragments

### LINKED ISSUE #265: Minor UI bug: keyboard doesn't hide after navigating away from the search tab
This is for 1.7.8 of the app, if that's relevant.
Please close if this is irrelevant, already fixed, or not possible.

0. Summary: The on-screen keyboard doesn't hide after navigating away from the search tab, it's a bit annoying to have to tap the hide icon every time.
0. Steps to reproduce: Open the app, tap the menu (3 dot thing in the corner), tap "my stops", it defaults to "search" (not sure if that's most preferred behavior, especially for daily commuters) & the keyboard shows up because the search box is selected automatically, then tap either the starred or recent tabs and the on screen keyboard will still be there.
0. Expected behavior: The on screen keyboard should hide itself if the user navigates away from the search tab
0. Observed behavior: Already mentioned in point 2
0. Device and Android version: Nexus 4, CyanogenMod 11 M 12 (Android 4.4.4 KitKat) - but I believe I saw this even when I was using stock Android.
0. Screenshots: 
  - 
screenshot_2015-05-08-23-55-51

0. Possible fixes: Not sure since I haven't done much Android programming. A theoretical fix would be when the user is in the "my stops" menu, check which tab the user is on, if it's not "Search" then hide the keyboard if it's showing.

Thanks for this app! I've been using it for a few years and it's saved me many times :smile:

barbeau:
@cagryInside Would you mind giving this a quick review in the next few days?  It works fine on my LG G3 w/ Android 5.0.1 to fix the problem described in #265, but I'd like to confirm it works on the devices you have access to as well.

### DIFF
--- a/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/MySearchFragmentBase.java
+++ b/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/MySearchFragmentBase.java
@@ -15,7 +15,16 @@
  */
 package com.joulespersecond.seattlebusbot;
 
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GooglePlayServicesUtil;
+import com.google.android.gms.location.LocationClient;
+
+import com.joulespersecond.seattlebusbot.util.LocationHelp;
+import com.joulespersecond.seattlebusbot.util.UIHelp;
+import com.joulespersecond.view.SearchView;
+
 import android.app.Activity;
+import android.content.Context;
 import android.content.Intent;
 import android.location.Location;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


cagryInside:
@barbeau I just try with Nexus 6 (Android 5.1) and a tablet (Android 4.4.2). It works fine on both devices. However, I noticed one behavior: 

1 . Switched from "search" tab to "recent" or "starred" tab (it closes the keyboard, which is desired)
2 . Switched from "recent" or "starred" tab to "search" tab. (This time, it doesn't focus on the edit text, so the keyboard doesn't show up. It doesn't look like a big deal but I wanted to share)

device-2015-05-10-222000

cagryInside:
Also tested with HTC EVO 3D (android 4.0.1) and the behavior was same with  https://github.com/OneBusAway/onebusaway-android/pull/266#issuecomment-100741082

barbeau:
Thanks @cagryInside!  Yeah, I considered changing the keyboard opening behavior too.  I even tried a few quick fixes (calling `mSearchView.requestFocus()` in `onStart()`, and executing the opening keyboard code directly in `onStart()`, but neither of these changed the behavior.  I'm going to just leave it as-is for now, since I'd prefer not to sink a lot of time into this code, since it's going to be abandoned in favor of the unified search integrated into the Action Bar in the `develop` branch as part of #177 UI redesign.

Issue #264, opened 2015-04-23
Improve stop/trip problem reporting API requests


barbeau:
Looks like there is an issue with how OBA Android is sending REST API requests for stop/trip problem reporting - see https://github.com/OneBusAway/onebusaway-application-modules/issues/126#issuecomment-94544761.

If we can confirm that the documented APIs are currently working, we should change to using that method:

* http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/report-problem-with-stop.html
* http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/report-problem-with-trip.html

barbeau:
@cagryInside Could you please take a shot at fixing this?  You should be able to use the app.staging.obahart.org machine to test (i.e., you can log into the admin webapp to see the problem reports).  There is an existing PR open here, but there are still some issues to be worked out - https://github.com/OneBusAway/onebusaway-android/pull/382.

barbeau:
Newest status - issue reporting seems to be working in Puget Sound, but not Tampa or Atlanta.  So to fix this issue, we need to test against at least those three regions and make sure it works.

barbeau:
And we should probably confirm with admins of Puget Sound and Atlanta that they actually get the problems reports after we fix this.

cagryInside:
@barbeau I just pushed a fix for this issue.  There was a problem with post requests (Sheldon mentioned   [here] (https://github.com/OneBusAway/onebusaway-application-modules/issues/126#issuecomment-94544761)). So, I used get requests instead of post requests for stop/trip problem reporting. Further, there was a parsing problem [here] (https://github.com/cagryInside/onebusaway-android/commit/10fa4b34e7095ee8e271034a700767a1454ba355#diff-0c511ed41cc723fd7c7df26cc2136ed6R179). It wasn't checking the response's error fields.

barbeau:
@cagryInside Awesome!  Just copied you on an email - I want to test this with the 3 main production OBA instances to confirm it works, then you can submit a PR.

cagryInside:
Atlanta's server returns an internal error for API request:

Request:
```
[...]

PR #276, opened 2015-07-11
report-problem-with-* requests follow API, added tests

k7lim:
addresses #264 

- Changed requests to use the APIs documented in #264, id value moved from query string to request filename (e.g. from `report-problem-with-stop.json` with `stopId` in post body, to `report-problem-with-stop/1_75403.json`)  

- added tests for report-problem-with-stop and report-problem-with-trip, along with necessary `MockConnection.post()` implementation and test uri resources.

- Does not address POST request issues specified in : https://github.com/OneBusAway/onebusaway-application-modules/issues/126#issuecomment-94544761

- EDIT: also cleans out redundant `data` parameter, appears to be a relic of old API.

TODO: verify that this change (still) correctly submits problem reports. In my debug install, I'm seeing empty 200 responses to the submit problem requests.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemRequestTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemRequestTest.java
@@ -1,0 +1,62 @@
+package org.onebusaway.android.io.test;
+
+
+import org.onebusaway.android.UriAssert;
+import org.onebusaway.android.io.request.ObaReportProblemWithStopRequest;
+import org.onebusaway.android.io.request.ObaReportProblemWithStopResponse;
+import org.onebusaway.android.io.request.ObaReportProblemWithTripRequest;
+import org.onebusaway.android.io.request.ObaReportProblemWithTripResponse;
+
+import java.util.HashMap;
+
+/**
+ * Tests Report Problems API requests and responses
+ */
+public class ReportProblemRequestTest extends ObaTestCase {
+
+    public void testProblemWithStopRequest() {
+        ObaReportProblemWithStopRequest.Builder builder =
+                new ObaReportProblemWithStopRequest.Builder(getContext(), "1_75403");
+        ObaReportProblemWithStopRequest request = builder.build();
+        ObaReportProblemWithStopResponse response = request.call();
+        assertOK(response);
+    }
+
+    public void testProblemWithStopBuilder() {
+        ObaReportProblemWithStopRequest.Builder builder =
+                new ObaReportProblemWithStopRequest.Builder(getContext(), "1_75403");
+        ObaReportProblemWithStopRequest request = builder.build();
+        assertNotNull(request);
+        UriAssert.assertUriMatch(
+                "http://api.pugetsound.onebusaway.org/api/where/report-problem-with-stop/1_75403.json",
+                new HashMap<String, String>() {{
+                    put("key", "*");
+                    put("version", "2");
+                }},
+                request
+        );
+    }
+
+    public void testProblemWithTripRequest() {
+        ObaReportProblemWithTripRequest.Builder builder =
+                new ObaReportProblemWithTripRequest.Builder(getContext(), "1_28322560");
+        ObaReportProblemWithTripRequest request = builder.build();
+        ObaReportProblemWithTripResponse response = request.call();
+        assertOK(response);
+    }
+
+    public void testProblemWithTripBuilder() {
+        ObaReportProblemWithTripRequest.Builder builder =
+                new ObaReportProblemWithTripRequest.Builder(getContext(), "1_28322560");
+        ObaReportProblemWithTripRequest request = builder.build();
+        assertNotNull(request);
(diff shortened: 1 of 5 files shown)


barbeau:
Thanks @k7lim!  I'll likely take a look at this on Monday.  If you have already, could you also please sign the [ICLA ](https://docs.google.com/forms/d/12jV-ByyN186MuPotMvxJtNKtSaGGTnEHm8rXomM2bm4/viewform)?

barbeau:
@k7lim This fails to send a problem report when I point the app at OBA Tampa - I see:

>07-13 12:11:21.462    2324-3298/com.joulespersecond.seattlebusbot E/RequestBase﹕ java.io.FileNotFoundException: http://api.tampa.onebusaway.org/api/api/where/report-problem-with-trip/Hillsborough Area Regional Transit_92510.json?app_ver=51&app_uid=3c65025dab64fde246049111c2714275&version=2&key=v1_BktoDJ2gJlu6nLM6LsT9H8IUbWc%3DcGF1bGN3YXR0c0BnbWFpbC5jb20%3D

Could you take a look at this?  My guess is that it's no longer encoding the spaces in the agency name (which is used if `agency_id` isn't provided in the GTFS data).  If so, then we could also use a test for this case when agency name is used instead of an interger ID value.

nbcklim:
@barbeau signed the ICLA

nbcklim:
@barbeau will look into it. I think your hunch sounds right.

k7lim:
@barbeau signed the ICLA

k7lim:
@barbeau will look into it! your hunch sounds right

k7lim:
@barbeau added uri encoding to the tripid parameter

barbeau:
Thanks @k7lim!

Unfortunately, though, there is still something going wrong.  Even though an HTTP 200 is returned, I'm not seeing anything appear in the OBA webapp admin console when reporting problems in OBA Tampa.

When using the production (`master` branch) version of OBA as well as the `develop` branch, I can see the problem report in OBA Tampa:
image

Without digging in myself, I'm assuming this is a potential side effect of the issue discussed in the thread https://github.com/OneBusAway/onebusaway-application-modules/issues/126#issuecomment-94544761.

We should really update OBA Android to use the officially documented APIs here (assuming they are now working):
* http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/report-problem-with-stop.html
* http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/report-problem-with-trip.html

If you're able to tackle this yourself that would be great, if not I'll leave this PR open for when I eventually get to this.

Either way, thanks for your work on this!

k7lim:
@barbeau I had intended to sync the Android requests with those APIs.  The main difference I see is the `master` hits `/report-problem-with-*.json` with the entity ID as a POST param.   So I changed it to reflect the docs, which use `report-problem-with-*/entityID.json` .  

The docs are not clear on a few things: 
Is the report-problem-with-* request supposed to be a POST? I assumed so since it's creating a new record.

Is it important that the request be for .XML ?  The examples in the documentation (which actually 500 out) are for the XML RESTful API, but the parent documentation indicates that .JSON is fine too.

k7lim:
So, in short, I can't verify if the API is working, and I thought I had changed Android to use the documented APIs.

barbeau:
>The main difference I see is the master hits /report-problem-with-*.json with the entity ID as a POST param. So I changed it to reflect the docs, which use report-problem-with-*/entityID.json .

Yes, this is good.

>Is the report-problem-with-* request supposed to be a POST? I assumed so since it's creating a new record.

I had assumed so as well.  This code was written before I started maintaining OBA, so unfortunately I'm in the same boat as you :).

>Is it important that the request be for .XML ? The examples in the documentation (which actually 500 out) are for the XML RESTful API, but the parent documentation indicates that .JSON is fine too.

JSON is fine and preferred.  All API requests made by OBA Android should (and are) using the JSON endpoints.  No XML please.

>EDIT: the issue could also be around the mix of query string and POST body parameter sending.

Yes, I think this is the key remaining issue.  From [this comment](https://github.com/OneBusAway/onebusaway-application-modules/issues/126#issuecomment-94544761):

>Not to pick on anybody or defer the blame, but its quirky and against the HTML spec how the form POSTs currently work on the android app; they post a query string AND form values:
```
/api/where/report-problem-with-trip.json 500 ?app_ver=41&app_uid=<uid>&version=2&key=<key>D POST /api/where/report-problem-with-trip.json?app_ver=41&app_uid=<uid>&version=2&key=<key> HTTP/1.1
```

This should be fixed in OBA Android, no matter what the server implementation is doing.

An aside - the above comment also states that OBA iOS is using HTTP GET instead of POST?

(@aaronbrethorst, could you confirm that OBA iOS uses HTTP GET to send trip/stop problem reports?)

We could switch to GET as well, but that's not technically RESTful.  I'd prefer to fix the above POST issue first and see if that works.

barbeau:
This was automatically closed by Github as a result of deleting the develop branch - I'll re-open shortly in a new PR against master.

PR #280, opened 2015-08-05
Fix #278 - Changed alert inclusion logic to check if situation.activ…

### LINKED ISSUE #278: Active time windows of Alerts (i.e., Situations) aren't respected
Reported on the [OBA Dev list](https://groups.google.com/forum/#!topic/onebusaway-developers/GWiGsMzae94):

>I am having issue with old alerts still showing up on Android phone. Looking at the '/api/where/arrivals-and-departures-for-stop' there is 'data.references/situations.activeWindows' element but I do not find any code that checks these values on Android.

>Is there update / patch available for this?

If we can get a hold of a response that includes this info we can add support for it in OBA Android.

softrunit:
…eWindow includes now.

### DIFF
--- a/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ArrivalsListFragment.java
@@ -55,6 +55,7 @@ import com.joulespersecond.oba.provider.ObaContract;
 import com.joulespersecond.oba.request.ObaArrivalInfoResponse;
 import com.joulespersecond.seattlebusbot.util.MyTextUtils;
 import com.joulespersecond.seattlebusbot.util.UIHelp;
+import com.joulespersecond.oba.elements.ObaSituation.ActiveWindow;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


barbeau:
Thanks Mike!  I'll likely review fully tomorrow.  I'd like to see a test
added for this as well so I can convince myself this doesn't adversely
affect servers that don't have active windows, but I can knock that out.

onebusaway-android/src/main/java/com/joulespersecond/seattlebusbot/ArrivalsListFragment.java
```diff
@@ -891,10 +892,26 @@ private void refreshSituations(List<ObaSituation> situations) {
...
+            if(isActiveWindowForSituation(situation)) {
+                SituationAlert alert = new SituationAlert(situation);
+                mSituationAlerts.add(alert);
+            }
         }
         mAlertList.addAll(mSituationAlerts);
     }
 
+    private boolean isActiveWindowForSituation(ObaSituation situation) {
+        boolean isActiveWindowForSituation = situation.getActiveWindows().length == 0;
```
barbeau: This works for a single active window, but if there are multiple windows A and B, and the current time falls between A and B (but not within either of those windows), then this incorrectly returns true.

I'm going to open a new PR shortly that includes unit tests for these cases, including a test case for an agency that's not providing active windows in situations, and an alternate implementation that pass these tests.  So, stay tuned for another patch shortly...

barbeau: Sorry, just kidding.  I had a bug in the test data.  I'll blame it on lack of sleep ;).  This implementation works as intended, although I'm still going to push a new PR that includes the tests and refactoring of this to allow testing.  Thanks for putting this together @softrunit!

barbeau:
@softrunit I'm going to close this in favor of https://github.com/OneBusAway/onebusaway-android/pull/281, which I just opened.  https://github.com/OneBusAway/onebusaway-android/pull/281 includes your work here, but also adds unit tests for agencies that have and don't have active windows for situations.  Please take a look at https://github.com/OneBusAway/onebusaway-android/pull/281 and make sure it still works for you.

PR #295, opened 2015-09-20
Add trip details information

### LINKED ISSUE #120: Add "Vehicle Info" option to provide arrival list/map for a single vehicle
From a former OBA iPhone user:

> On the iPhone OneBusAway app, if you click on a specific bus's status, one of your choices is "Vehicle Info" which leads to a screen that shows the vehicle number (not the route number) and a list of times that the bus reached previous stops and is projected to reach future stops. This functionality is valuable for estimating when a specific bus -- say, the one my kid is on -- will reach a specific stop. I don't see how to access this functionality on the Android version.

If we want feature parity with OBA iPhone we should add this feature.

bendu:
Added a trip details for an individual trip accessible from the arrivals list view to resolve #120

showtripdetails
I replaced "Show Route Information" with "Show trip details", making the route information inaccessible from the arrivals list. The route information is still used when searching for a particular route number though.

obafeaturescreenshot
For the trip details, the times for stops the bus should have already passed are in bold, while the ones the bus has not visited are normal font weight.

realtime
Header when there's realtime schedule deviation/vehicle data.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsActivity.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (C) 2012-2015 Paul Watts (paulcwatts@gmail.com), University of South Florida,
+ * Benjamin Du (bendu@me.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.ui;
+
+import org.onebusaway.android.io.ObaAnalytics;
+import org.onebusaway.android.util.FragmentUtils;
+import org.onebusaway.android.util.UIHelp;
+
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.v4.app.FragmentManager;
+import android.support.v7.app.AppCompatActivity;
+import android.view.MenuItem;
+
+public class TripDetailsActivity extends AppCompatActivity {
+
+    public static class Builder {
+
+        private Context mContext;
+
+        private Intent mIntent;
+
+        public Builder(Context context, String tripId) {
+            mContext = context;
+            mIntent = new Intent(context, TripDetailsActivity.class);
+            mIntent.putExtra(TripDetailsListFragment.TRIP_ID, tripId);
+        }
+
+        public Builder setStopId(String stopId) {
+            mIntent.putExtra(TripDetailsListFragment.STOP_ID, stopId);
+            return this;
+        }
+
+        public Builder setUpMode(String mode) {
+            mIntent.putExtra(NavHelp.UP_MODE, mode);
+            return this;
+        }
+
+        public Intent getIntent() {
+            return mIntent;
+        }
+
+        public void start() {
+            mContext.startActivity(mIntent);
+        }
+    }
+
+    public static void start(Context context, String tripId) {
+        new Builder(context, tripId).start();
+    }
+
+    public static void start(Context context, String tripId, String stopId) {
+        new Builder(context, tripId).setStopId(stopId).start();
+    }
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        UIHelp.setupActionBar(this);
+
+        FragmentManager fm = getSupportFragmentManager();
+
+        if (fm.findFragmentById(android.R.id.content) == null) {
+            TripDetailsListFragment list = new TripDetailsListFragment();
+            list.setArguments(FragmentUtils.getIntentArgs(getIntent()));
+
+            fm.beginTransaction().add(android.R.id.content, list).commit();
+        }
+    }
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+        ObaAnalytics.reportActivityStart(this);
+    }
+
+    @Override
+    protected void onStop() {
+        super.onStop();
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == android.R.id.home) {
+            NavHelp.goUp(this);
+            return true;
+        }
+        return false;
+    }
+}
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
@@ -1,0 +1,497 @@
+/*
+ * Copyright (C) 2012-2015 Paul Watts (paulcwatts@gmail.com), University of South Florida,
+ * Benjamin Du (bendu@me.com), and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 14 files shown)


barbeau:
Thanks @bendu, this is great!  I'll likely review in the next few days.

barbeau:
Checked ICLA manually :+1:

onebusaway-android/src/main/res/layout/trip_details_head.xml
```diff
@@ -0,0 +1,68 @@
...
+                          android:layout_gravity="left"
+                          android:layout_width="fill_parent"
+                          android:layout_height="wrap_content">
+                <TextView
+                        android:id="@+id/agency"
+                        style="@style/TripDetailsAgency"
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content">
+                </TextView>
+                <TextView
```
barbeau: Can you please move the `vehicle` TextView to the following line, on its own line?  This gets really cramped for agencies that have longer names, and also use their name as the agency_id that prefaces the vehicle number:

image

onebusaway-android/src/main/res/values/styles.xml
```diff
@@ -196,4 +196,19 @@
         <item name="android:paddingLeft">5dp</item>
         <item name="android:paddingRight">5dp</item>
     </style>
+
+    <!-- Trip details -->
+    <style name="TripDetailsAgency">
+        <item name="android:textSize">12sp</item>
+        <item name="android:textStyle">bold</item>
+        <item name="android:textColor">@android:color/white</item>
```
barbeau: Can you please change these colors to `@color/header_text_color`?

The app can be white-labeled, in which case the theme (i.e., Action Bar) colors could change.  If this happens, we want to be able to change the text color of the Action Bar items in one place.  `@color/header_text_color` is currently `@android:color/white`, so this change won't have any impact on the current presentation.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
```diff
@@ -0,0 +1,352 @@
...
+            TextView time = (TextView) convertView.findViewById(R.id.time);
+            time.setText(DateUtils.formatDateTime(getActivity(),
+                    mStatus.getServiceDate() + stopTime.getArrivalTime() * 1000
+                            + mStatus.getScheduleDeviation() * 1000,
+                    DateUtils.FORMAT_SHOW_TIME |
+                            DateUtils.FORMAT_NO_NOON |
+                            DateUtils.FORMAT_NO_MIDNIGHT
+            ));
+
+            if (position < mNextStopIndex) {
```
barbeau: I like changing the styling of the item based on whether or not the bus has visited - that will be super useful.  However, I don't think the current styling difference of making the time bold/normal font is very obvious - its very easy to miss if you're not looking for it.

I'd prefer to see either the text color changed, or the background color changed for the stops that have already been visited.

I'd first try the text color change - any easy way to do this is to make the "passed" color transparent - for example, in the sliding panel header these two colors are used - one full white, and the other slightly transparent, to reduce emphasis on some of the lesser TextViews:

~~~
<color name="header_text_color">@android:color/white</color>
<color name="header_text_faded_color">#bbffffff</color>
~~~

On a related note, we should refresh this info occasionally - users will expect this information to update over time.  This should be fairly each to implement - you can following the existing model used in `ArrivalsListFragment`, and implement the same in `TripDetailsListFragment`.

You'll need the refresh handler:

~~~
private final Handler mRefreshHandler = new Handler();

    private final Runnable mRefresh = new Runnable() {
        public void run() {
            refresh();
        }
    };
~~~

...and a `refresh()` method:

~~~
private void refresh() {
        if (isAdded()) {
            UIHelp.showProgress(this, true);
            getArrivalsLoader().onContentChanged();
        }
    }
~~~

You'll also need to start the refresh in `onResume()`, and stop it in `onPause`.  Again, see `ArrivalsListFragment` for details.

And, we should also have a refresh icon in the Action Bar - see `menu/arrivals_list.xml` and `ArrivalsListFragment.onCreateOptionsMenu()` / `onOptionsItemSelected()` for how to implement.  

In the future (not for this PR), it would be good to add some kind of vehicle icon to show where the vehicle is add for each stop - we can tackle that later, though, when we do the other styling with the circles/lines for the route.

If you have any questions about any of this please let me know.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
```diff
@@ -0,0 +1,352 @@
...
+        setListAdapater(mAdapter);
+
+        if (mStopId != null) {
+            final int stopIndex = findIndexForStop(mTripInfo.getSchedule().getStopTimes(), mStopId);
+            if (stopIndex != -1) {
+                final ListView listView = getListView();
+                listView.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        listView.smoothScrollToPosition(stopIndex);
```
barbeau: The smooth scroll looks really strange to me - let's change to just setting the position to the `ListView` snaps to that position without scrolling.  IIRC, `listview.setSelection()` is the method that does this.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsActivity.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class TripDetailsActivity extends ActionBarActivity {
+
+    public static void start(Context context, String tripId, String stopId) {
+        context.startActivity(makeIntent(context, tripId, stopId));
+    }
+
+    public static Intent makeIntent(Context context, String tripId, String stopId) {
+        Intent myIntent = new Intent(context, TripDetailsActivity.class);
+        myIntent.setData(ObaContract.Trips.buildUri(tripId, stopId));
```
barbeau: Because you're not actually using the `URI`, and you're just parsing it back to `tripId and stopId` in `TripDetailsListFragment.onActivityCreated()`, I think I'd prefer to just have you pass the separate Strings in the `Intent`.  And, this can be set up as an `IntentBuilder` in `TripDetailsListFragment` - you can basically copy/paste the code from `ArrivalListFragment` for this:
https://github.com/OneBusAway/onebusaway-android/blob/develop/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java#L175

..and you'd put the `tripId`/`stopId` as a Strings in the bundle, like stop name and other fields are being handled in `ArrivalsListFragment`:
`mIntent.putExtra(ArrivalsListFragment.STOP_NAME, stopName);`

barbeau:
@bendu Thanks for all your work on this!  It looks good in general, some specific feedback inline.  Please let me know if you have questions on any of this.

onebusaway-android/src/main/res/layout/trip_details_listitem.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (C) 2010-2015 Paul Watts (paulcwatts@gmail.com)
```
barbeau: For new files, if you created them from scratch please change the header to read:
`<!-- Copyright (C) 2015 Your full name (youremail@email.com)`

If they were copy/pastes from elsewhere in the project, please update the header to use the previous copyright from that file, and add your name at the end:
`<!-- Copyright (C) 2010-2015 Paul Watts (paulcwatts@gmail.com), Your full name (youremail@email.com), and individual contributors`

I'm trying to do a better job of having the license header copyright match the history of contributors.

bendu: Should I add my name to files I modified as well, given that you want to keep track of contributors?

barbeau: @bendu Yes, thanks, please do.

bendu:
@barbeau I made changes to address your code review comments. I also updated my Activity to use AppCompatActivity

updatedfeaturescreenshot

barbeau:
Thanks @bendu!  I'll likely review in the next day or so.

barbeau:
@bendu Thanks for your work on this!

This looks good - it actually helped find a bug that exists in `ArrivalsListFragment`, too.

For `ArrivalsListFragment` and `TripDetailsListFragment`, tapping the refresh button seems to spawn a new refresh thread.  In other words, if the initial 60 sec refresh is scheduled at 4:00:01pm, and you hit the refresh button at 4:00:30pm, you'll get two refreshes, one at 4:01:01pm and one at 4:01:30pm.  The desired behavior would be that if you hit refresh at 4:00:30pm, you'd reset the cycle for every 60 second updates at that point, so the next update would be at 4:01:30pm

I'll go ahead and merge this and fix the refresh bug in both fragments.

Thanks again!

PR #299, opened 2015-09-23
Fix  #298 - Add analytics to the Sort By buttons

### LINKED ISSUE #298: Add analytics to the "Sort By" buttons, negative arrival setting
In `develop` branch, there are now "Sort By" buttons in the Action Bar for the `ArrivalsListFragment` (which sorts arrivals by route or ETA), as well as `MyStarredStopsFragment` (which sorts stops by name or frequency of use in the app).

We need to add Google Analytics to both of these.

@cagryInside Would you be able to work on this tomorrow morning so I can get a new alpha release out tomorrow?

Also, please note that the sort by arrivals is replacing the old analytics for the Style A / Style B preference.  The preference to change this is now replaced by the Sort By button in `ArrivalsListFragment`, and we're referring to Style A as "sort by ETA", and Style B as "sort by route".  So I think we should have two new analytics labels for these sort buttons, and forget about the Style A / Style B labels.

cagryInside:
New Analytics messages:

| app_settings  | general  | Show Departed Buses: YES |
| app_settings  | general  | Show Departed Buses: NO |
| ui_action  | button_press  | Clicked sort by ETA from ArrivalsListFragment |
| ui_action  | button_press  | Clicked sort by route from ArrivalsListFragment |
| ui_action  | button_press  | Clicked sort by name from MyStarredStopsFragment |
| ui_action  | button_press  | Clicked sort by most frequently used from MyStarredStopsFragment |

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -982,11 +982,19 @@ public class ArrivalsListFragment extends ListFragment
             Toast.makeText(getActivity(), r.getString(R.string.stop_info_option_sort_route),
                     Toast.LENGTH_SHORT).show();
             newValue = 1;
+            //Analytics
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getActivity().getString(R.string.analytics_action_button_press),
+                    getActivity().getString(R.string.analytics_label_sort_by_route_arrival));
         } else {
             // Currently we're sorting by route - change to sorting by eta
             newValue = 0;
             Toast.makeText(getActivity(), r.getString(R.string.stop_info_option_sort_eta),
                     Toast.LENGTH_SHORT).show();
+            //Analytics
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getActivity().getString(R.string.analytics_action_button_press),
+                    getActivity().getString(R.string.analytics_label_sort_by_eta_arrival));
         }
 
         PreferenceHelp.saveString(getResources()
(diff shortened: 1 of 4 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -982,11 +982,19 @@ private void doSortBy() {
             Toast.makeText(getActivity(), r.getString(R.string.stop_info_option_sort_route),
                     Toast.LENGTH_SHORT).show();
             newValue = 1;
+            //Analytics
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getActivity().getString(R.string.analytics_action_button_press),
+                    getActivity().getString(R.string.analytics_label_sort_by_eta_arrival));
```
barbeau: I believe this should be `analytics_label_sort_by_route_arrival` - it should match up with the `Toast` message.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -982,11 +982,19 @@ private void doSortBy() {
...
         } else {
             // Currently we're sorting by route - change to sorting by eta
             newValue = 0;
             Toast.makeText(getActivity(), r.getString(R.string.stop_info_option_sort_eta),
                     Toast.LENGTH_SHORT).show();
+            //Analytics
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getActivity().getString(R.string.analytics_action_button_press),
+                    getActivity().getString(R.string.analytics_label_sort_by_route_arrival));
```
barbeau: I believe this should be `analytics_label_sort_by_eta_arrival` - it should match up with the `Toast` message.

barbeau:
Thanks @cagryInside!  Mostly looks good.  One item - I think the sort by ETA/route analytics labels need to be switched - please take a look and let me know what you think.

cagryInside:
> I think the sort by ETA/route analytics labels need to be switched - please take a look and let me know what you think.

@barbeau Sorry for the simple mistake, now it should be fixed.

barbeau:
No worries!  Looks good now, thanks!

PR #301, opened 2015-09-24
Bump Travis ADB install timeout

barbeau:
We've been seeing a lot of build failures due to timeouts, so let's see if this fixes it.


barbeau:
This passed, so I'm going to merge and we can continue to monitor Travis builds.

PR #320, opened 2015-09-29
Restyle the trip details list as a line with dots, plus bus icon

barbeau:
I previously mentioned re-styling the trip details list in https://github.com/OneBusAway/onebusaway-android/issues/120#issuecomment-133772637, and so I wanted to open a PR just to document that this is complete to avoid potential duplication of effort by others.

This patch changes to using a more graphical representation of the route stop list, with the traditional dots for stops and a line connecting them for the route.  The real-time bus position is represented by an icon.

If an agency provides a GTFS `route_color`, it will use this color on the route stops/line and bus icons.  If not, it will use the default theme color.

Below is an example of GTFS `route_color` with HART data - most of their routes are blue (left), but their 400 BRT line is green (right):

image

Thanks again to @bendu for his work on trip details that made this possible!


barbeau:
@landonreed I was looking at some of the route colors that show up in Atlanta, in the context a new trip details screen that we just added to OBA Android - it uses the GTFS routes `route_color` for the line/stop/bus color.  Some of them look good:

image

...while others are kind of hard on the eyes:

image

Are the yellow and light blue above the exact actual colors used for the 165 and 6 routes in Atlanta, respectively?

bendu:
The colors do seem harsh. Would changing the opacity to a smaller value make the colors less harsh while maintaining the desired color? I'm also thinking of adding a border onto the line/dots to address this issue. I'll try playing around with this soon, although I think things are about to get busy for me elsewhere, so I might not be able to.

landonreed:
Honestly, I'm pretty sure all of the bus route colors were determined in
some kind of randomized fashion.  AFAIK, the only colors worth keeping on
MARTA's feed are the rail lines.  I'll add this to my list of GTFS
questions for MARTA.


---
Landon Reed  | http://landonreed.net
@eltiar <https://twitter.com/eltiar>

barbeau:
In `develop` I just changed the bus and real-time indicator colors to match our normal OBA red/blue/green status colors, to make this more visible and constant, no matter what the GTFS `route_color` is.  This also matches the status color background in the header, and ties into the real-time themes on the other screens in the app.  I also added a placemark icon to mark the current stop.

image


@bendu If we're going to use GTFS `route_color` for stops/line, I think we need to use exact what the agency provides (e.g., no changing opacity).  Adding a stroke (border) on the circle and line would probably help - feel free to give it a try if you have time.  Right now the placemark icon and route circles/lines are what are colored according to the GTFS `route_color`.  I'm open to possibly changing this as well - it might get a little color-crazy in cities like Atlanta that have a lot of different route colors, combined with the real-time status colors.

@landonreed That would be great if you can follow up with MARTA - if those aren't real route colors, they should be changed to the real ones.

PR #338, opened 2015-10-20
Fix #336 - Reduce GA data usage for develop branch

### LINKED ISSUE #336: Google Analytics Data Volume Exceeded Warning
OneBusAway is getting closer to exceed our free quota for Google Analytics.

OBA submits the application settings information every time the app starts. And, this causes a lot of hits in GA console.

screen shot 2015-10-19 at 11 01 10 am

cagryInside:
Some app information reporting was disabled on startup in order to reduce data usage.

### Disabled reporting fields:
| app_settings  | configured_region  | API Region: Tampa (or other regions) |
| app_settings  | general  | Show Experimental Regions: NO |
| app_settings  | general  | Show Experimental Regions: YES |
| app_settings  | general  | Set Region Automatically: NO |
| app_settings  | general  | Set Region Automatically: YES |

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -96,7 +96,6 @@ public class Application extends android.app.Application {
         initObaRegion();
 
         ObaAnalytics.initAnalytics(this);
-        reportAnalytics();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


barbeau:
@cagryInside How many events did the above fields generate over the last month?  I'd like to get a handle on how much of a dent the above changes will make in our GA overage.  For example, if we're 100M over per month, and the above only reduces our monthly event reporting by 1000, then we're probably better off just keeping these events and reducing the sampling rate.

barbeau:
FYI - This is closed as a result of deleting the develop branch - we're moving all our ongoing work back to master.  If we determine it's worth making the change, we'll need to rebase and open PR against master.

cagryInside:
> How many events did the above fields generate over the last month? I'd like to get a handle on how much of a dent the above changes will make in our GA overage. For example, if we're 100M over per month, and the above only reduces our monthly event reporting by 1000, then we're probably better off just keeping these events and reducing the sampling rate.

@barbeau Our data volume for last month (including IOS & Android) is `108M hits` and the below fields generate `41M hits`.

Event Category  | Event Action | Event Label | # of hits |
| ------------- | ------------- | ------------- | ------------- |
| app_settings  | configured_region  | API Region: Tampa (or other regions) |13,467,568 |
| app_settings  | general  | Show Experimental Regions: NO | 13,216,913 |
| app_settings  | general  | Show Experimental Regions: YES | 144,810 |
| app_settings  | general  | Set Region Automatically: NO | 8,462,043 |
| app_settings  | general  | Set Region Automatically: YES | 4,858,417 |

barbeau:
@cagryInside Thanks!  Ok, that's a good chunk.  IIRC, these are the events that are reported each time the app starts up, right?  If so, and we're going to move to sampling, the question is if these events are of equal importance to the other events reported within the app.  My feeling is no, they aren't - if not, we should strongly consider eliminating them for both platforms (Android and iOS), to keep sampling consistent across platforms, and to increase the number of data points we're getting from other events (assuming I understand GA sampling correctly).

bbodenmiller:
I believe this data is worth capturing but not in this method using events. I posted about moving this to another technique elsewhere so I'll have to find that comment.

bbodenmiller:
See https://github.com/OneBusAway/onebusaway-android/issues/336#issuecomment-149623543 for my suggestion. I can help implement but it may not be until late December.

PR #339, opened 2015-10-20
Fix arrivals list stop name while loading from the trip details list.

bendu:
While loading the arrivals list for a particular stop from the trip details view, the stop name is shown as "University Area Transit Center", when it should be shown as the proper name as the stop. The name gets corrected after the arrivals list is fully loaded, but this should eliminate any confusion that may occur on a slow network.


barbeau:
Thanks @bendu, looks good!

PR #347, opened 2015-10-29
Fix #327 successfully star route for all stops after removing star for a single stop

### LINKED ISSUE #327: Starring route for all stops after removing star for a single stop doesn't work
If you follow the below steps:

1. Go to a stop
2. Star a route for this stop
3. Unstar the route for this stop
4. Star the route for all stops

The route isn't starred for the current stop.  It should be starred.

bendu:
Only add exclusion when unstarring a single with "all stops" starred. Verified in Android Emulator


barbeau:
Thanks @bendu!  Great timing :).

PR #349, opened 2015-10-29
WIP - Initial work at allowing swipe down on sliding panel header to hide it

barbeau:
Doesn't work as currently implemented

**Please do not merge.**


barbeau:
I'm going to close this for now - we've implemented the back button press to hide the sliding panel in https://github.com/OneBusAway/onebusaway-android/issues/357, and that's consistent with how Google Maps and other apps with sliding panels work.  If someone wants to revive this and take another shot please feel free.

PR #359, opened 2015-11-15
clarify that app is for more than bus information

bbodenmiller:
Remove word "bus" from "Real-time bus arrival information for public transit" to clarify that app has more than just bus data.


barbeau:
Thanks!  I don't think that text had been touched in 5+ years :).

PR #362, opened 2015-11-15
update header depth in readme

bbodenmiller:
* These headers don't really belong under "Build Setup"
* Minor link improvement


barbeau:
Thanks!

PR #378, opened 2015-11-19
store amazon maps maven repo in git repo

bbodenmiller:
Store Amazon Maps Maven repo in git repo so that installing Maven and Amazon Maps files separately is not needed.


barbeau:
@bbodenmiller Thanks, this looks cool!  So if amazon provides an updated AAR and pom file for its library, what is the exact command to run to update the bundled mvn repo?  We should record that somewhere (probably under new subcategory under Readme "Build Setup" so it doesn't get lost.

onebusaway-android/build.gradle
```diff
@@ -272,7 +275,6 @@ tasks.whenTaskAdded { theTask ->
 }
 
 dependencies {
-    compile fileTree(dir: 'libs', include: '*.jar')
```
barbeau: This line is actually unrelated to the Amazon library - it includes any bundled JAR files in `onebusaway-android/libs`, which doesn't currently exist.  I'm fine with removing it, though, as its currently unused.

onebusaway-android/build.gradle
```diff
@@ -27,11 +27,14 @@ apply plugin: 'com.android.application'
 
 repositories {
     jcenter()
-    mavenLocal()
     maven {
         // OBA Releases - for comparator to sort alphanumeric routes
         url "http://nexus.onebusaway.org/nexus/content/repositories/releases"
```
bbodenmiller: Could we add the Amazon libraries to this Nexus repo so we could avoid them in the git repo all together?

bbodenmiller:
@barbeau added details about updating Amazon Maps files under Build Setup.

onebusaway-android/build.gradle
barbeau: I'd prefer not to - we've added 3rd party artifacts to the OBA nexus repo in the past, and its come back to bite us (https://github.com/OneBusAway/onebusaway-application-modules/issues/103).  I don't know why Amazon isn't hosting this on JCenter or Maven Central - I asked when I first set this up and they didn't know why either.  I'll check in with them again to see if they've made progress.

bbodenmiller: Got it. Hopefully Amazon can add it to JCenter or Maven Central since binaries in git isn't the best idea ever and so we wouldn't have to deal with doing there updates.

bbodenmiller: For the time being I think this is good though. Since this just moves the existing files it isn't actually bloating the repo at all.

barbeau: Yeah, this is a good improvement - it prevents the developer from having to do the mvn install themselves and simplifies the build setup.  It may result in performance improvements in Android Studio too - I recall a few issues where Android Studio indexes the entire local maven repo which results in huge files that Android Studio then has to parse.

barbeau:
Thanks @bbodenmiller!

PR #361, opened 2015-11-15
improve visibility of testing details

bbodenmiller:
* Move testing details closer to top of readme
* Change wording from "beta testing" to testing since there is alpha and
beta testing options

PR #384, opened 2015-11-25
improve visibility of testing details

bbodenmiller:
Replaces #361

* Move testing details closer to top of readme
* Change wording from "beta testing" to testing since there is alpha and
beta testing options
* Cleans up double spaces because they are not rendered as double in Markdown or HTML and are not needed because the font already makes the space the desired width :)


barbeau:
Thanks @bbodenmiller!

PR #360, opened 2015-11-15
cleanup build setup details

bbodenmiller:
Clarify and simplify build setup details

PR #385, opened 2015-11-25
cleanup build setup details

bbodenmiller:
Replaces #360

* clarify and simplify build setup details
* change all numbers to start with 1 since Markdown will correctly
render them as 1,2,3,... and it makes adding/changing the order easier


barbeau:
Thanks @bbodenmiller!

PR #392, opened 2015-12-01
Fix #187 - Issue reporting integration

### LINKED ISSUE #187: Improve issue reporting

cagryInside:
This is the issue reporting enhancement implementation as described in (https://github.com/OneBusAway/onebusaway-android/issues/187). 

List of enhancements:
* Open311 issue reporting specification (http://wiki.open311.org/GeoReport_v2/) was adopted and implemented as library project called [open311-client](https://github.com/CUTR-at-USF/open311-client).
* Open311-client library was used to implement new issue reporting enhancements in OBA.
* Integration with Open311-compliant reporting sites such as SeeClicFix.com.
* New issue reporting flow

Send Feedback section  |  Region Validation
:-------------------------:|:-------------------------:
device-2015-12-01-104856 | device-2015-12-01-104956

Send Feedback Categories  |  Contact Customer Service
:-------------------------:|:-------------------------:
device-2015-12-01-105040 | device-2015-12-01-105655

Infrastructure issue reporting 1   | Infrastructure issue reporting 2
:-------------------------:|:-------------------------:
device-2015-12-01-105156 | device-2015-12-01-105240

Infrastructure issue reporting 3   | Infrastructure issue reporting 4
:-------------------------:|:-------------------------:
device-2015-12-01-105308 | device-2015-12-01-105351

Infrastructure issue reporting 5   | Submit an issue
:-------------------------:|:-------------------------:
device-2015-12-01-105431 | device-2015-12-01-105546

Examples of how issue reports appear in SeeClickFix (Pinellas County's Open311 provider):

* http://seeclickfix.com/issues/2375570-psta-bus-stops
* http://seeclickfix.com/issues/2375567-psta-arrival-times-schedules)


barbeau:
I've written an official proposal to add `agency_email` to the GTFS spec here:
https://groups.google.com/forum/#!topic/gtfs-changes/aezjQsriLYA

cagryInside:
@barbeau This should be ready for code review. Latest trip problem results look like this:
```
test

---------------------
agency_name=PSTA;
stop_id=PSTA_5022;
trip_headsign=ST PETE BCH;
route_id=PSTA_SBT;
trip_id=PSTA_1327350010;
vehicle_id=PSTA_821;
service_date=1459656000000;
predicted=true;
vehicle_location=27.923952102661133 -82.8431396484375;
schedule_deviation=16.333333333333332 min;
stop_arrival_time=10:39 PM;
stop_departure_time=10:39 PM; 
```

Link: test.seeclickfix.com/issues/1315850-traffic-sign-or-signal-issue

barbeau:
@cagryInside Awesome, thanks!

barbeau:
@cagryInside A few requests for the trip problem results:

* Could you add a human readable date to `service_date`, so it looks something like this:

`service_date=1459656000000 (Aug 15, 2016);`

* Could you limit schedule deviation to 3 decimal digits, something like:

`schedule_deviation=16.333 min;`

* Could you add `route_short_name` under `route_id`, something like:

`route_id=PSTA_SBT;
`route_short_name=SBT;`

If `route_short_name` hasn't been populated in GTFS, then we should show `route_long_name` instead.

cagryInside:
cc @barbeau 
Trip reporting when predicted = true :
```
agency_name=PSTA;
stop_id=PSTA_963;
trip_headsign=NORTH VIA 116 AVE;
route_id=PSTA_4;
route_short_name=4;
trip_id=PSTA_1334930020;
vehicle_id=PSTA_2609;
service_date=04-04-2016;
predicted=true;
vehicle_location=27.838199615478516 -82.6385269165039;
schedule_deviation=2.867 min;
stop_arrival_time=01:21 PM;
stop_departure_time=01:21 PM; 

```

Trip reporting when predicted = false :
```
agency_name=PSTA;
stop_id=PSTA_2394;
trip_headsign=SEMINOLE MALL;
route_id=PSTA_58;
route_short_name=58;
trip_id=PSTA_1323787020;
service_date=04-04-2016;
predicted=false;
stop_arrival_time=01:27 PM;
stop_departure_time=01:27 PM; 

```

PS: I am currently pushing all the new modifications with separate commits. After, we finish code reviewing, I will squash them into one commit.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/Open311ProblemFragment.java
```diff
@@ -492,16 +498,27 @@ private String getTransitIssueParameters(Service service) {
...
+
                 sb.append(getResources().getString(R.string.ri_append_trip_name, mArrivalInfo.getHeadsign()));
                 sb.append(getResources().getString(R.string.ri_append_route_id, mArrivalInfo.getRouteId()));
+                if (!TextUtils.isEmpty(mArrivalInfo.getShortName())) {
+                    sb.append(getResources().getString(R.string.ri_append_route_short_name,
+                            mArrivalInfo.getShortName()));
+                } else if (!TextUtils.isEmpty(mArrivalInfo.getRouteLongName())) {
+                    sb.append(getResources().getString(R.string.ri_append_route_long_name,
+                            mArrivalInfo.getRouteLongName()));
+                }
```
cagryInside: @barbeau So, first I put short_name if it is available. Otherwise, I put long_name.

barbeau: Do you have a reference to `ObaRoute` here?  If so, let's use the `UIUtils.getRouteDisplayName(ObaRoute route)` method for this - this is used elsewhere in the app.  If not, what you have is fine.

barbeau: Thinking more about this - it's probably better to create a new UIUtils method that takes in an ArrivalInfo object, and then refactor UIUtils.getRouteDisplayName(ObaRoute route) and this new method to both use another new method UIUtils.getRouteDisplayName(String routeShortName, String routeLongName), and returns the correct name.  This way you don't have to pull out a ObaRoute object when you already have the same info in the ArrivalInfo object.

onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
```diff
@@ -169,19 +170,25 @@ public static final int getStopDirectionText(String direction) {
...
     }
 
+    public static final String getRouteDisplayName(ObaRoute route) {
+        return getRouteDisplayName(route.getShortName(), route.getLongName());
+    }
+
+    public static final String getRouteDisplayName(ObaArrivalInfo arrivalInfo) {
+        return getRouteDisplayName(arrivalInfo.getShortName(), arrivalInfo.getRouteLongName());
+    }
+
```
cagryInside: @barbeau just added `getRouteDisplayName` to `UIUtils`.

barbeau: Nice, thanks!

onebusaway-android/build.gradle
```diff
@@ -298,5 +298,6 @@ dependencies {
     // Google Play Services Maps (only for Google flavor)
     googleCompile 'com.google.android.gms:play-services-maps:8.3.0'
     // Amazon Maps (only for Amazon flavor)
-    amazonCompile 'com.amazon.android:amazon-maps-api:2.0'
+    amazonCompile('com.amazon.android:amazon-maps-api:2.0')
+    compile 'edu.usf.cutr:open311client:1.0.6'
```
barbeau: Could you please move this open311 library above the libraries that are specific to a build variant?  I'd like to keep all the ones required for all variants in the same section of `build.gradle`.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -35,6 +35,7 @@
 import org.onebusaway.android.io.elements.ObaRoute;
 import org.onebusaway.android.io.elements.ObaShape;
 import org.onebusaway.android.io.elements.ObaStop;
+import org.onebusaway.android.io.elements.ObaStopElement;
```
barbeau: I believe this is an extra unneeded import?  If so, could you please remove it?

onebusaway-android/src/main/AndroidManifest.xml
```diff
@@ -145,7 +149,7 @@
                     android:name="android.support.PARENT_ACTIVITY"
                     android:value="org.onebusaway.android.ui.HomeActivity"/>
         </activity>
-        <activity android:name="org.onebusaway.android.ui.ReportStopProblemActivity"
+        <activity android:name=".report.ui.ReportStopProblemActivity"
```
barbeau: Let's keep the full package name here, just for consistency with the other activities.

onebusaway-android/src/main/AndroidManifest.xml
```diff
@@ -153,7 +157,7 @@
                     android:name="android.support.PARENT_ACTIVITY"
                     android:value="org.onebusaway.android.ui.HomeActivity"/>
         </activity>
-        <activity android:name="org.onebusaway.android.ui.ReportTripProblemActivity"
+        <activity android:name=".report.ui.ReportTripProblemActivity"
```
barbeau: Let's keep the full package name here, just for consistency with the other activities.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
```diff
@@ -256,6 +305,7 @@ public String toString() {
                 ", obaBaseUrl='" + obaBaseUrl + '\'' +
                 ", siriBaseUrl='" + siriBaseUrl + '\'' +
                 ", bounds=" + Arrays.toString(bounds) +
+                ", open311Servers=" + Arrays.toString(open311Servers) +
```
barbeau: I believe there should be a ` + '\''` here as well?

barbeau: or is that output by `Arrays.toString()`?

onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -122,4 +125,28 @@
 
     <!-- General strings -->
     <string name="http_prefix">http://</string>
+
+    <!-- Problem reporting additions -->
+    <string name="ri_append_start">\n\n---------------------\n</string>
+    <string name="ri_append_stop_id">stop_id=%1$s;\n</string>
```
barbeau: Let's change this to `gtfs_stop_id`, so we don't confuse people reviewing the issue when `stop_id` is different from `stop_code`.

onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -122,4 +125,28 @@
...
+    <string name="ri_append_route_id">route_id=%1$s;\n</string>
+    <string name="ri_append_route_display_name">route_display_name=%1$s;\n</string>
+    <string name="ri_append_trip_name">trip_headsign=%1$s;\n</string>
+    <string name="ri_append_vehicle_id">vehicle_id=%1$s;\n</string>
+    <string name="ri_append_vehicle_location">vehicle_location=%1$s;\n</string>
+    <string name="ri_append_service_date">service_date=%1$s;\n</string>
+    <string name="ri_append_predicted">predicted=%1$s;\n</string>
+    <string name="ri_append_arrival_time">stop_arrival_time=%1$s;\n</string>
+    <string name="ri_append_departure_time">stop_departure_time=%1$s;\n</string>
+    <string name="ri_append_schedule_deviation">schedule_deviation=%1$s min;\n</string>
```
barbeau: I'm wondering if we should add some kind of label here for "early" or "late".

What if we drop the negative sign, and instead have:

* `schedule_deviation=1.550 min early;` // If sign was negative
* `schedule_deviation=1.550 min late;`  // If sign was positive
* `schedule_deviation=0 min;` // If zero (i.e., on time)

This makes it more human readable, and if we're going to parse it we could still convert back from early/late to negative/positive.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
cagryInside: @barbeau In ```BaseMapFragment.java ``` there are other unused variables: `CAMERA_MOVE_NOTIFY_THRESHOLD_MS`,  `mStopUserMap`, `mFocusStop`, `mFocusStopRoutes`. Do you want me to clean them up, or are we planning to use them in the future?

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
cagryInside: Yes it seems we do not need that character for arrays:
screen shot 2016-04-06 at 10 10 19 pm

However, I don't think `Arrays.tostring` works correctly. It prints the memory address of an array.

onebusaway-android/src/main/res/values/donottranslate.xml
cagryInside: It look like this now for each case: 
```
schedule_deviation=5.333 min late;
schedule_deviation=1.167 min early;
schedule_deviation=0 min;
```

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
barbeau: Good catch - let's go ahead and remove those if they are unused.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
barbeau: Good catch - I think you need to override `Bounds.toString()` and `Open311Servers.toString()`.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
```diff
@@ -72,6 +72,45 @@ public double getLonSpan() {
         }
     }
 
+    public static class Open311Servers implements ObaRegion.Open311Servers {
```
barbeau: If this object represents a single Open311 server, it should be `Open311Server`.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegion.java
```diff
@@ -35,6 +35,15 @@
         public double getLonSpan();
     }
 
+    public interface Open311Servers {
```
barbeau: If this interface represents a single Open311 server, it should be `Open311Server`.

onebusaway-android/src/main/res/values/donottranslate.xml
barbeau: :+1:

onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -37,12 +37,17 @@
     <string name="preference_key_tutorial">preference_key_tutorial</string>
     <string name="preference_key_show_header_arrivals">preference_key_show_header_arrivals</string>
 
-    <!-- Regions API URL -->
-    <string name="regions_api_url">http://regions.onebusaway.org/regions-v3.json</string>
 
+    <string name="regions_api_url">http://regions.onebusaway.org/regions-v3.json</string>
+<!-- Regions API URL
```
barbeau: Please go ahead and remove this commented out URL.

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaRegionElement.java
cagryInside: Yes, not it look like this with `tostring` implementation:
```
ObaRegionElement{id=0, regionName='Tampa Bay', active=true, obaBaseUrl='http://api.tampa.onebusaway.org/api/', siriBaseUrl='http://tampa.onebusaway.org/onebusaway-api-webapp/siri/', bounds=[[lat=27.976910500000002,lon=-82.445851,latSpan=0.5424609999999994,lonSpan=0.576357999999999], [lat=27.919249999999998,
lon=-82.652145,latSpan=0.47208000000000183,lonSpan=0.3967700000000036]], 
open311Servers=[[jurisdictionId=,apiKey=937033cad3054ec58a1a8156dcdd6ad8a416af2f,
baseUrl=http://test.seeclickfix.com/open311/v2/], [jurisdictionId=,apiKey=937033cad3054ec58a1a8156dcdd6ad8a416af2f,
baseUrl=http://www.publicstuff.com/api/open311/]], language='en_US',
 contactEmail='onebusaway@gohart.org', supportsObaDiscoveryApis=true,
 supportsObaRealtimeApis=true, supportsSiriRealtimeApis=true,
 twitterUrl='http://mobile.twitter.com/OBA_tampa', experimental=false, stopInfoUrl='null'}
```

cagryInside:
@barbeau Commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/7ffaf5d0f38fbd5471c70a2c13978eabf53846fe fixes the comments you mentioned.

onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -122,4 +125,28 @@
...
+    <string name="ri_append_vehicle_id">vehicle_id=%1$s;\n</string>
+    <string name="ri_append_vehicle_location">vehicle_location=%1$s;\n</string>
+    <string name="ri_append_service_date">service_date=%1$s;\n</string>
+    <string name="ri_append_predicted">predicted=%1$s;\n</string>
+    <string name="ri_append_arrival_time">stop_arrival_time=%1$s;\n</string>
+    <string name="ri_append_departure_time">stop_departure_time=%1$s;\n</string>
+    <string name="ri_append_schedule_deviation">schedule_deviation=%1$s min;\n</string>
+    <string name="ri_static_user_name">OneBusAway</string>
+    <string name="ri_static_user_last_name">User</string>
+    <string name="ri_static_user_email">test@oba.com</string>
```
barbeau: I just created the email account `issue-report@onebusaway.org`.  Let's use this email here.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -549,6 +543,47 @@
...
+    <string name="report_dialog_out_of_region_message"> Please select a bus stop.</string>
+    <string name="report_dialog_trip"> Please select a trip.</string>
+
+    <!-- report issue screen-->
+    <string name="ri_desc">Issue Description</string>
+    <string name="ri_button_camera">Camera</string>
+    <string name="ri_button_image">ADD IMAGE</string>
+    <string name="ri_button_gallery">Gallery</string>
+    <string name="ri_address_not_found">Address not found. Please check your address.</string>
+    <string name="ri_location_problem_info">Error getting location, please tap to map</string>
```
barbeau: Should be "please tap on map".

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -549,6 +543,47 @@
...
+    <string name="ri_location_problem">Error getting location</string>
+    <string name="ri_service_description_problem">Service description could not found.</string>
```
barbeau: I think this should be "Service description could not be found."

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -549,6 +543,47 @@
...
+    <string name="ri_service_default">Choose an issue</string>
+    <string name="ri_service_stop">Stop Problem</string>
+    <string name="ri_service_trip">Trip Problem</string>
+    <string name="ri_user_name">Name</string>
```
barbeau: I haven't looked at this in the UI yet, but should one of these be first name, and the other last name?

onebusaway-android/src/main/res/xml/app_tracker.xml
```diff
@@ -15,7 +15,6 @@
             type="integer">120
     </item>
 
-    <string name="ga_sampleFrequency">90</string>
```
barbeau: Is this string no longer used?

onebusaway-android/src/main/res/xml/global_tracker.xml
```diff
@@ -14,7 +14,6 @@
             type="integer">120
     </item>
 
-    <string name="ga_sampleFrequency">90</string>
```
barbeau: Is this string no longer used?

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/Open311ProblemFragment.java
```diff
@@ -0,0 +1,1070 @@
...
+                getResources().getColor(R.color.material_gray));
+        ((ImageView) findViewById(R.id.ri_ic_image_picker)).setColorFilter(
+                getResources().getColor(R.color.material_gray));
+        ((ImageView) findViewById(R.id.ri_ic_username)).setColorFilter(
+                getResources().getColor(R.color.material_gray));
+        ((ImageView) findViewById(R.id.ri_ic_customer_service_email)).setColorFilter(
+                getResources().getColor(R.color.material_gray));
+        ((ImageView) findViewById(R.id.ri_ic_customer_service_phone)).setColorFilter(
+                getResources().getColor(R.color.material_gray));
+        ((ImageView) findViewById(R.id.ri_ic_anonymous)).setColorFilter(
```
barbeau: I think the anonymous icon wasn't an official Google material design icon, but was from https://materialdesignicons.com/?  I think https://materialdesignicons.com/?  If so we need to add Christopher Schreiner to the About screen under the icon credits section.  If any of the other icons also aren't from Google, we need to do the same for them too.

onebusaway-android/src/main/res/xml/app_tracker.xml
cagryInside: It is still used. I deleted the redundant one. The original one is in line 20 https://github.com/OneBusAway/onebusaway-android/pull/392/files/66091e1e292b1a5e0bd2c39c82ea1e9c0d179b90#diff-e4da1a00f624408d4a24907a01d92bb6L21

onebusaway-android/src/main/res/xml/global_tracker.xml
cagryInside: It is still used. I deleted the redundant one. The original one is in line 20  https://github.com/OneBusAway/onebusaway-android/pull/392/files/66091e1e292b1a5e0bd2c39c82ea1e9c0d179b90#diff-e4da1a00f624408d4a24907a01d92bb6L21

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/Open311ProblemFragment.java
cagryInside: Yes, I have to icons from https://materialdesignicons.com/. Incognito and stop id flag. I will put the names.

cagryInside: @barbeau What does `MK` and `SE` means in the `about_text`.  I am trying to put the new names.
```
        "* Arrow by Zlatko Najdenovski, MK\n"
        "* Map Marker by Housin Aziz, SE\n"
```

barbeau: Not sure :).  I think this was just part of their names as they had them listed on materialdesignicons.com or thenounproject.  As long as you have the person's full name it should be fine.

onebusaway-android/build.gradle
```diff
@@ -298,5 +300,5 @@ dependencies {
     // Google Play Services Maps (only for Google flavor)
     googleCompile 'com.google.android.gms:play-services-maps:8.3.0'
     // Amazon Maps (only for Amazon flavor)
-    amazonCompile 'com.amazon.android:amazon-maps-api:2.0'
+    amazonCompile('com.amazon.android:amazon-maps-api:2.0')
```
barbeau: Could you remove the parentheses here, just for consistency?

cagryInside:
Commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/cffb474037ef4838ba7685ae27c8b0b1816f8490 fixes the text changes mentioned in comments.

barbeau:
For the "Are you in Tampa?" prompt after tapping "Send feedback" in navigation drawer, let's only ask that the first time the user wants to send feedback.  In a shared preference, we can store that the user confirmed, and store the regionId when they confirmed.  If the region changes (automatically or via user manually setting it via Settings), then we should ask again when they tap "Send feedback".

barbeau:
For the Send feedback list:

image

...let's match the style/color of the first/second text lines to the Starred Stop, My Reminders, and `TripDetailsListFragment` text - I believe the sub-text (second line) above is black, where the others are more of a gray:

image

barbeau:
Looks like the "Report a Bus Stop Problem" list item has the wrong icon:

image

We can use the same flag icon we're using for "Bus Stop ID" within the problem report UI itself, or I'm open to suggestions if you have a better idea.

barbeau:
For the "Send App Feedback" issue category, this should send the email to `android-app@onebusaway.org` instead of the regional contact email address.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -549,6 +543,47 @@
...
+    <string name="rt_stop_problem">Report a Bus Stop Problem</string>
+    <string name="rt_arrival_problem">Report Arrival Time Problem</string>
+    <string name="rt_app_feedback">Send App Feedback</string>
+    <string name="rt_ideas">Ideas For New Features</string>
+    <string name="rt_message_infrastructure_problem">Missing bus stop, broken bench, etc.</string>
+    <string name="rt_message_customer_service">Driver compliments and complaints, lost &amp; found, etc.</string>
+    <string name="rt_message_stop_problem">Wrong bus stop, etc.</string>
+    <string name="rt_message_arrival_problem">Late bus, early bus, etc.</string>
+    <string name="rt_message_app_feedback">App crash, bugs.</string>
+    <string name="rt_message_ideas">New ideas for future improvements.</string>
```
barbeau: Let's remove the `.` after "improvements".

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -549,6 +543,47 @@
...
+    <string name="rt_customer_service">Contact Customer Service</string>
+    <string name="rt_stop_problem">Report a Bus Stop Problem</string>
+    <string name="rt_arrival_problem">Report Arrival Time Problem</string>
+    <string name="rt_app_feedback">Send App Feedback</string>
+    <string name="rt_ideas">Ideas For New Features</string>
+    <string name="rt_message_infrastructure_problem">Missing bus stop, broken bench, etc.</string>
+    <string name="rt_message_customer_service">Driver compliments and complaints, lost &amp; found, etc.</string>
+    <string name="rt_message_stop_problem">Wrong bus stop, etc.</string>
+    <string name="rt_message_arrival_problem">Late bus, early bus, etc.</string>
+    <string name="rt_message_app_feedback">App crash, bugs.</string>
```
barbeau: Let's change this to `App crash, bugs, etc.` to match the other lists (i.e., add the `, etc.`).

barbeau:
Let's match the action bar title in the problem reporting screen to the problem type we show in the "Send feedback" list.

image


For example, when choosing "Report a Bus Stop Problem," the action bar title is "Infrastructure Issue":

image

Let's change this to "Bus Stop Problem".

When choosing "Report an Arrival Time Problem", the action bar title is also "Infrastructure Issue":

image

Let's change this to "Arrival Time Problem".

We should also change the Action Bar title when the user changes from "Stop Problem" to "Trip problem" (or it's Open311 equivalent) in the drop down box under the stop address.

barbeau:
@cagryInside This all looks great, by the way :) - lot's of complex logic, just a few items to polish up.  Thanks for implementing this!!  I have a few more comments I'll add shortly.

barbeau:
@cagryInside Could you please do a double-check and make sure we have everything instrumented with Google Analytics?

I'd also like to try and capture the issue problem types and categories that users are selecting (e.g., stop vs. trip problems, what type of stop problems, what type of trip problems).  Currently, we need to ask OBA regions for a dump of their MySQL databases to count the number of each types of issue reported.  It would be great if we could capture this centrally in Google Analytics - different regions will obviously have different categories (in the case of Open311), but it would be good to know.  Ideally we'd want to be able to run reports to determine how many transit vs. non-transit problems are submitted as well.

barbeau:
We should probably base the examples for problems shown in the "Send feedback" list based on the most frequently reported issues - we did a database dump of issue categories for Puget Sound, Tampa (just HART), and Atlanta - results are at https://github.com/OneBusAway/onebusaway-android/issues/187#issuecomment-207443382.

Looks like there are the top trip problem categories for these 3 regions:

1. `vehicle_never_came`
1. `vehicle_came_late`
1. `vehicle_came_early` 

Top stop problem categories, again for all 3 regions):
1. `Other` (I wonder if these are all infrastructure related?? The current OBA issue categories for stop are all data related)
1. `route_or_trip_missing`
1. `stop_name_wrong`

It's interesting that all 3 regions have the top 3 issue categories being reported in each region.  The order differs slightly (in Atlanta, `vehicle_never_came` is 2nd, while it's first in Tampa and Puget Sound), but the top 3 are the same.

@cagryInside Let's discuss this this afternoon.  The one caveat with the stop problems is that via Open311, we're now collecting infrastructure issues as well as data issues, so we should consider that.

cagryInside:
> For the Send feedback list:
> 
> image
> 
> ...let's match the style/color of the first/second text lines to the Starred Stop, My Reminders, and TripDetailsListFragment text - I believe the sub-text (second line) above is black, where the others are more of a gray:

Now it looks like this:

device-2016-04-08-112659

cagryInside:
> Looks like the "Report a Bus Stop Problem" list item has the wrong icon:

Flag icon looks like this for `stop problem`. I think it is good, I couldn't find a better option.

 
device-2016-04-08-113110

cagryInside:
> For the "Send App Feedback" issue category, this should send the email to android-app@onebusaway.org instead of the regional contact email address.

@barbeau Send app feedback email address is coming from the selected region:
```
UIUtils.sendEmail(getActivity(), region.getContactEmail(), locationString);
```
Should I read this one from static string?

barbeau:
>Send app feedback email address - Should I read this one from static string?

Yes, I think this should be pulled from `donottranslate.xml`.

cagryInside:
Commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/0e80f621662275fee3daf0d8a1844a0fec14f48c uses the new version of the open311-client (https://github.com/CUTR-at-USF/open311-client/commit/5de7d4127dd3efb49cb63bea9e1f41fbc21b611e) that fixes the parse crashing issue when there is an unknown property in json data.

barbeau:
Looks like there is a small alignment issue with the new "Report a Stop Problem" icon - the title and subtitle are shifted left slight:

image

Actually, it could be that the stop icon itself is shifted left (it doesn't look centered under the headphones?)- I haven't looked at the layout.

barbeau:
@cagryInside when you fix the above, could you please squash the commits as well?  Thanks!

cagryInside:
> Looks like there is a small alignment issue with the new "Report a Stop Problem" icon - the title and subtitle are shifted left slight:
> 
> image
> 
> Actually, it could be that the stop icon itself is shifted left (it doesn't look centered under the headphones?) - I haven't looked at the layout.


I accidentally put `mdpi` image to `hdpi` folder. That's why it was getting the smaller icon. Now it fixed.

device-2016-04-10-214408

barbeau:
@cagryInside Just wanted to check in on the remaining items that still need to be completed here - I think checking Google Analytics is the final item to review?

cagryInside:
> @cagryInside Just wanted to check in on the remaining items that still need to be completed here - I think checking Google Analytics is the final item to review?

Now we are submitting an event on a problem submission. This is the commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/9a281a3cd151e41827ab328c74dbb329d8e881be for Google Analytics.

barbeau:
@cagryInside Ok, thanks - so this would track non-Open311 submissions too?

Also, I just looked at the `ReportTypeListFragment`, and I don't believe we have analytics to know which report type the user selected ("Customer Service", vs. "Send app feedback", vs. "Stop problem", vs. "IdeaScale".  Could you add analytics to this as well?

For our project, we want to make sure that we're able to tell exactly how users are sending feedback, and where they are sending it to.

barbeau:
Sorry, one more item as well - let's change the order of adding fields to the Open311 description fields a bit to group similar fields:

~~~
* service_date
* agency_name
* stop_id
* route_id
* route_display_name
* trip_id
* trip_headsign
* predicted
* vehicle_id
* vehicle_location
* schedule_deviation
* stop_arrival_time
* stop_departure_time
~~~

cagryInside:
> @cagryInside Ok, thanks - so this would track non-Open311 submissions too?

Non-open311 submissions were tracked in `ReportStopProblemFragment` and `ReportTripProblemFragment` (https://github.com/OneBusAway/onebusaway-android/pull/392/commits/31e6be6962ceba466a8b0dd8e768659d4fad59d8#diff-efe35209cc2a30d4303bb197ed8a4d28R203).

I just added analytic to each problem type (https://github.com/OneBusAway/onebusaway-android/pull/392/commits/4672207ad9115c214e5d02e4306995cb85ba7618). 

But, I used 'UI_ACTION' instead of 'SUBMIT' in the event category.

```
 ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
                   getString(R.string.analytics_action_problem),                  
                   getString(R.string.analytics_label_stop_problem));
```

cagryInside:
> Sorry, one more item as well - let's change the order of adding fields to the Open311 description fields a bit to group similar fields:
> 
> * service_date
> * agency_name
> * stop_id
> * route_id
> * route_display_name
> * trip_id
> * trip_headsign
> * predicted
> * vehicle_id
> * vehicle_location
> * schedule_deviation
> * stop_arrival_time
> * stop_departure_time


Np, now it look like this: http://test.seeclickfix.com/pinellas_county

```
service_date=04-11-2016;
agency_name=PSTA;
gtfs_stop_id=PSTA_2683;
route_id=PSTA_19;
route_display_name=19;
trip_id=PSTA_1323037020;
trip_headsign=US 19 NORTH TARPON SPRING;
predicted=true;
vehicle_id=PSTA_12108;
vehicle_location=27.844192504882813 -82.68711853027344;
schedule_deviation=5.450 min late;
stop_arrival_time=11:40 AM;
stop_departure_time=11:40 AM; 
```

cagryInside:
I found a UI problem with older versions of Android (version 2.X). `FrameLayout` wasn't see margins of the views. This was an official bug for Android: https://code.google.com/p/android/issues/detail?id=28057
The UI was looking like this:

img_20160411_094539

So, I applied the solution mentioned here: http://stackoverflow.com/questions/5401952/framelayout-margin-not-working?answertab=active#tab-top in this commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/5fec7454e49b16c4db52719f993b6dafbb5ca2e4.

barbeau:
@cagryInside That solution for 2.x looks good to me.

One more item I just caught:
* In "Help->Contact us", could you please make this have the same behavior as "Send feedback"?  I thought about removing "Contact us" completely, but if people are used to using this, I don't want them to think we yanked the feature completely.  Directing them to the same place as "Send feedback" should be a gentle re-direction.

barbeau:
Also, please leave the same analytics reporting in place for "Help->Contact Us", so we can see if the usage of this drops.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportTypeListFragment.java
```diff
@@ -0,0 +1,162 @@
...
+                    getString(R.string.analytics_action_problem),
+                    getString(R.string.analytics_label_app_feedback));
+        } else if (getString(R.string.rt_ideas).equals(rti.getTitle())) {
+            // Direct to ideascale website
+            goToIdeaScale();
+
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getString(R.string.analytics_action_problem),
+                    getString(R.string.analytics_label_idea_scale));
+        } else if (getString(R.string.rt_customer_service).equals(rti.getTitle())) {
```
barbeau: Could you please re-order the IF statement so it matches the order of the items in the UI?  Makes it a little easier to read.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/dialog/CustomerServiceDialog.java
```diff
@@ -0,0 +1,168 @@
...
+            ImageButton emailButton = (ImageButton) view.findViewById(R.id.ricsi_email);
+            emailButton.setColorFilter(getActivity().getResources().
+                    getColor(R.color.navdrawer_icon_tint_selected));
+
+            if (TextUtils.isEmpty(agency.getEmail())) {
+                emailButton.setVisibility(View.INVISIBLE);
+            } else {
+                emailButton.setOnClickListener(new View.OnClickListener() {
+                    @Override
+                    public void onClick(View view) {
```
barbeau: Very important - could you please add analytics to each of the customer service contact methods (email, phone, website)?  We'd definitely want to know how the user is reaching out to the agency, as it will be hard to get metrics about this from the agency itself.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportTypeListFragment.java
```diff
@@ -0,0 +1,162 @@
...
+            ((ReportActivity) getActivity()).createInfrastructureIssueActivity(
+                    getString(R.string.ri_selected_service_trip));
+
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getString(R.string.analytics_action_problem),
+                    getString(R.string.analytics_label_trip_problem));
+        } else if (getString(R.string.rt_app_feedback).equals(rti.getTitle())) {
+            // Send App feedback
+            String email = getString(R.string.ri_app_feedback_email);
+            UIUtils.sendEmail(getActivity(), email, locationString);
```
barbeau: We just talked about this, but I'll add a comment here too just so its official :) - it looks like this location is always null.  I believe we should be grabbing the location from the Application class here - checkout the current "Help->Contact Us" process for getting location, and we can just imitate that here.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportTypeListFragment.java
```diff
@@ -0,0 +1,162 @@
...
+
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getString(R.string.analytics_action_problem),
+                    getString(R.string.analytics_label_stop_problem));
+        } else if (getString(R.string.rt_arrival_problem).equals(rti.getTitle())) {
+            // Report bus stop issue
+            ((ReportActivity) getActivity()).createInfrastructureIssueActivity(
+                    getString(R.string.ri_selected_service_trip));
+
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
```
barbeau: Can you please double-check that these analytics events fire when they are called after opening the new activity, or firing the intent for the IdeaScale website?  They should, but I would just like to confirm.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/dialog/CustomerServiceDialog.java
cagryInside: @barbeau Added the following:

| Event Category  | Event Action | Event Label |
| ui_action  | [AGENCY_NAME]_customer_service  | Open web page |
| ui_action  | [AGENCY_NAME]_customer_service  | Send email |
| ui_action  | [AGENCY_NAME]_customer_service  | Make phone call |

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportTypeListFragment.java
cagryInside: untitled 2

cagryInside:
@barbeau commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/a0bf9519f5c2a4e810051573e7f8a867610f5b00 fixes the null location problem and adds more analytics

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -976,10 +976,10 @@ private void goToContactEmail(Context ctxt) {
     private void goToSendFeedBack() {
         if (mFocusedStop != null) {
             ReportActivity.start(this, mFocusedStopId, mFocusedStop.getName(),
-                    mFocusedStop.getLatitude(), mFocusedStop.getLongitude());
+                    mFocusedStop.getLatitude(), mFocusedStop.getLongitude(), mGoogleApiClient);
         } else {
             Location loc = Application.getLastKnownLocation(this, mGoogleApiClient);
```
barbeau: This `loc` can be `null` - can you please update this to protect against an NPE?  I believe this means you'll need to allow for the situation where the lat and long aren't available within `ReportActivity`.  This would also apply to `HomeActivity.goToSendFeedBack()`.

onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
```diff
@@ -428,13 +428,22 @@ public static void sendEmail(Context context, String email, String location) {
...
+                    Build.VERSION.RELEASE,
+                    Build.VERSION.SDK_INT,
+                    location);
+        } else {
+            body = context.getString(R.string.bug_report_body_without_location,
+                    appInfo.versionName,
+                    Build.MODEL,
+                    Build.VERSION.RELEASE,
+                    Build.VERSION.SDK_INT);
+        }
```
cagryInside: @barbeau We do not show `Loc` field if there is no location

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportTypeListFragment.java
```diff
@@ -129,9 +129,12 @@ public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
                     .getStringExtra(BaseReportActivity.LOCATION_STRING);
             UIUtils.sendEmail(getActivity(), email, locationString);
 
+            String gaEventLabel = locationString != null ?
+                    getString(R.string.analytics_label_app_feedback) :
+                    getString(R.string.analytics_label_app_feedback_without_location);
```
cagryInside: @barbeau If there is no location, we send `Send Email Link without location` as a GA label.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/dialog/CustomerServiceDialog.java
```diff
@@ -123,11 +123,14 @@ protected void initView(View view, ObaAgencyWithCoverage coverage) {
                     public void onClick(View view) {
                         String locationString = getActivity().getIntent().
                                 getStringExtra(BaseReportActivity.LOCATION_STRING);
-                        String agencyName = "\nAgency: " + agency.getName();
-                        UIUtils.sendEmail(getActivity(), agency.getEmail(), locationString + agencyName);
+                        UIUtils.sendEmail(getActivity(), agency.getEmail(), locationString);
+
+                        String gaEventLabel = locationString != null ?
```
barbeau: Could we change the logic on this, where `analytics_label_customer_service_email` is always reported, and then a separate event ` analytics_label_customer_service_email_without_location` is only reported if the location is null?

barbeau: In other words, I'd prefer not to have to combine `analytics_label_customer_service_email` and `analytics_label_customer_service_email_without_location` to count how many customer service events there are.  I'd prefer to get this from `analytics_label_customer_service_email` directly, and then look at `analytics_label_customer_service_email_without_location` to know how many null location issues we've seen.  Might make sense to change the name of `analytics_label_customer_service_email_without_location` as well.

cagryInside: Fixed in this commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/2e130cac93ed9a1aad285c04e51f77fb3f4777e3.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/dialog/RegionValidateDialog.java
```diff
@@ -0,0 +1,59 @@
...
+                .setMessage(message.toString())
+                .setPositiveButton("YES", new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int which) {
+                        ((ReportActivity) getActivity()).createIssueTypeListFragment();
+                    }
+                })
+                .setNegativeButton("NO", new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int which) {
+                        closeSuperActivity();
+                        ((ReportActivity) getActivity()).createPreferencesActivity();
```
barbeau: @cagryInside This transition is currently a little abrupt for users if they answer "no".  For example, assume I am physically in Seattle.  I tap "Send feedback", and the app asks me (for a reason unknown to the user) "Are you in Tampa Bay?"  I tap "No", and suddenly I'm looking at the Preferences.  I don't know why.  I tap back to go back to submitting an issue, and go through all this again.

Let's add a dialog in the PreferenceActivity that is only shown if the user is directed to the PreferencesActivity from this screen (we can pass a variable via the Intent to trigger this).  It should have the title of "Check Region", with a single "Ok" button, and the text "Your region is currently set to (regionName), but you said you weren't located in (regionName).  Please tap on 'Your Region" and change to the correct region".

onebusaway-android/src/main/java/org/onebusaway/android/util/RegionUtils.java
```diff
@@ -505,7 +564,8 @@ public static ObaRegion getRegionFromBuildFlavor() {
         ObaRegionElement region = new ObaRegionElement(regionId,
                 BuildConfig.FIXED_REGION_NAME, true,
                 BuildConfig.FIXED_REGION_OBA_BASE_URL, BuildConfig.FIXED_REGION_SIRI_BASE_URL,
-                boundsArray, BuildConfig.FIXED_REGION_LANG, BuildConfig.FIXED_REGION_CONTACT_EMAIL,
+                boundsArray,new ObaRegionElement.Open311Server[0], BuildConfig.FIXED_REGION_LANG,
```
barbeau: For this to be set up properly from build variants, you'll need to add three new fields to `oba`, `agencyX`, and `agencyY` in `build.gradle`.

For `oba` and `agencyX`, make the values `null`:

`buildConfigField "String", "FIXED_REGION_OPEN311_JURISDICTION_ID", "null"`
`buildConfigField "String", "FIXED_REGION_OPEN311_API_KEY", "null"`
`buildConfigField "String", "FIXED_REGION_OPEN311_BASE_URL", "null"`

For `agencyY`, we'll hard code these values to the SeeClickFix test server (because the `agencyY` build variant is fixed to the Tampa OBA API):

`buildConfigField "String", "FIXED_REGION_OPEN311_JURISDICTION_ID", "null"`
`buildConfigField "String", "FIXED_REGION_OPEN311_API_KEY", "937033cad3054ec58a1a8156dcdd6ad8a416af2f"`
`buildConfigField "String", "FIXED_REGION_OPEN311_BASE_URL", "http://test.seeclickfix.com/open311/v2/"`

`ObaRegionElement.Open311Server[0]` should then be instantiated above, and you should pass in `jurisdictionId`, `apiKey`, `baseUrl` from `BuildConfig`, like:

~~~
ObaRegionElement.Open311Server[] open311Array = new ObaRegionElement.Open311Server[1];

ObaRegionElement.Open311Server open311Server;

if (FIXED_REGION_OPEN311_BASE_URL != null) {
        open311Server = new ObaRegionElement.Open311Server (
                BuildConfig.FIXED_REGION_OPEN311_JURISDICTION_ID, BuildConfig.FIXED_REGION_OPEN311_API_KEY,
                BuildConfig.FIXED_REGION_OPEN311_BASE_URL);
} else {
open311Server = new ObaRegionElement.Open311Server[0]
}
open311Array [0] = open311Server;

ObaRegionElement region = new ObaRegionElement(regionId,
                BuildConfig.FIXED_REGION_NAME, true,
                BuildConfig.FIXED_REGION_OBA_BASE_URL, BuildConfig.FIXED_REGION_SIRI_BASE_URL,
                boundsArray,
open311Array, 
BuildConfig.FIXED_REGION_LANG,
                BuildConfig.FIXED_REGION_CONTACT_EMAIL,
                BuildConfig.FIXED_REGION_SUPPORTS_OBA_DISCOVERY_APIS,
                BuildConfig.FIXED_REGION_SUPPORTS_OBA_REALTIME_APIS,
                BuildConfig.FIXED_REGION_SUPPORTS_SIRI_REALTIME_APIS,
                BuildConfig.FIXED_REGION_TWITTER_URL, false,
                BuildConfig.FIXED_REGION_STOP_INFO_URL);
~~~

barbeau:
@cagryInside When looking at the Arrival Time Problem, it appears that the drop downs for problem type and sub-category have a different padding between the text and the bottom line?  See the red lines in the below:

image

...and the two placed side-by-side:

image

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/dialog/RegionValidateDialog.java
cagryInside: @barbeau What do you think?

device-2016-04-13-101344

barbeau: Hmmm...let's tweak it a bit to be more conversational:

>**Please check region**

>Your region is currently set to Atlanta.  If we got this wrong, please tap on 'Your Region" to change it!

cagryInside:
Commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/41a6cc285e77865125454a132bbe67ced556d1c6 fixes the following issues mentioned in the comments:

* https://github.com/OneBusAway/onebusaway-android/pull/392/files/2e130cac93ed9a1aad285c04e51f77fb3f4777e3#r59560427
* https://github.com/OneBusAway/onebusaway-android/pull/392#issuecomment-209446292
* https://github.com/OneBusAway/onebusaway-android/pull/392/files/2e130cac93ed9a1aad285c04e51f77fb3f4777e3#r59548144

barbeau:
Thanks @cagryInside, this all looks really good.  Let's go ahead and squash the commits to prepare for merge, and if anything else pops up we can open individual issues for it.  Great work!

cagryInside:
@barbeau Just squashed all commits. Many thanks for your helps.

barbeau:
@cagryInside Ah, sorry, I missed something with the Gradle build flavors.  In `build.gradle` under `agencyY`, you'll need to escape the new Strings properly for the new Open311 fields.  See the other strings above it for examples.  The agencyY build variant currently fails to build without the escaped strings - select it via Build Variants window in Android Studio (see [this](https://github.com/OneBusAway/onebusaway-android#when-running-the-project-i-get-a-nullpointerexception-in-basemapfragment-related-to-mmap) for screenshot), or run `gradlew install agencyYGoogleDebug` from command line to see the errors.

cagryInside:
> @cagryInside Ah, sorry, I missed something with the Gradle build flavors. In build.gradle under agencyY, you'll need to escape the new Strings properly for the new Open311 fields. See the other strings above it for examples. The agencyY build variant currently fails to build without the escaped strings - select it via Build Variants window in Android Studio (see this for screenshot), or run gradlew install agencyYGoogleDebug from command line to see the errors.

My bad, I didn't test that. Now, It should work.

barbeau:
Sorry, more stuff related to the build flavor - before launching the "Are you in X?" dialog, we should check and make sure this isn't the agency Y variant, and if it's agency Y variant, don't show it (because Agency Y is locked to Tampa API, and doesn't support multiple regions).  Otherwise we get this:

image

I believe to do this you'll need to create a new constant `BuildFlavorUtils.AGENCYY_FLAVOR_BRAND` with the value `agencyY`, and then do an IF statement like this:

~~~
if (BuildConfig.FLAVOR_brand
                        .equalsIgnoreCase(BuildFlavorUtils.AGENCYY_FLAVOR_BRAND)) {
    // Don't show dialog for Agency Y build variant, because it's locked to a single region
    return;
}
~~~

On a related note, I believe in `RegionUtils.getRegionFromBuildFlavor()`, we should be passing a `null` value for `open311Server` when contructing the `ObaRegionElement`.  Elsewhere (including `RegionUtils.saveToProvider()`) we check if `region.getOpen311Servers()` is equal to null, and if it's not, we assume there is a Open311 server and try to do something with it.

Also, small tweak - in `RegionUtils.saveToProvider()`, there is a typo - `open311Serverses` should be `open311Servers`.

barbeau:
Sorry, to clarify - in `RegionUtils.getRegionFromBuildFlavor()`, we should be passing a null value for open311Server when contructing the ObaRegionElement **IFF** it's a null value in the `build.gradle`.  We won't actually have that condition in our current setup, because all other variants get their Open311 servers from the Regions API, and we have a non-null value in `build.gradle` for Agency Y.

cagryInside:
> Sorry, to clarify - in RegionUtils.getRegionFromBuildFlavor(), we should be passing a null value for open311Server when contructing the ObaRegionElement IFF it's a null value in the build.gradle. We won't actually have that condition in our current setup, because all other variants get their Open311 servers from the Regions API, and we have a non-null value in build.gradle for Agency

@barbeau I just pushed another commit https://github.com/OneBusAway/onebusaway-android/pull/392/commits/a17f4d8764d3a71ea7966506b13ada2bf7ff9411. If you think this is good, then I will `squash`.

barbeau:
Thanks @cagryInside, this all looks good!  Please squash!

cagryInside:
@barbeau squashed the commits.

barbeau:
Thanks @cagryInside!  Awesome to get this merged!

barbeau:
@cagryInside When a user reports a stop or trip (arrival time) problem in an area with Open311 coverage, we allow them to attach a picture, right?

I'm not seeing this in my current debug build, but I think it's just an effect of using the test Open311 endpoint instead of production.

cagryInside:
> @cagryInside When a user reports a stop or trip (arrival time) problem in an area with Open311 coverage, we allow them to attach a picture, right?
> 
> I'm not seeing this in my current debug build, but I think it's just an effect of using the test Open311 endpoint instead of production.

Yes, if the area is covered by Open311, we always allow them to attach a picture. [This](http://seeclickfix.com/issues/2424715-psta-bus-stops) was the latest submission for stop problem, and I successfully attached an image.

barbeau:
Awesome, thanks!

PR #394, opened 2015-12-02
Replace Gitter badge with Slack badge in Readme

barbeau:


bbodenmiller:
Here's details on how this was setup: http://rauchg.com/slackin/

PR #395, opened 2015-12-02
Add code of conduct

barbeau:


bbodenmiller:
Should we add this to other OBA repos?

barbeau:
IMHO yes - but, I'm actually running this by the OBA Board right now to get consensus and official approval for it to be used project-wide.  So hopefully we'll approve/announce this shortly.

Issue #420, opened 2016-02-16
Agency X and Y tutorial colors are wrong


barbeau:
For the Agency X and Y demo apps, the tutorial colors are currently shown as green instead of the theme colors for each of these apps - red and blue respectively.  The color, and reference to color in the tutorial text (e.g., "green dots") should be changed for these apps to be the theme colors.

This affects the Agency X and Agency Y demo apps for build flavors, and not the OBA Google or Amazon builds, so it's not an emergency.  It can be fixed and released whenever the next update happens to roll out.

barbeau:
Whoops, forgot to change tutorial text accordingly.  Re-opening.

themonki:
@barbeau I think it is also necessary in the key tutorial_arrival_header_arrival_info_text:

This long! Colored background tells you
        if the bus is early (red) or late (blue). **On time = green background.** If we don\'t know
        where the bus is, the scheduled time is shown (gray background).

On time use the default background color.

For example on Agency Y maybe should say:

This long! Colored background tells you
        if the bus is early (red) or late (**dark blue**).**On time = blue background.** If we don\'t know
        where the bus is, the scheduled time is shown (gray background).

barbeau:
@themonki Ah - good catch.  You're right - re-opening to fix that.

barbeau:
@themonki Ok, this should be fixed in https://github.com/OneBusAway/onebusaway-android/commit/b9872aa9292d4969d427a6580204a9c4348e19f8.  The choice of the arrival time colors isn't the best in the demo apps, as both the blue and red of the theme colors overlap (to some degree) the arrival time early and late colors.  But, I think it at least gets the point across of how the colors can be customized in white-labeled versions of OBA.

barbeau:
@themonki Actually, there should be a `values-es` folder under `src/agencyX/res` and `src/agencyY/res` with a translated `strings.xml` for those build flavors that match `src/agencyX/res/values/strings.xml` and `src/agencyY/res/values/strings.xml`, respectively.  Feel free to tackle if you'd like for Spanish ;).

themonki:
@barbeau Yes, is cool the colors can be customized, but i think for the arrival time colors should be a standard for OBA. Because the user must already be accustomed to seeing these colors.

I send a PR soon for `values-es` for AgencyY and AgencyX.

[...]

PR #424, opened 2016-02-18
Translations for AgencyX and AgencyY brands

themonki:
This translation relate to #420 and example how to fun with colors. 
CC @barbeau


barbeau:
Awesome, thanks @themonki!

PR #382, opened 2015-11-23
report-problem-with-* requests follow API, added tests

barbeau:
This is a re-opening of the previous PR https://github.com/OneBusAway/onebusaway-android/pull/276, this time opened against the `master` branch instead of `develop`.  It is intended to fix https://github.com/OneBusAway/onebusaway-android/issues/264.

Please see https://github.com/OneBusAway/onebusaway-android/pull/276 for full history, including comments.

PR #427, opened 2016-02-25
Fix #264 - Fix stop/trip problem reporting API requests

### LINKED ISSUE #264: Improve stop/trip problem reporting API requests

cagryInside:
This PR fixes issue #264.  

Get requests are now used to report stop/trip problems instead of post requests.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemWithStopRequestTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemWithStopRequestTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (C) 2016 University of South Florida (cagricetin@mail.usf.edu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


barbeau:
Thanks @cagryInside!  My only comment is that we may want to use the normal GET request construction in OBA that's used for the other HTTP requests instead of turning a POST into a GET, for two reasons:

1. We may have a genuine need for a POST in the future - currently it seems like that's what is offered from the methods, but internally we're turning it into a GET.
2. We have [a number of unit tests](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/UrlFormatTest.java) to check to make sure all types of OBA base URLs are formatted correctly, when using the internal GET construction - we've hit issues in the past with base URLs that have port numbers, spaces in agency names, partial paths, etc.  I'd like to confirm via those tests that the problem reporting URLs are formatted correctly.

We can discuss more in meeting today.

barbeau:
Actually, looks like there are some unit tests for this you can build on from https://github.com/OneBusAway/onebusaway-android/pull/382.

cagryInside:
@barbeau I updated the PR with new test cases.

barbeau:
@cagryInside Thanks, looks good!

Issue #434, opened 2016-03-07
Update Finnish translation


barbeau:
@arirusakko Would you be able to do another quick translation?

I believe all the new strings are contained in these commits, although you probably want to double-check:

* https://github.com/OneBusAway/onebusaway-android/commit/0b0991ee53a040ab032f18528ad3fca0a0026580
* https://github.com/OneBusAway/onebusaway-android/commit/5bf07287a6d378566648c1123ad1f873340edc47
* https://github.com/OneBusAway/onebusaway-android/commit/c25879d45fc254ad2bd2930eacf1c15ec378828e

arirusakko:
I will try to look at some point, maybe next week. Currently having too much ongoing at work.

arirusakko:
made the updates: https://github.com/OneBusAway/onebusaway-android/pull/446

PR #446, opened 2016-03-22
Update Finnish translation #434

arirusakko:
Only dummy translation update. 

Added some line breaks too for easing the comparing of the english and finnish translation files.


barbeau:
Awesome, thanks @arirusakko!

Issue #436, opened 2016-03-08
Show only departures at first/last stop for loop routes


barbeau:
Related to https://github.com/OneBusAway/onebusaway-android/issues/287, but this issue is specifically for loop routes.

Same issue on OBA iOS at https://github.com/OneBusAway/onebusaway-iphone/issues/562.

For the first/last stop on loop routes, we currently show both the arrival and the departure, which is confusing.  For example:

![image](https://cloud.githubusercontent.com/assets/928045/13612857/9787cbc0-e537-11e5-9db4-d59a6077b531.png)

The 32 is arriving in 3 minutes, and departing in 8 minutes.  So, same vehicle, it just has a 5 minute layover.

Via https://github.com/OneBusAway/onebusaway-application-modules/issues/148 and PR https://github.com/OneBusAway/onebusaway-application-modules/pull/149, we now have access to a new API parameter `totalStopsInTrip` on some OBA installations that should help with this.

>* totalStopsInTrip - the total number of stops visited on the trip for this arrival.  If the same stop is visited more than once in this trip, each visitation is counted towards the total.
* stopSequence - the index of the stop into the sequence of stops that make up the trip for this arrival.  This value is 0-indexed, and is generated internally by OneBusAway (it is **not** the GTFS `stop_sequence`).  The first stop in the trip will always have `stopSequence = 0`, while the last stop in the trip will always have `stopSequence = totalStopsInTrip - 1`.

Here is a request for the above stop - you can see the `totalStopsInTrip` field in the response:

http://app1.staging.obahart.org/api/where/arrivals-and-departures-for-stop/PSTA_1767.json?minutesAfter=65&key=TEST

A possible rule to help with this display issue may be to only show departures (and not arrivals) for trips that are in the same block with sequential `blockTripSequence` with the same headsign.  In this case, the top 32 arrival in 3 min would be hidden, while the departure in 8 min would be shown.

bbodenmiller:
Rather than hide the arrival time I think it'd be more valuable to find a way to label them appropriately. Sometimes you want to be able to see the arrival time, such as when you are trying to make a connection and want to see if you can actually make it.

barbeau:
@bbodenmiller Better labeling could definitely help when there is a layover - we'd certainly need this in the unfiltered view anyway (if we choose to automatically filter).

[...]

PR #449, opened 2016-03-25
Fix #287 - Better distinguish between arrivals/departures

### LINKED ISSUE #287: Distinguish between arrivals and departures clearly at terminal stops for routes
The stop info screen for a regular stop shows only departures from that stop, but that screen of a terminal stop shows departures as well as arrivals. This is inconsistent with the behavior of all other stop info screens.

Let me know how you guys think this can be addressed.. My suggestion is to show the schedules/predicted arrival time underneath/beside the scheduled departure time if that route starts at that stop. For all other routes, these times are going to be the same, so we can safely ignore them.

Thoughts?

barbeau:
~~**Please don't merge**~~

This is a simple addition to the UI, just adding "Arriving at" or "Departing at" in front of the time (or the past tense as appropriate).

image

TODO:

- [x] The logic surrounding marking an arrival/departure needs to be examined in case of loop routes
- [x] The status label wrongly says "departed" when it should actually be "arrived"
- [x] More testing to make sure we didn't miss anything else
- [x] Better method to control showing "arrival" or "departure" word in the status label Style A vs. Style B? (we only want to show "Scheduled" in Style A now, because right below it we're showing "arriving" or "departing" next to the time)
- [x] Unit tests for the wording logic would be nice

### DIFF
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
@@ -652,7 +652,7 @@ public class VehicleOverlay implements AmazonMap.OnInfoWindowClickListener {
                 statusView.setPadding(pSides, pTopBottom, pSides, pTopBottom);
             } else {
                 // Scheduled info
-                statusView.setText(r.getString(R.string.stop_info_scheduled_arrival));
+                statusView.setText(r.getString(R.string.stop_info_scheduled));
                 statusColor = R.color.stop_info_scheduled_time;
                 d.setColor(r.getColor(statusColor));
                 lastUpdatedView.setText(r.getString(R.string.vehicle_last_updated_scheduled));
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ArrivalInfoRequestTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ArrivalInfoRequestTest.java
@@ -161,6 +161,45 @@ public class ArrivalInfoRequestTest extends ObaTestCase {
         assertTrue(nearbyStops.size() > 0);
     }
 
+    public void testTotalStopsInTrip() throws Exception {
+        // Test by setting API directly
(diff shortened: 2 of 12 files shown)


barbeau:
This does help a bit with the loop route issue (https://github.com/OneBusAway/onebusaway-android/issues/436) as well.  You can see that the 2nd item in the list is the 96 arriving in 15 min (with a 6 min delay), and the 3rd item in list is the 96 departing in 15 min (with a 1 min delay):

image

An aside - not sure about the exact delay values here - somehow the bus is expected to make up a 5 min delay in 1 min (at 6:01 PM) - but that's beyond the scope of this issue/PR.  Something else to consider in #436.

barbeau:
Ah - it's a Monday - so the delay values make sense here.  Bus was scheduled to arrive at 5:55pm and have a 5 min layover, and depart at 6pm.  It's running 6 min late, which means it will eat up the layover time and still depart 1 min late.  So all's well.

barbeau:
This is ready for merge now - final commit description:

* For Style A arrival cards, show "Arriving/Departing at <time>" in the time field, or past tense "Arrived/Departed at <time>" that occurred in the past.  Also, remove "Departed" from arrival info status field, because this wasn't accurate in the case of arriavls, and we're now showing this info in the time field.
* For Style B arrival cards, show past tense "Arrived/Departed on time / X min late / X min early" for past times as part of the status label.
 * Change vehicle marker on map status from "Scheduled arrival" to "Scheduled"
* Add unit tests for arrival info status and time labels
* Add support for totalStopsInTrip field in arrivals response (including tests), even though we didn't end up using it here

barbeau:
A view of some of the past arrivals/departures in the Style B cards, as well as a scheduled arrival:

image

barbeau:
Here's the current logic for indicating an arrival vs. departure:

~~~
// If this is the first stop in the sequence, show the departure time.
if (info.getStopSequence() != 0) {
    scheduled = info.getScheduledArrivalTime();
    predicted = info.getPredictedArrivalTime();
    mIsArrival = true;
} else {
    // Show departure time
   scheduled = info.getScheduledDepartureTime();
   predicted = info.getPredictedDepartureTime();
   mIsArrival = false;
}
~~~

PR #455, opened 2016-04-01
Update Finnish translation

arirusakko:
I've added the stuff caused by https://github.com/OneBusAway/onebusaway-android/pull/449

Also I went through the existing stuff and made some changes that felt better (and usually shorter)


barbeau:
@arirusakko Great, thanks!

PR #462, opened 2016-04-15
Fix  #458 - Add legend for arrival time colors

### LINKED ISSUE #458: Add legend for arrival time colors
This has been requested numerous times before, and I'd like to add it along side the work we're doing for issue reporting (https://github.com/OneBusAway/onebusaway-android/pull/392).

We should add a Legend option under "Help", that looks something like this (but obviously much prettier):

image

Top circle is the animating real-time indicator we use next to the real-time arrival estimate.

@cagryInside Could you please work on this?  We can get https://github.com/OneBusAway/onebusaway-android/pull/392 merged and released, and then add this after the fact.  So, lower priority than https://github.com/OneBusAway/onebusaway-android/pull/392, but I'd like to get it in the public v2.0 release.

cagryInside:
This PR adds a Legend section to `Help` section.

device-2016-04-15-113301
device-2016-04-15-113406

A view in a smaller screen:

device-2016-04-15-113821
device-2016-04-15-113846


cagryInside:
@barbeau Just pushed a commit. Please let me know if you have further comments.

onebusaway-android/src/main/res/layout/arrivals_list_header_row_template.xml
```diff
@@ -2,145 +2,107 @@
...
-                        android:layout_marginTop="16dp"
-                        android:layout_marginLeft="1dp"/>
-        </RelativeLayout>
+    <include
+            android:id="@+id/eta_and_min"
+            layout="@layout/eta_header_view"
+            android:layout_marginTop="3dp"
+            android:layout_marginBottom="3dp"
+            android:layout_height="wrap_content"
+            android:layout_width="wrap_content"/>
```
barbeau: @cagryInside Could you put the `RelativeLayout` attributes here that are in the included view layout?

It should look something like:

~~~
    <include
            android:id="@+id/eta_and_min"
            layout="@layout/eta_header_view"
            android:layout_height="wrap_content"
            android:layout_width="wrap_content"
            android:layout_column="2"
            android:layout_marginLeft="@dimen/arrival_header_text_pad_left"
            android:layout_marginTop="3dp"
            android:layout_marginBottom="3dp"
            android:paddingLeft="4dp"
            android:paddingRight="4dp"
            android:gravity="right"
            android:orientation="horizontal"/>
~~~

(This is what I tested with - it seems `android:background="@drawable/round_corners_style_b_header_status"` is inherited from the included layout without a problem)

We need to list these attributes in this `<include>`, otherwise not all of them get carried over to the final layout (this isn't documented very well in Android, and IIRC inheritance behavior can be different for each type of attribute).  There are very small but visible differences in the padding and compression behavior of the table columns.  It's really hard to replicate this issue and compare against `master` branch because you need to catch the exact same combination of trip headsign and single/double digit ETAs to get the exact same layout result.  I wasn't convinced myself that the behavior was different until I caught the right headsign/ETAs :).

Here's an image created by layering a transparent screenshot above a solid screenshot with the same route/headsign, and very similar ETAs:
image

The transparent image on top is the one without the additional layout attributes included (i.e., this PR currently).  You can see that the colored box surrounding the ETA is slightly wider on the left than the ETA box in the solid image underneath, which includes all the attributes in the `<include>` element.

So, by including all attributes in the `<include>` element, it results in a slightly more compact ETA box.  I believe this behavior should match the `master` branch, although additional testing is welcome :).

barbeau:
Thanks @cagryInside!  One comment in-line that we need to fix.

onebusaway-android/src/main/res/layout/arrivals_list_header_row_template.xml
barbeau: I think you might need to add these to the legend layout as well.  It looks like the padding on the left is less for the scheduled time than it is for real-time arrivals.

cagryInside:
@barbeau just pushed a commit that fixes in-line comments:

This is the comparison between master and this: 
device-2016-04-23-232714

This fixes the small padding problem in legend:

device-2016-04-23-235846

barbeau:
Awesome, thanks!

PR #463, opened 2016-04-16
Fix #460 - Trip Status activity started from Vehicle marker tap now shows select…

### LINKED ISSUE #460: Pass in stopId to TripDetailsActivity when tapping on vehicle marker balloon
**Summary:** 

When tapping on the vehicle marker balloon on the map:

image

...you are taken to the TripDetailsListFragment.  

image

However, if you had a stop selected, we don't show the little landmark icon to indicate which stop you were viewing.  

This is contrary to if you tap on an arrival time, and select "Show trip status" - then we show the landmark marker.

image

We should show this same landmark marker when opening TripDetailsListFragment from tapping on the vehicle marker balloon - we just need to pass the currently selected stopId via an Intent to TripDetailsActivity.

azizmb9494:
As per issue #460, when a vehicle marker is tapped while a stop is selected, the stop marker shows up in Trip Status activity.

### DIFF
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -27,6 +27,25 @@
  */
 package org.onebusaway.android.map.googlemapsv2;
 
+import android.app.Activity;
+import android.app.Dialog;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.graphics.drawable.Drawable;
+import android.location.Location;
+import android.os.Bundle;
+import android.os.Handler;
+import android.provider.Settings;
+import android.support.v4.graphics.drawable.DrawableCompat;
+import android.support.v7.app.AlertDialog;
+import android.text.TextUtils;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Toast;
+
 import com.amazon.geo.mapsv2.CameraUpdateFactory;
 import com.amazon.geo.mapsv2.LocationSource;
 import com.amazon.geo.mapsv2.OnMapReadyCallback;
@@ -57,25 +76,6 @@ import org.onebusaway.android.util.LocationHelper;
 import org.onebusaway.android.util.LocationUtils;
 import org.onebusaway.android.util.UIUtils;
 
-import android.app.Activity;
-import android.app.Dialog;
-import android.content.Context;
-import android.content.DialogInterface;
-import android.content.Intent;
-import android.graphics.drawable.Drawable;
-import android.location.Location;
-import android.os.Bundle;
-import android.os.Handler;
-import android.provider.Settings;
-import android.support.v4.graphics.drawable.DrawableCompat;
-import android.support.v7.app.AlertDialog;
-import android.text.TextUtils;
-import android.util.Log;
-import android.view.LayoutInflater;
-import android.view.View;
-import android.view.ViewGroup;
-import android.widget.Toast;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -102,7 +102,8 @@ public class BaseMapFragment extends SupportMapFragment
         MapModeController.ObaMapView,
         LocationSource, LocationHelper.Listener,
         com.amazon.geo.mapsv2.AmazonMap.OnCameraChangeListener,
-        StopOverlay.OnFocusChangedListener, OnMapReadyCallback {
+        StopOverlay.OnFocusChangedListener, OnMapReadyCallback,
+        VehicleOverlay.Controller {
 
     public static final String TAG = "BaseMapFragment";
 
@@ -183,7 +184,7 @@ public class BaseMapFragment extends SupportMapFragment
 
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container,
-            Bundle savedInstanceState) {
+                             Bundle savedInstanceState) {
         View v = super.onCreateView(inflater, container, savedInstanceState);
 
         mMap = getMap();
@@ -348,6 +349,7 @@ public class BaseMapFragment extends SupportMapFragment
         Activity a = getActivity();
         if (mVehicleOverlay == null && a != null) {
             mVehicleOverlay = new VehicleOverlay(a, mMap);
+            mVehicleOverlay.setController(this);
         }
     }
 
@@ -529,7 +531,7 @@ public class BaseMapFragment extends SupportMapFragment
     final Handler mStopChangedHandler = new Handler();
 
     public void onFocusChanged(final ObaStop stop, final HashMap<String, ObaRoute> routes,
-            final Location location) {
+                               final Location location) {
         // Run in a separate thread, to avoid blocking UI for long running events
         mStopChangedHandler.post(new Runnable() {
             public void run() {
@@ -537,6 +539,7 @@ public class BaseMapFragment extends SupportMapFragment
                     mFocusStopId = stop.getId();
                     //Log.d(TAG, "Focused changed to " + stop.getName());
                 } else {
+                    mFocusStopId = null;
                     //Log.d(TAG, "Removed focus");
                 }
 
@@ -653,7 +656,7 @@ public class BaseMapFragment extends SupportMapFragment
      */
     @Override
     public void setMapCenter(Location location, boolean animateToLocation,
-            boolean overlayExpanded) {
+                             boolean overlayExpanded) {
         if (mMap != null) {
             CameraPosition cp = mMap.getCameraPosition();
 
@@ -789,7 +792,7 @@ public class BaseMapFragment extends SupportMapFragment
      */
     @Override
     public void zoomIncludeClosestVehicle(HashSet<String> routeIds,
-            ObaTripsForRouteResponse response) {
+                                          ObaTripsForRouteResponse response) {
         if (mMap == null) {
             return;
         }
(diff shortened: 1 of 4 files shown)


onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -53,6 +53,7 @@
 import org.onebusaway.android.map.RouteMapController;
 import org.onebusaway.android.map.StopMapController;
 import org.onebusaway.android.region.ObaRegionsTask;
+import org.onebusaway.android.ui.HomeActivity;
```
barbeau: @azizmb9494 Could you please make sure you have "Optimize imports" checked in Android Studio when you go to commit the changes?  It should automatically remove any unused imports like this when you commit.

barbeau:
@azizmb9494 Thanks for tackling this!  I think instead of relying on pulling a string from an intent from the activity, I'd prefer to set up a controller interface where the `VehicleOverlay` could pull the `stopId` from the controller.  This would allow for a more formal contract that's also decoupled from any activities.  IIRC, we have a similar setup with `ArrivalListFragment` and `ArrivalListHeader`, with the header pulling `stopIds` and more from the `ArrivalListHeader`.

To do this, I'd suggest setting up a new interface `Controller` in `VehicleOverlay`, with a single `getFocusedStopId()` method, and add a new `setController(Controller controller)` method in `VehicleOverlay`.  `BaseMapFragment` would then implement `VehicleOverlay.Controller` and return the focused `stopId` if a stop is in focus, or `null` if no stop is in focus.  You'd also need to call `VehicleOverlay.setController(this)` when instantiating the vehicle overlay.

If you want to give this a shot it would be awesome.  Easiest way to tackle would be to roll back existing changes locally with `git reset HEAD^`, which will re-stage your existing changes in the same branch (effectively "uncommitting" those changes).  Then you can make changes, and commit when you're ready, and then `git commit vehicle-marker-stopid:vehicle-marker-stopid -f`, which will force push your new commit into this PR, replacing the existing commit in this PR and in your current `vehicle-marker-stopid` branch.

Thanks again for taking time to tackle this issue!

azizmb9494:
@barbeau No problem! I will look into doing that next weekend, or if time allows, this week!

barbeau:
Great, sounds good!  Also, you probably already figured this out, but make sure you edit the `BaseMapFragment` under the `/google` folder path (and not the `amazon` folder path).  When you build the project, it will automatically generate the same class under the `amazon` folder.

azizmb9494:
Hey @barbeau, thanks for the advice, I've implemented as you recommended and pushed it!

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -287,7 +300,13 @@ public void onInfoWindowClick(Marker marker) {
 
         // Show trip details screen for the vehicle associated with this marker
         ObaTripStatus status = mMarkerData.getStatusFromMarker(marker);
-        TripDetailsActivity.start(mActivity, status.getActiveTripId());
+
+        String focusedStopId = mController.getFocusedStopId();
```
barbeau: You should check if `mController` is null before calling `mController.getFocusedStopId()`, to prevent potential NPEs.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Dec 10 11:19:41 EST 2014
+#Mon Apr 25 00:16:55 EDT 2016
```
barbeau: Could you remove the modification to this file from this commit?

I believe these are from Android Studio 2.0 - I'd like to bump them in a separate commit.

onebusaway-android/build.gradle
```diff
@@ -20,7 +20,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.0'
+        classpath 'com.android.tools.build:gradle:2.0.0'
```
barbeau: Could you remove the modification to this file from this commit?

I believe these are from Android Studio 2.0 - I'd like to bump them in a separate commit.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -518,7 +520,7 @@ public void setOnFocusChangeListener(OnFocusChangedListener onFocusChangedListen
     final Handler mStopChangedHandler = new Handler();
 
     public void onFocusChanged(final ObaStop stop, final HashMap<String, ObaRoute> routes,
-            final Location location) {
+                               final Location location) {
```
barbeau: From a quick look at `onFocusChanged()`, it looks like `mFocusStopId` isn't reset to `null` when focus is lost:

~~~
                if (stop != null) {
                    mFocusStopId = stop.getId();
                    //Log.d(TAG, "Focused changed to " + stop.getName());
                } else {
                    //Log.d(TAG, "Removed focus");
                }
~~~

I believe the ELSE block should have `mFocusStopId = null`, although you'd need to be sure there aren't any other unintended consequences of this.  `onSaveInstanceState()` is the only place where this value is used in `BaseMapFragment`, so it looks like we currently retain the previously selected stopId here, even if the stop isn't currently in focus.  Not sure off the top of my head if this was intended behavior.

barbeau:
@azizmb9494 Thanks, this is looking good!  A few comments in-line that need to be addressed, but conceptually it looks good.

azizmb9494:
@barbeau: Suggested changes made!

barbeau:
Thanks @azizmb9494!  It looks good, although I want to do a little more testing myself to make sure that the change to reset the stopId to null in `BaseMapFragment` doesn't have any adverse effects.

azizmb9494:
Awesome @barbeau! Let me know if there are any issues and I'll adjust accordingly!

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -298,7 +311,13 @@ public void onInfoWindowClick(Marker marker) {
 
         // Show trip details screen for the vehicle associated with this marker
         ObaTripStatus status = mMarkerData.getStatusFromMarker(marker);
-        TripDetailsActivity.start(mActivity, status.getActiveTripId());
+
+        String focusedStopId = mController.getFocusedStopId();
```
barbeau: @azizmb9494 It looks like the Amazon implementation is slightly different here than the Google implementation.  Sometimes this can happen if the Gradle copy script doesn't execute properly when you're building the Google variant.  Could you try re-building this project again using any of the Google build variants (oba, agencyx, agencyy) to re-generate the Amazon code, and then push that new version here?

azizmb9494:
@barbeau: Rebased, rebuilt & pushed. Please let me know if that did the trick.

barbeau:
@azizmb9494 Thanks!  Looks like that did the trick - the amazon implementation now matches the google one.  Hopefully I'll have a chance to finish testing this soon and I can get this merged.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -529,14 +531,15 @@ public void setOnFocusChangeListener(OnFocusChangedListener onFocusChangedListen
...
-            final Location location) {
+                               final Location location) {
         // Run in a separate thread, to avoid blocking UI for long running events
         mStopChangedHandler.post(new Runnable() {
             public void run() {
                 if (stop != null) {
                     mFocusStopId = stop.getId();
                     //Log.d(TAG, "Focused changed to " + stop.getName());
                 } else {
+                    mFocusStopId = null;
```
barbeau: @cagryInside Could you please take a look at this PR as well?  I'm about 99.9% sure that this line `mFocusStopId = null` should be here when the stop loses focus - I believe I somehow missed this in the initial implementation of `BaseMapFragment`.  i'd like you to take a look and confirm that adding the assignment to `null` isn't going to cause any problems for any of the activities using this fragment, including the issue reporting activities, as well as the `BaseMapFragment` internally.  Specifically, the `mFocusStopId` is written to the `Bundle` in `onSaveInstanceState()` - so previous to this patch, the previous String stopId would always have been saved to the bundle (e.g., on orientation change).  Now, if the stop lost focus, that value will be `null`.  Inside `BaseMapFragment`, it doesn't appear as though `mFocusStopId` has been used - this patch is the first code to use it, AFAIK (which is probably why I missed the initial null assignment - it wasn't actually being used).  I looked at HomeActivity and I don't believe there should be any adverse affects there.  Could you give this a second look, and take a look at your use of `BaseMapFragment` with this patch in issue reporting as well?

barbeau: @cagryInside Friendly ping on this - could you please take a look in the next few days so we can get this merged?  See above for my notes.

cagryInside: > @cagryInside Friendly ping on this - could you please take a look in the next few days so we can get this merged? See above for my notes.

@barbeau I totally forgot this one! I look the `InfrastructureIssueActivity` which uses the `BaseMapFragment`. There is no bad affect here too.

barbeau:
Sorry @azizmb9494, looks like we introduced a conflict and this PR won't merge - could you please rebase against master?  We should then be able to get this merged.  Sorry for the delay and thanks for your patience!

azizmb9494:
@barbeau: No worries, happy to help! Done!

barbeau:
ICLA manually confirmed :+1:

barbeau:
Thanks @azizmb9494, looks good!

PR #464, opened 2016-04-18
Use material design style in Supported Agencies View

cagryInside:
A material design list view was implemented for `supported agencies view`.
device-2016-04-18-115227
device-2016-04-18-115058

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/MaterialListItem.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/MaterialListItem.java
@@ -1,5 +1,5 @@
 /*
-* Copyright (C) 2014 University of South Florida (sjbarbeau@gmail.com)
+* Copyright (C) 2014-2016 University of South Florida (sjbarbeau@gmail.com)
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/MaterialListItem.java
```diff
@@ -13,24 +13,31 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
-package org.onebusaway.android.report.ui.model;
+package org.onebusaway.android.ui;
 
 /**
  * Model for report type
```
barbeau: Could you please update the comment to reflect that this is a more generic list item with material design now?

barbeau:
Thanks @cagryInside, this looks good!  One small comment inline.

onebusaway-android/src/main/java/org/onebusaway/android/ui/MaterialListItem.java
cagryInside: Fixed via https://github.com/OneBusAway/onebusaway-android/pull/464/commits/e54128318c2a48dfc8cab14fb78d071c4ef50230

barbeau:
Thanks @cagryInside!

PR #466, opened 2016-04-25
Prevent showing only one issue category for Problem Reporting

cagryInside:
If open311 endpoint returns only one issue category, we do not show it in the screen:
Before this improvement:
device-2016-04-25-135307

After this improvement:
device-2016-04-25-135145

cc'd @barbeau


barbeau:
Thanks @cagryInside!

PR #470, opened 2016-04-25
Remove open311 servers on region update

cagryInside:
Delete all existing open311 endpoints from DB on region update.

cc'd @barbeau


barbeau:
Thanks @cagryInside!

PR #471, opened 2016-04-26
WIP - Fix #437 - Add stop code to header

### LINKED ISSUE #437: Add more stop details to the header when hiding arrivals
In https://github.com/OneBusAway/onebusaway-android/issues/423 we now hide the arrivals in the header by default when coming into the ArrivalsListActivity (i.e., from the starred stops fragment or via a Home screen shorcut), which looks like this:

image

v1.x included the direction for the stop in the header as well:

image

When the arrivals are hidden, we could definitely show the direction and stop #.  Stop # would be a great help when debugging issues, as I'm getting screenshots with arrival times, but without the `stop_code` it takes a while to actually find which stop the screenshot refers to.

Here's a possible mockup:

image

barbeau:
image


barbeau:
I may end up exposing this info in a "UI debug mode", but for now I'm going to close this WIP without merging.  I'm not happy with the current presentation enough to make it a default for all users.

PR #472, opened 2016-04-27
Issue Reporting - Move Geo Coding to a separate thread and use sentence case in questions

cagryInside:
To convert the selected location to address string, a separate thread was used.


barbeau:
Thanks @cagryInside!

PR #476, opened 2016-04-29
WIP - Fix Travis

barbeau:
See if we can get Travis working again after bumping Android Gradle plugin and wrapper to work with Android Studio 2.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,14 @@ language: android
 jdk: oraclejdk7
 # Turn off caching to avoid any caching problems
 cache: false
-# Use the Travis Container-Based Infrastructure (see #203)
-sudo: false
+# Don't use the Travis Container-Based Infrastructure - See #476
+sudo: true
 env:
   global:
     - ANDROID_API_LEVEL=22
     - ANDROID_BUILD_TOOLS_VERSION=23.0.2
     - ANDROID_ABI=armeabi-v7a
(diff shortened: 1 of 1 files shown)


barbeau:
See also https://github.com/travis-ci/travis-ci/issues/5582

barbeau:
And https://github.com/mockito/mockito/pull/369

barbeau:
From log (for later note):

>To run dex in process, the Gradle daemon needs a larger heap.
It currently has approximately 1664 MB.
For faster builds, increase the maximum heap size for the Gradle daemon to more than 2048 MB.
To do this set org.gradle.jvmargs=-Xmx2048M in the project gradle.properties.
For more information see https://docs.gradle.org/current/userguide/build_environment.html

barbeau:
From https://github.com/travis-ci/travis-ci/issues/5582#issuecomment-181113633:

>This is probably a variation of https://github.com/travis-ci/travis-ci/issues/4192 and other previously reported issues where a container build gets killed because it (or a build that shares the VM with that build) consumes too much memory and gets terminated by the host.

The fix is to turn off container-based execution - in `.travis.yml` we should set:

`sudo: true`

barbeau:
Main downside of setting `sudo: true` is that builds could take longer, so I've bumped the timeout to 20 min just in case.  We really don't have much of a choice, as builds seem to always fail now with `sudo: false`.

PR #477, opened 2016-05-02
Issue Reporting improvements

cagryInside:
* Open stop problem category from Send Feedback screen
* Reset services spinner on back button press
* Validate empty fields before submitting an issue (Fixed in Open311client project https://github.com/CUTR-at-USF/open311-client/commit/59b6995f39b5e213b9d5b77329c5ff794011e17b)
* Fix code typos and clean up unused classes
* Implement the same callback method for all reporting fragments

cc'd @barbeau


cagryInside:
Added to `ServiceUtils` method that determines if the given `service` is transit related and coming from `Open311` endpoint.

cc'd @barbeau

barbeau:
@cagryInside Thanks!  So how would I use `isTransitOpen311ServiceByType()`, in the context of showing a tutorial saying that non-transit options are available by changing the category, and only showing this if an Open311 endpoint is available for this location?

cagryInside:
@barbeau I think you can use as following.
In the  `if (transitServiceFound)` block you can check if the selected service is coming from open311:
`if (ServiceUtils.isTransitOpen311ServiceByType(service.getType()))`

```
    private void selectDefaultTransitCategory(ArrayList<SpinnerItem> spinnerItems) {
        int selectPosition = -1;
        for (int i = 0; i < spinnerItems.size(); i++) {
            SpinnerItem item = spinnerItems.get(i);
            if (item instanceof ServiceSpinnerItem) {

                Service service = ((ServiceSpinnerItem) item).getService();

                boolean transitServiceFound = false;

                if (getString(R.string.ri_selected_service_stop).equals(mDefaultIssueType) &&
                        ServiceUtils.isTransitStopServiceByType(service.getType())) {
                    transitServiceFound = true;
                } else if (getString(R.string.ri_selected_service_trip).equals(mDefaultIssueType) &&
                        ServiceUtils.isTransitTripServiceByType(service.getType())) {
                    transitServiceFound = true;
                }

                if (transitServiceFound) {
                    selectPosition = i;
                    // If transit service selected and no bus stop selected remove markers
                    if (ServiceUtils.isTransitServiceByType(service.getType()) &&
                            mIssueLocationHelper.getObaStop() == null) {
                        mIssueLocationHelper.clearMarkers();
                    }
                  if (ServiceUtils.isTransitOpen311ServiceByType(service.getType())) {
                        // Show tutorial "non-transit options are available by changing the category"
                   }
                    break;
                }
            }
        }

        // Set mDefaultIssueType = null to prevent auto selections
        mDefaultIssueType = null;

        // Set selected category if it is in the spinner items list
        if (selectPosition != -1) {
            showServicesSpinner();
            mServicesSpinner.setSelection(selectPosition, true);
        }
    }
```

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/SimpleArrivalListFragment.java
```diff
@@ -127,6 +128,11 @@ public void onActivityCreated(Bundle savedInstanceState) {
         mgr.initLoader(ARRIVALS_LIST_LOADER, getArguments(), this).forceLoad();
     }
 
+    @Override
```
barbeau: If `onAttach()` is just calling `super.onAttach()`, could you please just delete this entire method here?

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/SimpleArrivalListFragment.java
```diff
@@ -273,6 +279,6 @@ private ArrivalsListLoader getArrivalsLoader() {
     }
 
     public void setCallback(Callback callback) {
-        this.callback = callback;
+        this.mCallback = callback;
```
barbeau: This `this.` should no longer be necessary here - could you please remove it?

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportProblemFragmentBase.java
```diff
@@ -237,4 +220,9 @@ public ObaResponse loadInBackground() {
     protected abstract int getLayoutId();
 
     protected abstract ReportLoader createLoader(Bundle args);
+
+    public void setCallback(ReportProblemFragmentCallback callback) {
+        this.mCallback = callback;
```
barbeau: You should be able to remove `this.` here.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportProblemFragmentBase.java
```diff
@@ -105,16 +88,22 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
             // Currently in a layout without a container, so no
             // reason to create our view.
             return null;
         }
+
         return inflater.inflate(getLayoutId(), null);
     }
 
     @Override
+    public void onResume() {
```
barbeau: You should be able to remove `onResume()` here, as it's just making a call to `super`.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/Open311ProblemFragment.java
```diff
@@ -1109,4 +1108,8 @@ public void setArrivalInfo(ObaArrivalInfo arrivalInfo) {
     public void setAgencyName(String agencyName) {
         this.mAgencyName = agencyName;
     }
+
+    public void setCallback(ReportProblemFragmentCallback callback) {
+        this.mCallback = callback;
```
barbeau: You should be able to remove `this.` here.

barbeau:
Thanks @cagryInside, looks pretty good!  Just a few comments inline.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/SimpleArrivalListFragment.java
cagryInside: This is for `onArrivalItemClicked` event defined in `SimpleArrivalListFragment`. I think this is necessary. Because We are setting this callback from its parent activity: https://github.com/cagryInside/onebusaway-android/blob/master_open311_prp/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/SimpleArrivalListFragment.java#L83

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportProblemFragmentBase.java
cagryInside: We are using this setter here: https://github.com/cagryInside/onebusaway-android/blob/master_open311_prp/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportStopProblemFragment.java#L77

barbeau: Sorry, to clarify - you should be able to change this line to:
`mCallback = callback;`

So, just remove the reference to "this" instance of the class.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/SimpleArrivalListFragment.java
barbeau: Sorry, to clarify - you should be able to change this line to:
`mCallback = callback;`

So, just remove the reference to "this" instance of the class.

cagryInside: Oh ok! Thanks

cagryInside:
> Thanks @cagryInside, looks pretty good! Just a few comments inline.

Latest commits should fix the inline comments.

barbeau:
Thanks @cagryInside!

Issue #482, opened 2016-05-06
Issue reporting fragments dosn't finish upon orientation change


cagryInside:
**Summary:** 

Issue reporting fragments (`ReportStopProblemFragment`, `ReportTripProblemFragment`, `Open311ProblemFragment`) don't get closed after orientation change. 

**Steps to reproduce:** 
1. Go to send feedback
2. Fill an issue form
3. Rotate screen
4. Click submit button

**Expected behavior:** 

What did you expect the app to do?

**Observed behavior:** 

It should close the activity and fragments.

**Device and Android version:** 

Nexus 6 w/ Android 6.0.1

PR #483, opened 2016-05-06
Fix #480 and #482 - Restore callback methods on orientation change

### LINKED ISSUE #480: Route favorite dialog crashes upon orientation change
**Summary:** 

Reported via Android Developer Console:

~~~
java.lang.NullPointerException: Attempt to invoke interface method 'void org.onebusaway.android.ui.RouteFavoriteDialogFragment$Callback.onSelectionComplete(boolean)' on a null object reference
	at org.onebusaway.android.ui.RouteFavoriteDialogFragment$1.onClick(RouteFavoriteDialogFragment.java:207)
	at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:162)
	at android.os.Looper.loop(Looper.java:135)
	at android.app.ActivityThread.main(ActivityThread.java:5438)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
~~~

**Steps to reproduce:** 

I was able to reproduce by doing the following:

1. Open any stop
2. Tap on any arrival
3. Select "Add star to route"
4. Rotate the screen
5. Tap on "Save"

**Expected behavior:** 

The app should complete the rotation and allow the user to save the favorite route.

**Observed behavior:** 

The app crashes with the above stacktrace.

I reproduced this on a stock LG G4 with Android 6.0

cagryInside:
This PR fixes two issues (Issue https://github.com/OneBusAway/onebusaway-android/issues/480, Issue https://github.com/OneBusAway/onebusaway-android/issues/482).

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -56,6 +56,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
+import android.support.annotation.Nullable;
 import android.support.v4.app.DialogFragment;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
@@ -265,6 +266,18 @@ public class ArrivalsListFragment extends ListFragment
         initArrivalInfoViews(BuildFlavorUtils.getArrivalInfoStyleFromPreferences(), inflater);
 
         return inflater.inflate(R.layout.fragment_arrivals_list, null);
+    }
+
+    @Override
+    public void onViewStateRestored(@Nullable Bundle savedInstanceState) {
+        super.onViewStateRestored(savedInstanceState);
+        //This is required to avoid an NullPointerException when the user rotates
+        // their phone while setting a route favorite - see #480
+        RouteFavoriteDialogFragment dialogFragment = (RouteFavoriteDialogFragment)
+                getFragmentManager().findFragmentByTag(RouteFavoriteDialogFragment.TAG);
+        if (dialogFragment != null) {
+            setCallbackToDialogFragment(dialogFragment);
+        }
     }
 
     @Override
@@ -718,14 +731,8 @@ public class ArrivalsListFragment extends ListFragment
                             .setFavorite(!isRouteFavorite)
                             .build();
 
-                    routeDialog.setCallback(new RouteFavoriteDialogFragment.Callback() {
-                        @Override
-                        public void onSelectionComplete(boolean savedFavorite) {
-                            if (savedFavorite) {
-                                refreshLocal();
-                            }
-                        }
-                    });
+                    setCallbackToDialogFragment(routeDialog);
+
                     routeDialog.show(getFragmentManager(), RouteFavoriteDialogFragment.TAG);
                 } else if (which == 1) {
                     showRouteOnMap(arrivalInfo);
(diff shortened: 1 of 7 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -268,6 +269,23 @@ public View onCreateView(LayoutInflater inflater,
     }
 
     @Override
+    public void onViewStateRestored(@Nullable Bundle savedInstanceState) {
+        RouteFavoriteDialogFragment dialogFragment = (RouteFavoriteDialogFragment)
+                getFragmentManager().findFragmentByTag(RouteFavoriteDialogFragment.TAG);
+        if (dialogFragment != null) {
```
barbeau: Could you put the code that's a duplicate of lines 721-728 (setting the callback) in a method, and call that method from both locations?  I could see having the duplicate code causing headaches down the road, of someone changing the code in lines 721-728 and then trying to figure out why the behavior isn't changing.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -268,6 +269,23 @@ public View onCreateView(LayoutInflater inflater,
     }
 
     @Override
+    public void onViewStateRestored(@Nullable Bundle savedInstanceState) {
```
barbeau: Could you please add a comment here saying why this is added, and reference the issue number?  Something like "This is required to avoid an NPE when the user rotates their phone while setting a route favorite - see #480".

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -268,6 +269,23 @@ public View onCreateView(LayoutInflater inflater,
...
+            dialogFragment.setCallback(new RouteFavoriteDialogFragment.Callback() {
+                @Override
+                public void onSelectionComplete(boolean savedFavorite) {
+                    if (savedFavorite) {
+                        refreshLocal();
+                    }
+                }
+            });
+        }
+        super.onViewStateRestored(savedInstanceState);
```
barbeau: Seems like the call to `super.onViewStateRestored(savedInstanceState)` might be better as the first line of this method?  In other words, the super implementation should be called before our code?  Not sure about this.

barbeau:
Thanks @cagryInside, looks pretty good.  A few inline comments on the route favorites commit.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
cagryInside: I did a little bit google search on it. Seems most people prefer to call first the super class then run their code. However there is no official recommendation exist for this.

PS: Example google fragment code calls the super class first. http://developer.android.com/reference/android/app/Fragment.html

Therefore, I am going to move `super.onViewStateRestored(savedInstanceState);` to the beginning.

cagryInside:
> Thanks @cagryInside, looks pretty good. A few inline comments on the route favorites commit.

Latest commits fixes inline comments.

barbeau:
@cagryInside Thanks!  Also, could you please rebase this on master?  Looks like there are conflicts now.

cagryInside:
> @cagryInside Thanks! Also, could you please rebase this on master? Looks like there are conflicts now.

I should be good now.

barbeau:

PR #484, opened 2016-05-06
WIP - Fix #29 - Initial work to calculate arrival times using current time from server

### LINKED ISSUE #29: Arrival times can differ due to clock sync issues
Reported by project member paulcwatts, Sep 24, 2010
From a user who is reporting the times on the Android client are different from those on the iPhone. I'll probably need to make sure I'm synchronizing with OBA's system time rather than the phone's time.

Hi Paul, I may have found the cause.  My Droid's clock is about 4 minutes faster than the IPhone's.  4 minutes appears to be the difference between the bus times as well.

FYI.  ONEBUSAWAY's times are consistent with Iphone so you might want to inform Droid to sync with you to avoid it to show a bus has departed when in fact it hasn't.

Thanks, Robbie

On Sep 24, 2010 6:00 PM, "robbie avila" robbieavila98@gmail.com wrote:

> Hi Paul,
> I mostly wait at:
> 
> 6:30 AM: NE 8th St & 164th Ave NE Westbound, Bellevue, 6:30AM
> 
> 3:30 PM: Stevens Way & Construction Pull, Southbound, UW Campus,
> 
> I ride route 272, although I noticed the discrepancy for other routes.
> I also observed the same at other stops when I checked.

Comment 1 by cookieguru, Dec 6, 2010
I can confirm that this is an issue for me as well.  I ran a simple test loading a webpage that runs the current-time API call as well as a JavaScript getter to display the phone's time.  My phone is consistently 76 seconds fast.  I'm on Verizon's network.

This is odd as most carriers have a GPS reciever at the cell site which is used for timing.  I guess timing isn't as important when you're using CDMA versus TDMA.

I'll try and run some more tests, as this morning the route 181 which stayed as "on time" arrived when the countdown displayed -4 minutes which is in line with Robbie's observations.

Comment 2 by cookieguru, Dec 7, 2010
This morning my phone.got down to 78 seconds behind.  However, if one goes to Settings --> Date & time make sure "Automatic" is checked.  Now I'm running 17 seconds fast.

barbeau:
- ONLY use server time if we just got a fresh response from the server.  Otherwise use device current time for cached responses.

TODO:
- Sliding panel header still always uses current time, because we don't currently keep track if the arrival info pulled by the header from the fragment is fresh (in which case server time could be used), or old (in which case device time should be used).

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -344,7 +344,7 @@ public class ArrivalsListFragment extends ListFragment
             ObaArrivalInfoResponse lastGood = loader.getLastGoodResponse();
             if (lastGood != null) {
                 setResponseData(lastGood.getArrivalInfo(), lastGood.getSituations(),
-                        lastGood.getRefs());
+                        lastGood.getRefs(), System.currentTimeMillis());
             }
         }
 
@@ -409,6 +409,8 @@ public class ArrivalsListFragment extends ListFragment
         List<ObaSituation> situations = null;
         ObaReferences refs = null;
 
+        long currentTime;
+
         if (result.getCode() == ObaApi.OBA_OK) {
             if (mStop == null) {
                 mStop = result.getStop();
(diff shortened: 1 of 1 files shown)


barbeau:
Note to myself - we're apparently already using server time as "current time" in the trip reporting flow.  User reported that they had a device clock issue, but trip problem reporting time was still correct (despite ArrivalsListFragment being 1 day off).  I think this is fine - we don't have the refresh and header/list content sync to worry about there, so the implementation is much simpler in trip reporting flow.  So this PR will eventually bring the same feature to the ArrivalsListFragment.

aaronbrethorst:
closing due to merge conflicts and age.

PR #492, opened 2016-05-11
Fix #491: Opening Send Feedback then backing out then re-entering shows prompt again.

### LINKED ISSUE #491: Tapping back in region confirmation dialog is treated as a "yes"
**Summary:** 

In the issue reporting process, when you first tap on "Send feedback" and you are prompted with the "Are you in the Tampa Bay region?", if you tap the back button and then tap on "Send feedback" again, it is treated as though you answered "Yes" to the region confirmation prompt dialog.

**Steps to reproduce:** 

1. Do a fresh install of OBA Android
2. Tap on "Send feedback"
3. Hit the back button when the dialog for "Are you in the Tampa Bay region?" is displayed
4. Tap on "Send feedback" again

**Expected behavior:** 

Prompt me again for the "Are you in the Tampa Bay region?"

**Observed behavior:** 

The list of feedback general categories is displayed, which (I believe) assumes that I answered "Yes" to the region confirmation dialog:

image

Stock LG G4 w/ Android 6.0

**Screenshots:** 

See above

azizmb9494:
Issue #491 fixed. Instead of the Validated Region ID preference being set instantly when the Dialog is showed, I added a member field to ReportActivity to keep track of the current region ID. In RegionDialog, once "Yes" is tapped, the validated region ID preference is set.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportActivity.java
@@ -16,6 +16,13 @@
 
 package org.onebusaway.android.report.ui;
 
+import android.content.Context;
+import android.content.Intent;
+import android.location.Location;
+import android.os.Bundle;
+import android.view.MenuItem;
+import android.view.Window;
+
 import com.google.android.gms.common.api.GoogleApiClient;
 
 import org.onebusaway.android.BuildConfig;
@@ -31,13 +38,6 @@ import org.onebusaway.android.ui.PreferencesActivity;
 import org.onebusaway.android.util.BuildFlavorUtils;
 import org.onebusaway.android.util.LocationUtils;
 import org.onebusaway.android.util.PreferenceUtils;
-
-import android.content.Context;
-import android.content.Intent;
-import android.location.Location;
-import android.os.Bundle;
-import android.view.MenuItem;
-import android.view.Window;
 
 /**
  * Fragment Activity for handling all report
@@ -67,7 +67,6 @@ public class ReportActivity extends BaseReportActivity {
             if (showValidateRegionDialog(currentRegion)) {
                 RegionValidateDialog rvd = new RegionValidateDialog();
                 rvd.show(getSupportFragmentManager(), ReportConstants.TAG_REGION_VALIDATE_DIALOG);
-                PreferenceUtils.saveLong(ReportConstants.PREF_VALIDATED_REGION_ID, currentRegion.getId());
             } else {
                 createIssueTypeListFragment();
             }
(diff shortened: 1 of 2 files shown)


barbeau:
Thanks @azizmb9494!

@cagryInside Could you please take a look at this PR and let me know if it looks good to you?  It should fix #491.  Feel free to leave comments in-line.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportActivity.java
```diff
@@ -65,9 +66,9 @@ protected void onCreate(Bundle savedInstanceState) {
             ObaRegion currentRegion = Application.get().getCurrentRegion();
 
             if (showValidateRegionDialog(currentRegion)) {
+                mRegionId = currentRegion.getId();
```
cagryInside: @azizmb9494 Could you please use `Bundle` to pass data to `RegionValidateDialog` fragment? Something like:

```
Bundle bundle = new Bundle();
bundle.putLong(key, regionId);
rvd.setArguments(bundle);
```
I think, we should try to avoid creating unnecessary members.

cagryInside:
Thanks @azizmb9494, looks pretty good. Just one inline comment.

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportActivity.java
barbeau: Agreed, this would also be a cleaner separation between the current hosting Activity and the Fragment.

azizmb9494:
Thanks @cagryInside! One question, RegionValidationDialog uses "Application.get().getCurrentRegion()" when building the message string, would it be safe just to get the ID from the current region like this on "Yes" tapped?

cagryInside:
> Thanks @cagryInside! One question, RegionValidationDialog uses "Application.get().getCurrentRegion()" when building the message string, would it be safe just to get the ID from the current region like this on "Yes" tapped?

Good point! Yes, it would be safe to use, because we already check the `region` before opening the `fragment`.

azizmb9494:
Thanks @barbeau & @cagryInside! Fixed and pushed!

cagryInside:
@azizmb9494 thanks. It looks pretty good now. 
One small thing, could you please add the issue number to the commit message? Something like `"Fix #492 - Validated Region ID only set on tap instead."` Like that, Github will automatically link your commit with the issue and it will be easier to track.

PS: See (this)[http://stackoverflow.com/questions/179123/edit-an-incorrect-commit-message-in-git] post to edit incorrect commit message.

barbeau:
Agreed - doing "Fix #X" also automatically closes the issue on Github when the commit is merged - so overall makes our lives easier :).

azizmb9494:
Completed. Thanks @cagryInside, @barbeau!

barbeau:
Thanks @azizmb9494!

Issue #498, opened 2016-05-16
Scheduled arrival time for a stop is truncated


hypergreatthing:
Information displaying schedule arrival time is off by a minute in some cases.

I'm a little bit of an outsider in how the structure of this application is created, but i assume you're using the oba api to pull stop information that includes scheduled arrival times.

Stop level arrival information is merely estimated time between timepoints using distance.  The stop times are in seconds.

It's my understanding that this code: https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
on line 228 is truncating the time values to minutes instead of rounding.
This causes a problem where a stop has the arrival time in gtfs of 3:42:53 but is showing up on the app as 3:42.

If it can only be displayed by the minute, at least the closest minute would be preferable.

barbeau:
@hypergreatthing Good point!  Do you have a pointer to a specific region/route/stop where this is visible?  Would like to grab some data to create some unit tests.  Screenshot would help too.

I'm surprised this issue has never come up before, although maybe most agencies either round schedules to the minute or rounding down has been mostly accurate with agency data.

barbeau:
This should also affect departure times too.

barbeau:
And from a quick scan of the code on my phone I think real-time too.

hypergreatthing:
WMATA: stop id 14078
I'm not sure if you use the siri call for stop info but this url will work:
http://buseta.wmata.com/api/stop-for-id?_=1463427413328&stopId=1_14078

I'll have to get a screenshot later.

barbeau:
Out of curiosity - does this display correctly in the other OBA UIs (iPhone, web, etc.)?  Would help to have something accurate to test against.

hypergreatthing:
![rosslyn](https://cloud.githubusercontent.com/assets/10711060/15324417/5e06e160-1c14-11e6-9658-00c37f0cad3a.jpg)
From GTFS:
868586830	13:49:57	13:49:57	14078	N MOORE ST + 19TH ST
868588830	14:29:57	14:29:57	14078	N MOORE ST + 19TH ST
Timetable:
http://www.wmata.com/bus/timetables/dc/05a.pdf?n

I don't think oba web shows scheduled arrival times when they have real time data.  When they do it's relative time (how many minutes until it should arrive).  I will ask however to make sure they're displaying the right data as it's also to the minute.

[...]

Issue #409, opened 2016-01-04
Improve display of ferry data


bbodenmiller:
Per https://github.com/SoundTransit/soundtransit-rds/issues/22#issuecomment-167672625 real-time ferry data is now in OBA. When looking at stop 7 I can't tell which rows corispond to which ferries as multiple have the same headsign and no route number:

![screenshot_20160104-103116](https://cloud.githubusercontent.com/assets/1192780/12097077/81b5469c-b2ce-11e5-8d79-3c5c296f74eb.png)

Additionally selecting view route map is resulting in:
![screenshot_20160104-103130](https://cloud.githubusercontent.com/assets/1192780/12097109/a7523194-b2ce-11e5-9251-3b362dca05f8.png)

barbeau:
@bbodenmiller Interesting, and cool that real-time ferry data is now included!  I'm guessing this is a GTFS data issue, as `route_short_name` and `route_long_name` are both required in the GTFS spec - or there is some really weird formatting in the text.  I can't seem to pull up a "stop 7" in Puget Sound - can you hit the menu button and do "Show Stop Details" to get the stop code?  Or, even better, if you could hook up to Android Studio and run a development build, and grab the API URL it's using.  Or, paste the link from the OBA Puget Sound website showing the arrival info for this stop and I'll grab it from there.

bbodenmiller:
![screenshot_20160104-130739](https://cloud.githubusercontent.com/assets/1192780/12100510/49dfb0ca-b2e4-11e5-9990-a6be0b450155.png)
![screenshot_20160104-130815](https://cloud.githubusercontent.com/assets/1192780/12100512/51acab28-b2e4-11e5-9d90-bccaa36249e9.png)

barbeau:
Ah, got it - the agency_id is 95 - http://pugetsound.onebusaway.org/where/standard/stop.action?id=95_7.

Here's the API request:
http://api.pugetsound.onebusaway.org/api/where/arrivals-and-departures-for-stop/95_7.json?key=TEST

Looks like `route_short_name` isn't provided in the response:

~~~
data: {
entry: {
arrivalsAndDepartures: [
{
arrivalEnabled: true,
blockTripSequence: 0,
departureEnabled: false,
distanceFromStop: 601.124216990771,
frequency: null,
lastUpdateTime: 0,
numberOfStopsAway: 0,
predicted: false,
predictedArrivalInterval: null,
predictedArrivalTime: 0,
[...]

Issue #333, opened 2015-10-08
Allow users to keep the same reminders across GTFS updates


barbeau:
Currently reminders are identified internally by OBA Android by their `trip_id` and `stop_id`.  When a user creates a reminder for a trip, that `trip_id` and `stop_id` are saved to the local database.

Then, for a particular `stop_id`, OBA will occasionally poll the estimated arrival times for that stop, search the results for an arrival for the saved `trip_id`, and then alert the user with the estimated arrival info for that `trip_id`.

The problem with this approach is that GTFS `trip_ids` typically don't persist across GTFS data updates.  So, if the agency releases a new schedule/GTFS dataset, the `trip_id` for a user's reminder in OBA Android may no longer exist in the new data - instead, the "same" trip has a new `trip_id` (this `trip_id` changes because it is tied to a new date range of service, which is different than the old range).  As a result, the user would never again be reminded for that trip.  To fix this, the user has to delete and re-create all their reminders.

Feedback from a user:

> Please somebody get the reminders fixed so we don't have to freaking rename and re-add all of them every stinking time that metro sends out a rider alert or tweaks the bus schedule.  This is really a pain in the butt to have to go through the process when you have about 10-15 named reminders. Please help fix this & thank you.

So, it would be nice to come up with a way to match the "same" trip, which has two different `trip_ids`, across GTFS updates.

The challenges:
1. In OBA Android, we have no way of knowing that the old `trip_id` is no longer valid - there is no indication that a schedule change has occurred and a particular `trip_id` is no longer valid.  So we'd need to continue looking for the old `trip_id`.
2. How to determine that two trips across two GTFS datasets are the "same"

One possible solution is to hash the relevant data for a particular trip (e.g., headsign, scheduled arrival time), and then compare trip hashes as well as the `trip_id` itself.  If either the `trip_id` or the trip hash match, then we would alert the user.  This would allow the user to keep the same reminder in OBA Android, and it _should_ match to the "same" trips in the future.

A similar issue for server-side trip hashes is discussed here:
https://github.com/opentripplanner/OpenTripPlanner/issues/1573

We should look at the hashing algorithm implemented in OTP, as well as the one that I commented on, and see what might work for OBA Android.

bmander:
Obsolete — reminders are now server-side alarms/push (`api/contract/ReminderWebService.kt` `createAlarm`) rather than client-side trip_id polling, so the GTFS-update drift this described is now a server concern rather than a client mechanism. Closing.

Issue #184, opened 2014-10-15
Notify me with an arrival time when I'm near favorite stops


barbeau:
Requested by a few users here in Tampa.  When the user nears a favorite stop, OBA should pop up a notification with real-time arrival info for that stop.

We should be able to do this in an energy-efficient manner using [Geofences in Google Play Services](http://developer.android.com/training/location/geofencing.html), since the proximity detection is offloaded to hardware.  We should notify a user on entry to radius near stop, to avoid pinging them multiple times while they are near the stop.  "Near" still needs to be defined.

We should also have a setting to turn this off, if it annoys some users, or becomes an energy hog on older devices that don't support geofences in hardware.

barbeau:
Now that we have favorite routes as well, we could only notify users at favorite stops for favorite routes, to help cut down on chatter for routes they don't care about.

PR #499, opened 2016-05-17
Fix  #496 - Correctly reflect arrival/departure information in reminder text

### LINKED ISSUE #496: Reminder text doesn't accurately reflect arrival/departure information
**Summary:** 

When viewing the reminder for a trip that has just arrived/departed, the reminder always says "Route X has departed", even if the trip was incoming to a transit center.  In that case, the reminder should say "Route X has arrived".

This doesn't affect trips that haven't yet arrived/departed - the text says "Route X is 5 min away".

This is related to issue https://github.com/OneBusAway/onebusaway-android/issues/287 for better distinguishing between arrivals and departures in the main arrival list, which was fixed in PR https://github.com/OneBusAway/onebusaway-android/pull/449.

**Steps to reproduce:** 
1. Set an alarm for a trip that has a negative arrival time for an arrival trip (e.g., an inbound trip at a transit center).  The arrival time in the `ArrivalListFragment` should say "Arrived at 2:00 PM".
2. Wait for the notification to trigger

**Expected behavior:** 

Reminder should read "Route X has arrived".

**Observed behavior:** 

Reminder says "Route X has departed".

Stock LG G4 w/ Android 6.0

**Screenshots:** 

Here's the reminder saying "Route 5 has departed" for an arrival (you can see the "Arrived at" text in the background) :

image

Here's an upcoming route that hasn't yet arrived/departed - the text is more generic so it fits both cases - if it makes sense, we might want to change this to "arrived/departed" as well:

image

cagryInside:
Add arrival and departure information to reminder notification text:

device-2016-05-16-222541

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2012-2015 Paul Watts (paulcwatts@gmail.com), University of South Florida,
+ * Copyright (C) 2012-2016 Paul Watts (paulcwatts@gmail.com), University of South Florida,
  * Benjamin Du (bendu@me.com), and individual contributors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -416,7 +416,7 @@ public class ArrivalsListFragment extends ListFragment
     @Override
     public void onLoadFinished(Loader<ObaArrivalInfoResponse> loader,
                                final ObaArrivalInfoResponse result) {
-        UIUtils.showProgress(this, false);
+        showProgress(false);
 
         ObaArrivalInfo[] info = null;
         List<ObaSituation> situations = null;
@@ -494,6 +494,14 @@ public class ArrivalsListFragment extends ListFragment
         showTutorials(result);
     }
 
+    @Override
+    protected void showProgress(boolean visibility) {
+        super.showProgress(visibility);
+        if (mHeader != null) {
+            mHeader.showProgress(visibility);
+        }
+    }
+
     private void showTutorials(ObaArrivalInfoResponse response) {
         // If we're already showing a ShowcaseView, we don't want to stack another on top
         if (ShowcaseViewUtils.isShowcaseViewShowing()) {
@@ -566,7 +574,7 @@ public class ArrivalsListFragment extends ListFragment
 
     @Override
     public void onLoaderReset(Loader<ObaArrivalInfoResponse> loader) {
-        UIUtils.showProgress(this, false);
+        showProgress(false);
         mAdapter.setData(null, mRoutesFilter, System.currentTimeMillis());
 
         mArrivalInfo = null;
@@ -1057,6 +1065,7 @@ public class ArrivalsListFragment extends ListFragment
 
     /**
      * Sets up the load more arrivals button in the empty list view
+     *
      * @param currentText the text that should populate the empty list entry, or null if no text
      *                    should be set at this point
      */
@@ -1441,7 +1450,7 @@ public class ArrivalsListFragment extends ListFragment
      */
     private void refresh() {
         if (isAdded()) {
-            UIUtils.showProgress(this, true);
+            showProgress(true);
             // Get last response length now, since its overwritten within
             // ArrivalsListLoader before onLoadFinished() is called
             ObaArrivalInfoResponse lastGood =
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListHeader.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListHeader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2014 Paul Watts (paulcwatts@gmail.com),
+ * Copyright (C) 2011-2016 Paul Watts (paulcwatts@gmail.com),
  * University of South Florida (sjbarbeau@gmail.com)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -1314,6 +1314,17 @@ class ArrivalsListHeader {
         }
         if (mPopup2 != null) {
             mPopup2.dismiss();
+        }
+    }
+
+    public void showProgress(boolean visibility) {
+        if (mProgressBar == null)
+            return;
+
+        if (visibility) {
+            mProgressBar.setVisibility(View.VISIBLE);
+        } else {
+            mProgressBar.setVisibility(View.GONE);
         }
     }
 }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ListFragment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011 The Android Open Source Project
+ * Copyright (C) 2011-2016 The Android Open Source Project, University of South Florida
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,6 +66,8 @@ public class ListFragment extends Fragment {
     TextView mStandardEmptyView;
 
     View mProgressContainer;
+
+    View mRefreshProgressContainer;
 
     View mListContainer;
 
@@ -159,7 +161,7 @@ public class ListFragment extends Fragment {
         mHandler.removeCallbacks(mRequestFocus);
         mList = null;
         mListShown = false;
-        mEmptyView = mProgressContainer = mListContainer = null;
+        mEmptyView = mProgressContainer = mRefreshProgressContainer = mListContainer = null;
         mStandardEmptyView = null;
         super.onDestroyView();
     }
@@ -341,6 +343,13 @@ public class ListFragment extends Fragment {
                 mStandardEmptyView.setVisibility(View.GONE);
             }
             mProgressContainer = root.findViewById(R.id.loading);
+            mRefreshProgressContainer = root.findViewById(R.id.refresh_loading);
+            // Set %50 transparency to progress bar
+            ProgressBar progressBar = (ProgressBar) root.findViewById(R.id.refresh_progress_small);
+            if (progressBar != null && progressBar.getIndeterminateDrawable() != null) {
+                progressBar.getIndeterminateDrawable().setAlpha(128);
+            }
+
             mListContainer = root.findViewById(R.id.listContainer);
             View rawListView = root.findViewById(android.R.id.list);
             if (!(rawListView instanceof ListView)) {
@@ -375,5 +384,16 @@ public class ListFragment extends Fragment {
             }
         }
         mHandler.post(mRequestFocus);
+    }
+
+    protected void showProgress(boolean visibility) {
+        if (mRefreshProgressContainer == null)
+            return;
+
+        if (visibility) {
+            mRefreshProgressContainer.setVisibility(View.VISIBLE);
+        } else {
+            mRefreshProgressContainer.setVisibility(View.GONE);
+        }
     }
 }
--- a/onebusaway-android/src/main/res/layout/fragment_arrivals_list.xml
+++ b/onebusaway-android/src/main/res/layout/fragment_arrivals_list.xml
@@ -35,5 +35,7 @@
                   style="@style/ListHintText"
                   android:layout_width="fill_parent"
                   android:layout_height="wrap_content"/>
+
+        <include layout="@layout/refresh_loading"/>
     </FrameLayout>
 </LinearLayout>
--- a/onebusaway-android/src/main/res/layout/refresh_loading.xml
+++ b/onebusaway-android/src/main/res/layout/refresh_loading.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (C) 2010-2016 Paul Watts (paulcwatts@gmail.com), University of South Florida
+
+     Licensed under the Apache License, Version 2.0 (the "License");
+     you may not use this file except in compliance with the License.
+     You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
+
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              android:id="@+id/refresh_loading"
+              android:orientation="horizontal"
+              android:layout_width="fill_parent"
+              android:layout_height="fill_parent"
+              android:gravity="center|top"
+              android:layout_marginTop="20dp"
+              android:visibility="gone">
+    <ProgressBar android:id="@+id/refresh_progress_small"
+                 style="@style/LoadingProgress"
+                 android:layout_width="wrap_content"
+                 android:layout_height="wrap_content"
+                 android:paddingLeft="5dp"/>
+</LinearLayout>


onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -396,10 +396,15 @@
...
-    <string name="trip_stat">Route %1$s is %2$d minutes away!</string>
+    <!--  Trip notifications -->
+    <string name="trip_stat_gone_arrived">Route %s has arrived.</string>
+    <string name="trip_stat_gone_departed">Route %s has departed.</string>
+    <string name="trip_stat_lessthanone_arriving">Route %s is arriving in less than a minute!</string>
+    <string name="trip_stat_lessthanone_departing">Route %s is departing in less than a minute!</string>
+    <string name="trip_stat_one_arriving">Route %s is arriving in 1 minute!</string>
+    <string name="trip_stat_one_departing">Route %s is departing in 1 minute!</string>
+    <string name="trip_stat_arriving">Route %1$s is arriving in %2$d minutes!</string>
+    <string name="trip_stat_departing">Route %1$s is departing in %2$d minutes!</string>
```
cagryInside: @barbeau I used these texts to reflect arrival and departure information. All suggestions are welcome.

barbeau: Let's actually change the following strings to abbreviate minutes to min:

~~~
<string name="trip_stat_arriving">Route %1$s is arriving in %2$d min!</string>
<string name="trip_stat_departing">Route %1$s is departing in %2$d min!</string>
~~~

This should eliminate `trip_stat_one_arriving` and `trip_stat_one_departing` because we don't need singular vs. plural anymore.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -396,10 +396,15 @@
...
 
     <!--  Trip notifications -->
-    <string name="trip_stat_gone">Route %s has departed.</string>
-    <string name="trip_stat_lessthanone">Route %s is less than a minute away!</string>
-    <string name="trip_stat_one">Route %s is 1 minute away!</string>
-    <string name="trip_stat">Route %1$s is %2$d minutes away!</string>
+    <!--  Trip notifications -->
+    <string name="trip_stat_gone_arrived">Route %s has arrived.</string>
+    <string name="trip_stat_gone_departed">Route %s has departed.</string>
+    <string name="trip_stat_lessthanone_arriving">Route %s is arriving in less than a minute!</string>
```
barbeau: Let's change this and the following string to:

~~~
<string name="trip_stat_lessthanone_arriving">Route %s is arriving now!</string>
<string name="trip_stat_lessthanone_departing">Route %s is departing now!</string>
~~~

...so it matches with what we show in the arrivals list.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -120,18 +121,19 @@ private void poll1(Cursor c) {
         ObaArrivalInfoResponse response = ObaArrivalInfoRequest
                 .newRequest(mContext, stopId).call();
 
-        Long departMS = null;
+        // Pair of departMs and is arriving information
+        Pair<Long, Boolean> depart = null;
```
barbeau: Let's change the name of `depart`, because it could no longer be a departure.  Maybe `eta`?  Although that implies an arrival.  I'd welcome any other naming ideas.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -156,14 +158,16 @@ private Long checkArrivals(ObaArrivalInfoResponse response, Cursor c) {
                 // when calculated with the *predicted* arrival time
                 // is past now.
                 long depart = info.getPredictedArrivalTime();
+                boolean isArriving = true;
                 if (depart == 0) {
                     depart = info.getScheduledArrivalTime();
+                    isArriving = false;
```
barbeau: To keep our arrival/departure labeling consistent with the normal arrival times lists, let's construct an instance of `ArrivalInfo` here from the `ObaArrivalInfo`, and then call `arrivalInfo.isArrival()` to determine if it's an arrival or departure.

This is actually a little more complicated than I was originally thinking, though, as previously we were always using arrival time for the reminder - technically I think we should probably be reminding users with the departure times for departures, and arrival time for arrivals.  I think we could use `arrivalInfo.getEta()` for this (internally we already calculate ETA from arrival or departure appropriately), but the resolution for `getEta()` is currently in seconds, and this method expects to return milliseconds.  Maybe we add another method that returns ETA in milliseconds.

barbeau:
@cagryInside Thanks for working on this.  A few comments in-line - it's a little more complicated than I thought because we also need to account for calculating the reminder time using either departure or arrival info (currently we're always using the arrival info).

Also, heads up that this will create a conflict with https://github.com/OneBusAway/onebusaway-android/pull/479.  Let's get https://github.com/OneBusAway/onebusaway-android/pull/479 merged first and make sure that's working correctly, and then rebase this PR on top of that.

I think I'm going to go ahead and push a new Android release based on the current master branch to get some of those bug fixes out, and then we can target this and https://github.com/OneBusAway/onebusaway-android/pull/479 in the next release.

barbeau:
@cagryInside Ok, #479 is merged, and I've added some better documentation surrounding reminders as well - these introduced conflicts.  I'm going to push out a new beta with #479, and then we can push this out in a subsequent release.  This will give us some separation as well between reminders changes, so if they stop working in the new release we know why.  So, please feel free to rebase at your leisure.

cagryInside:
> @cagryInside Ok, #479 is merged, and I've added some better documentation surrounding reminders as well - these introduced conflicts. I'm going to push out a new beta with #479, and then we can push this out in a subsequent release. This will give us some separation as well between reminders changes, so if they stop working in the new release we know why. So, please feel free to rebase at your leisure.

@barbeau latest commits rebase the upstream and fix the inline comments.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -150,7 +153,7 @@ private long getReminderMS(String tripId, String stopId) {
     // Return the difference between now and the predicted/scheduled
```
barbeau: Since we're modifying this method, can you please fix this Javadoc?  If I'm reading the code right, this doesn't return the difference between now and predicted/scheduled time - instead it returns either the predicted or scheduled time of this particular `tripId`, or null if that tripId can't be found.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -125,18 +127,19 @@ private void poll1(Cursor c) {
...
+        Pair<Long, Boolean> eta = null;
         if (response.getCode() == ObaApi.OBA_OK) {
-            departMS = checkArrivals(response, c);
+            eta = checkArrivals(response, c);
         }
 
-        if (departMS != null) {
-            final long diffTime = departMS - System.currentTimeMillis();
+        if (eta != null) {
+            final long diffTime = eta.first - System.currentTimeMillis();
```
barbeau: Instead of `System.currentTimeMillis()`, let's use `response.getCurrentTime()`.  This helps us move towards #29 and using current server time instead of device local time to avoid clock sync issues.  Please reference #29 in the commit description as well, so we can keep track of this progress.  This won't fix that issue entirely but will help move us in that direction.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -396,10 +396,13 @@
     <string name="trip_list_context_showroute">Show route</string>
 
     <!--  Trip notifications -->
```
barbeau: Minor item - could you please remove the extra comment?

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -165,10 +168,11 @@ private Long checkArrivals(ObaArrivalInfoResponse response, Cursor c) {
                     depart = info.getScheduledArrivalTime();
                 }
 
-                return depart;
+                ArrivalInfo arrivalInfo = new ArrivalInfo(mContext, info, 0l, false);
```
barbeau: I think `ArrivalInfo arrivalInfo = new ArrivalInfo(mContext, info, 0l, false);` should be `ArrivalInfo arrivalInfo = new ArrivalInfo(mContext, info, response.getCurrentTime(), false);`?  Any reason not to pass in the current time here?  If now, I'd prefer to construct this with actual data, in case someone comes along later and tries to use one of the methods that depends on the current time.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/NotifierTask.java
```diff
@@ -161,23 +165,34 @@ private Notification createNotification(String routeId,
...
                 .setDeleteIntent(deleteIntent)
                 .setContentTitle(title)
-                .setContentText(getNotifyText(routeId, timeDiff))
+                .setContentText(getNotifyText(routeId, timeDiff, isArriving))
                 .build();
 
     }
 
-    private String getNotifyText(String routeId, long timeDiff) {
+    private String getNotifyText(String routeId, long timeDiff, boolean isArriving) {
```
barbeau: The more I think about this, the more I think the contents of this method should be refactored into `ArrivalInfo` as part of the logic in the constructor.  It looks a lot like the code to generate the status and time labels there, and we should try to keep all the code that generates text from arrival/departure info in one place - we will likely use the reminder text elsewhere (e.g., notifying someone when they approach a stop - https://github.com/OneBusAway/onebusaway-android/issues/184).  You might actually be able to generate the same text from the status text IF/THEN block.  Could you please refactor this code there, and add a few unit test asserts to `UIUtilTest.testArrivalInfoLabels()`?

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
barbeau: After my last comment about refactoring text generation logic to `ArrivalInfo`, we'll definitely need to pass in `response.getCurrentTime()` here.

barbeau:
@cagryInside Thanks for your work on this!  A few comments in-line - major one is that we should refactor the reminder text generation logic to be part of `ArrivalInfo`.  Please let me know if you have any questions about this!

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/NotifierTask.java
barbeau: FYI - for the unit test in `UIUtilTest.testArrivalInfoLabels()`, the response in the code obtained from `"Hillsborough Area Regional Transit_6497"` is actually mocked from this file - `/onebusaway-android/src/androidTest/res/raw/arrivals_and_departures_for_stop_hart_6497.json`.

cagryInside:
@barbeau latest two commits fixes the inline comments and refactors `ArrivalInfo`

barbeau:
@cagryInside Thanks for the updates!  The latest build [failed on Travis](https://travis-ci.org/OneBusAway/onebusaway-android/builds/132913064) - looks like maybe it's just a white-space issue in `UIUtilTest`?

>org.onebusaway.android.util.test.UIUtilTest > testArrivalInfoLabels[test(AVD) - 5.1] FAILED 
	junit.framework.ComparisonFailure: expected:<... is departing in 35 [  ]min!> but was:<... is departing in 35 []min!>
	at org.onebusaway.android.util.test.UIUtilTest.testArrivalInfoLabels(UIUtilTest.java:625)

cagryInside:
> @cagryInside Thanks for the updates! The latest build failed on Travis - looks like maybe it's just a white-space issue in UIUtilTest?

Yes, sorry. It should be good now.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
```diff
@@ -451,31 +252,28 @@ private String computeTimeLabel(Context context) {
...
-                    delay);
-        } else if (delay < 0) {
-            // Arriving early
-            delay = -delay;
-            return res
-                    .getQuantityString(
-                            R.plurals.stop_info_arrive_early,
-                            (int) delay, delay);
+    private String computeNotifyText(Context context) {
+        final String routeName = mInfo.getShortName();
```
barbeau: This should be changed to:

`final String displayRouteName = UIUtils.getRouteDisplayName(mInfo.getRouteLongName(), mInfo.getShortName());`

Related unit tests will also need to be updated.

Only `route_long_name` OR `route_short_name` is required by GTFS, so unfortunately we can't count on `route_short_name` always existing.  I recently hit this issue in https://github.com/OneBusAway/onebusaway-android/issues/409#issuecomment-169457263 with the Seattle ferries, for which `route_short_name` doesn't exist.

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
```diff
@@ -552,6 +553,75 @@ public void testArrivalInfoLabels() {
         assertEquals("On time", arrivalInfo.get(29).getStatusText());
         assertEquals("9 min delay", arrivalInfo.get(30).getStatusText());
         assertEquals("On time", arrivalInfo.get(31).getStatusText());
+
+        /**
+         * Test notification texts
+         */
+
+        assertEquals("Route " + arrivalInfo.get(0).getInfo().getShortName()
```
barbeau: Need to be updated to use `UIUtils.getRouteDisplayName()` instead of `getShortName()`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
```diff
@@ -451,31 +252,28 @@ private String computeTimeLabel(Context context) {
...
+        if (mEta <= 0) {
```
barbeau: I think this should be:

`if (mEta < 0) {`

In `computeTimeLabel`, we have the following logic:

~~~
        if (mEta >= 0) {
            // Bus hasn't yet arrived
            ...
            }
        } else {
            // Arrival/departure time has passed
            ...
        }
~~~

Similar IF/ELSE block is also in `computeStatusLabel()`.

If you could add a comment here saying "// Bus hasn't yet arrived or departed" as well that would be helpful.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
```diff
@@ -451,31 +252,28 @@ private String computeTimeLabel(Context context) {
...
+            if (mIsArrival) {
+                return context.getString(R.string.trip_stat_gone_arrived, routeName);
+            } else {
+                return context.getString(R.string.trip_stat_gone_departed, routeName);
+            }
+        } else if (mEta < ONE_MINUTE) {
```
barbeau: Related to the above comment - I think this should be:

`} else if (mEta == 0) {`

With the current values, I don't think this would ever fire - `mEta` is rounded to the minute, and the preceding IF statement would catch any equivalence to 0, so we couldn't have any floating point values between `0` and `1`.

I think with these two changes, it should calculate "NOW" arrivals and departures correctly.

On a related note, I don't think there are currently any unit tests that cover this - we should add some to make sure all the IF branches are covered - we can use another raw mocked response if needed.

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
```diff
@@ -552,6 +553,75 @@ public void testArrivalInfoLabels() {
...
+        assertEquals("Route " + arrivalInfo.get(1).getInfo().getShortName()
+                + " has departed.", arrivalInfo.get(1).getNotifyText());
+        assertEquals("Route " + arrivalInfo.get(2).getInfo().getShortName()
+                + " has departed.", arrivalInfo.get(2).getNotifyText());
+        assertEquals("Route " + arrivalInfo.get(3).getInfo().getShortName()
+                + " has arrived.", arrivalInfo.get(3).getNotifyText());
+        assertEquals("Route " + arrivalInfo.get(4).getInfo().getShortName()
+                + " has departed.", arrivalInfo.get(4).getNotifyText());
+        assertEquals("Route " + arrivalInfo.get(5).getInfo().getShortName()
+                + " has departed.", arrivalInfo.get(5).getNotifyText());
```
barbeau: Related to my comment in `computeNotifyText()` - I don't think there are currently any unit tests that cover the "NOW" case of an arrival and departure - we should add some to make sure all the IF branches are covered - we can use another raw mocked response if needed.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
barbeau: Yes, I just looked at `ArrivalsListAdapterStyleA.initView()` and `SimpleArrivalListFragment.loadArrivalList()` both of them have the following logic:

~~~
        long eta = stopInfo.getEta();
        if (eta == 0) {
            etaView.setText(R.string.stop_info_eta_now);
            minView.setVisibility(View.GONE);
        } else {
            etaView.setText(String.valueOf(eta));
            minView.setVisibility(View.VISIBLE);
        }
~~~

So testing `mEta` equivalence to `0` is the proper way to see if the bus is arriving or departing "now".

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
barbeau: Could you actually change the order of the IF/ELSE blocks to match `computeTimeLabel()` and `computeStatusLabel()`?  I think its easier to follow the logic if we always use the same order.  I actually like your order here better as it's chronological, but given the number of other changes let's avoid refactoring `computeTimeLabel()` and `computeStatusLabel()` for now.  We can refactor all three methods later together.

barbeau:
@cagryInside Thanks for working on this!  I really like all the public static methods refactored out of `ArrivalInfo` into the utility class - much cleaner.  A few comments in-line that need to be addressed but all around looks pretty good.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -120,55 +121,51 @@ private void poll1(Cursor c) {
...
         if (response.getCode() == ObaApi.OBA_OK) {
-            departMS = checkArrivals(response, c);
+            arrivalInfo = checkArrivals(response, c);
         }
 
-        if (departMS != null) {
-            final long diffTime = departMS - System.currentTimeMillis();
-            if (diffTime <= reminderMS) {
+        if (arrivalInfo != null) {
+            if (arrivalInfo.getEta() <= reminderMin) {
```
barbeau: We're losing precision in our calculation here by changing from milliseconds to minutes, but I think it's ok in this context - our drop down to choose reminder times is rounded to the minute, and the values stored in the database for the reminder is also rounded to minutes.  I'd rather be consistent in calculating ETA throughout the application, and we can tackle application-wide precision increases (e.g., https://github.com/OneBusAway/onebusaway-android/issues/498) separately.

onebusaway-android/src/main/java/org/onebusaway/android/tripservice/PollerTask.java
```diff
@@ -120,55 +121,51 @@ private void poll1(Cursor c) {
...
-    //
-    private Long checkArrivals(ObaArrivalInfoResponse response, Cursor c) {
+    /**
+     * Checks arrivals from given ObaArrivalInfoResponse
+     *
+     * @param response arrival information
+     * @param c        holds the stored trip_id
+     * @return ArrivalInfo, or return null if the arrival can't be found.
+     */
+    private ArrivalInfo checkArrivals(ObaArrivalInfoResponse response, Cursor c) {
```
barbeau: I think we should pass in the `tripId` here, rather than the `Cursor c`.  It provides better separation from the underlying database.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalInfo.java
cagryInside: > Related to the above comment - I think this should be:
> 
> } else if (mEta == 0) {
> 
> With the current values, I don't think this would ever fire - mEta is rounded to the minute, and the preceding IF statement would catch any equivalence to 0, so we couldn't have any floating point values between 0 and 1.
> 
> I think with these two changes, it should calculate "NOW" arrivals and departures correctly.
> 
> On a related note, I don't think there are currently any unit tests that cover this - we should add some to make sure all the IF branches are covered - we can use another raw mocked response if needed.

@barbeau this is a really good catch. I just copy the old code and didn't realize at all. So, based on this one and the previous comments, I changed the logic as following:

```
        if (mEta == 0) {
            // Bus is arriving now
            if (mIsArrival) {
                return context.getString(R.string.trip_stat_lessthanone_arriving, routeDisplayName);
            } else {
                return context.getString(R.string.trip_stat_lessthanone_departing, routeDisplayName);
            }
        } else if (mEta > 0) {
            // Bus hasn't yet arrived
            if (mIsArrival) {
                return context.getString(R.string.trip_stat_arriving, routeDisplayName,
                        (int) (mEta));
            } else {
                return context.getString(R.string.trip_stat_departing, routeDisplayName,
                        (int) (mEta));
            }
        } else {
            // Bus arrived or departed
            if (mIsArrival) {
                return context.getString(R.string.trip_stat_gone_arrived, routeDisplayName);
            } else {
                return context.getString(R.string.trip_stat_gone_departed, routeDisplayName);
            }
        }
```

cagryInside:
@barbeau I pushed another commit that fixes in-line comments.

barbeau:
@cagryInside Sorry, but this PR has conflict now (it's one we discussed in our last meeting - just a friendly note :) ).  Would you mind rebasing please when you get a chance?  Thanks!

cagryInside:
@barbeau just rebased, It should be good now.

bbodenmiller:
@cagryInside looks like new conflict now.

cagryInside:
> @cagryInside looks like new conflict now.

The new conflict is now resolved.

barbeau:
@cagryInside Would you be able to rebase this in the next week or so?  I'd like to work on the reminder/changing tripID issue (https://github.com/OneBusAway/onebusaway-android/issues/333), but that will create more conflicts.  So, I'd prefer to merge this first, and then work on #333.

cagryInside:
> @cagryInside Would you be able to rebase this in the next week or so? I'd like to work on the reminder/changing tripID issue (#333), but that will create more conflicts. So, I'd prefer to merge this first, and then work on #333.

@barbeau I rebased this PR. But after I force pushed, my commits disappeared, and the PR was closed. So, I pushed this development into another branch and opened a new PR https://github.com/OneBusAway/onebusaway-android/pull/707. Please let me know if there are anything else to be done. I will be happy to contribute.

PR #502, opened 2016-05-19
Fix  #500 - Add analytics to the service alerts button

### LINKED ISSUE #500: Add analytics to the "service alerts button"
**Summary:** 

We would like to report how many times a service alert button is clicked. Therefore, we need to add analytics when a service alert button is clicked. We should also include the agency id to this report.

cagryInside:
Following analytics report was added when a `service alerts button` is clicked:

| ui_action  | service_alerts  | Clicked Service Alerts from: PSTA |

Screenshot from GA console:

untitled 2

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -79,7 +79,9 @@ import android.widget.Toast;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 //
 // We don't use the ListFragment because the support library's version of
@@ -1612,6 +1614,7 @@ public class ArrivalsListFragment extends ListFragment
         @Override
         public void onClick() {
             SituationActivity.start(getActivity(), mSituation);
+            reportAnalytics(mSituation);
         }
 
         @Override
@@ -1635,6 +1638,20 @@ public class ArrivalsListFragment extends ListFragment
                 return false;
             }
             return true;
+        }
+    }
+
+    private void reportAnalytics(ObaSituation situation) {
+        ObaSituation.AllAffects[] allAffects = situation.getAllAffects();
+        Set<String> agencyIds = new HashSet<>();
+        for (int i = 0; i < allAffects.length; i++) {
+            agencyIds.add(allAffects[i].getAgencyId());
+        }
+
+        for (String agencyId : agencyIds) {
+            ObaAnalytics.reportEventWithCategory(ObaAnalytics.ObaEventCategory.UI_ACTION.toString(),
+                    getString(R.string.analytics_action_service_alerts),
+                    getString(R.string.analytics_label_service_alerts) + agencyId);
         }
     }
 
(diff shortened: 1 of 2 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1638,6 +1639,23 @@ public boolean equals(Object obj) {
         }
     }
 
+    private void reportAnalytics(ObaSituation situation) {
+        ObaSituation.AllAffects[] allAffects = situation.getAllAffects();
+        List<String> agencyIds = new ArrayList<>();
```
barbeau: Could you change `agencyIds` to a `HashSet` instead of a `List`?  I doubt we'll ever have enough agencies in a region for iterating through the list in the `.contains()` method to become problematic performance-wise, but just to be sure... :)

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1638,6 +1639,23 @@ public boolean equals(Object obj) {
...
+        for (int i=0; i< allAffects.length; i++) {
```
barbeau: Minor formatting issue - let's change to:

`for (int i = 0; i < allAffects.length; i++) {`

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1638,6 +1639,23 @@ public boolean equals(Object obj) {
...
+            ObaSituation.AllAffects affect = allAffects[i];
+            if (!agencyIds.contains(affect.getAgencyId())){
+                agencyIds.add(affect.getAgencyId());
+            }
+        }
+
+        for (String agencyId: agencyIds) {
```
barbeau: Minor formatting issue - let's change to:

`for (String agencyId : agencyIds) {`

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1638,6 +1639,23 @@ public boolean equals(Object obj) {
         }
     }
 
+    private void reportAnalytics(ObaSituation situation) {
+        ObaSituation.AllAffects[] allAffects = situation.getAllAffects();
+        List<String> agencyIds = new ArrayList<>();
+        for (int i=0; i< allAffects.length; i++) {
+            ObaSituation.AllAffects affect = allAffects[i];
```
barbeau: I believe you can get rid of the declaration of `ObaSituation.AllAffects affect` here, and just reference `allAffects[i]` directly in the following lines of code?

So it would change to:

~~~
if (!agencyIds.contains(allAffects[i].getAgencyId())) {
    agencyIds.add(allAffects[i].getAgencyId());
}
~~~

Also, minor formatting issue - in the IF block, there should be a space between `) {` - I fixed that in the above example.

barbeau:
@cagryInside Thanks, looks good!  Just a few comments in-line.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
cagryInside: Very good point, my bad!

cagryInside:
@barbeau The latest commit fixes the inline comments.

barbeau:
Thanks @cagryInside!

bbodenmiller:
Same comment as https://github.com/OneBusAway/onebusaway-iphone/pull/626#issuecomment-220645517: I don't think you need to specify the region as I believe you can determine that by cross referencing the data.

barbeau:
We are referencing `agencId` here, not region, so we can break down service alert usage at the agency level (in the case of multiple agencies per region).  You are correct that region can be broken out in GA console.

PR #509, opened 2016-05-23
Fix #508 - Resize RouteShortName if it is more than 3 characters

### LINKED ISSUE #508: RouteShortName overflows if it is more than three character
**Summary:** 

In `ArrivalsListAdapterStyleA`, if the route short name is more than 3 characters it overflows to new line.

**Steps to reproduce:** 

1. Go to Tampa Bay region
2. Search for route `100X`
3. Look at the arrivals

**Expected behavior:** 

It should be in the same line.

Motorola Nexus 6 w/ Android 6.0.1

**Screenshots:** 
device-2016-05-23-142203

cagryInside:
This PR resizes the `routeShortName` if it is longer than 3 characters.

**Methodology**
```
if (routeShortName.length() == 4) {
            view.setTextSize(30);
} else if (routeShortName.length() > 4) {
            view.setTextSize(24);
}
```

**Results**
device-2016-05-23-150119
device-2016-05-23-150357
device-2016-05-23-144612
device-2016-05-23-144526

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
@@ -1,5 +1,6 @@
 package org.onebusaway.android.util.test;
 
+import org.onebusaway.android.R;
 import org.onebusaway.android.app.Application;
 import org.onebusaway.android.io.elements.ObaAgency;
 import org.onebusaway.android.io.elements.ObaArrivalInfo;
@@ -16,6 +17,7 @@ import org.onebusaway.android.ui.ArrivalInfo;
 import org.onebusaway.android.util.UIUtils;
 
 import android.text.TextUtils;
+import android.widget.TextView;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -552,6 +554,25 @@ public class UIUtilTest extends ObaTestCase {
         assertEquals("On time", arrivalInfo.get(29).getStatusText());
         assertEquals("9 min delay", arrivalInfo.get(30).getStatusText());
         assertEquals("On time", arrivalInfo.get(31).getStatusText());
+    }
+
+    public void testMaybeShrinkRouteName() {
+        TextView tv = new TextView(getContext());
+
+        String routeShortName = "TST";
+        float textSize = tv.getTextSize();
+        UIUtils.maybeShrinkRouteName(getContext(), tv, routeShortName);
+        assertEquals(textSize, tv.getTextSize());
+
+        routeShortName = "Test";
+        UIUtils.maybeShrinkRouteName(getContext(), tv, routeShortName);
+        assertEquals(getContext().getResources().getDimension(R.dimen.route_name_text_size_medium),
+                tv.getTextSize());
+
+        routeShortName = "Test2";
+        UIUtils.maybeShrinkRouteName(getContext(), tv, routeShortName);
+        assertEquals(getContext().getResources().getDimension(R.dimen.route_name_text_size_small),
+                tv.getTextSize());
     }
 
     /**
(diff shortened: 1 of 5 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
```diff
@@ -1141,4 +1138,19 @@ public static String formatTime(Context context, long time) {
...
+    /**
+     * Set smaller text size if the route short name has more than 3 characters
+     *
+     * @param view Text view
+     * @param routeShortName Route short name
+     */
+    public static void setTextSizeForArrivalsListAdapterStyleARoute(TextView view,
+                                                                    String routeShortName) {
+        if (routeShortName.length() == 4) {
+            view.setTextSize(30);
```
barbeau: Could you please move the `30` and `24` numbers here to `dimens.xml`?  Check out http://stackoverflow.com/a/16718062/937715 for how to read this correctly.

onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
```diff
@@ -1141,4 +1138,19 @@ public static String formatTime(Context context, long time) {
...
         );
     }
+
+    /**
+     * Set smaller text size if the route short name has more than 3 characters
+     *
+     * @param view Text view
+     * @param routeShortName Route short name
+     */
+    public static void setTextSizeForArrivalsListAdapterStyleARoute(TextView view,
```
barbeau: Let's rename this method to `maybeShrinkLongRouteName(TextView view, String routeShortName)`.  And, let's make the first line:
~~~
if (routeShortName.length() < 4) {
    return;  // No-op if text is short enough to fit
} else if (...
~~~
I think this makes it clearer that this method doesn't always modify the view/text.

barbeau:
@cagryInside Thanks, this looks much better!  A few comments inline.  Could you also please add a brief unit test for this to `UIUtilTest`?

Also, just curious - how many characters does it take to run into two lines with the reduced size?

onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
barbeau: Actually, to avoid confusion with `routeLongName`, let's call the method `maybeShrinkRouteName()`.

cagryInside:
> @cagryInside Thanks, this looks much better! A few comments inline. Could you also please add a brief unit test for this to UIUtilTest?
> 
> Also, just curious - how many characters does it take to run into two lines with the reduced size?

@barbeau It takes 6-7 characters (depends on their sizes) to run into two lines:

device-2016-05-24-102252
device-2016-05-24-102151

cagryInside:
@barbeau the latest commit fixes the inline comments

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
```diff
@@ -554,6 +556,20 @@ public void testArrivalInfoLabels() {
         assertEquals("On time", arrivalInfo.get(31).getStatusText());
     }
 
+    public void testMaybeShrinkRouteName() {
+        TextView tv = new TextView(getContext());
+
+        String routeShortName = "Test";
```
barbeau: Could you also please add one more test here with a route name with 3 characters, to verify that the TextView text size doesn't change?  You could set it to a hard-coded value here before the test just for a simple test.

barbeau:
@cagryInside Thanks, looks good!  Small request to add one more assertion in the unit test for 3 character size to make sure it doesn't get modified.

cagryInside:
> @cagryInside Thanks, looks good! Small request to add one more assertion in the unit test for 3 character size to make sure it doesn't get modified.

Done!

barbeau:
Thanks @cagryInside, looks great!!

PR #522, opened 2016-05-31
Fix #520 - Make navigation drawer scrollable

### LINKED ISSUE #520: Main navigation panel can not scroll in landscape orientation
**Summary:** 

When I change my device on horizontal view the menu look cut and I can't scroll down to see all options on the menu.

**Steps to reproduce:** 

1. Open onebusaway 
2. Change the position of my device to horizontal 
3. Open the principal menu 

**Expected behavior:** 

Scroll down and see all options on the menu 

**Observed behavior:** 

The menu look cut and can not scroll down to see all options 

Xperia Z3 Compact (D5803)
Manufacturer Sony 
Android version Android 6.0.1
RAM (MB) 2048 
Screen size 720 × 1280 
Screen density (dpi) 320 
OpenGL ES version 3.0 
Native platform armeabi-v7a 
CPU make Qualcomm 
CPU model MSM8974AC
Google Play Services 9.0.83
**EDIT** Onebusaway Android version 2.0.11

**Screenshots:** 

screenshot_20160526-201507

screenshot_20160526-201512

cagryInside:
Navigation drawer is now scrollable.

**Screenshots from Nexus 6 -- 5.96" screen**

device-2016-05-31-120549
device-2016-05-31-120643

**Screenshots from HTC EVO 3D -- 4.3" screen**

device-2016-05-31-120922
device-2016-05-31-120935

PS: Added multiple Navigation Drawer entities for test purposes. Also, I removed unused methods.

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
@@ -22,7 +22,7 @@ import org.onebusaway.android.R;
 import org.onebusaway.android.util.UIUtils;
 import org.onebusaway.android.view.ScrimInsetsScrollView;
 
-import android.app.Activity;
+import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.os.Bundle;
@@ -39,6 +39,7 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
+import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
```diff
@@ -289,10 +283,10 @@ private void setSelectedNavDrawerItem(int itemId) {
     }
 
     @Override
```
barbeau: To stay consistent with the `Fragment` implementation, let's leave this as:

`public void onAttach(Activity activity) {`

barbeau:
Thanks @cagryInside!  One small change - otherwise looks good!

onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
cagryInside: @barbeau `onAttach(Activity activity)` is deprecated. Thats why I used `onAttach` with `Context`. This is the reference https://developer.android.com/reference/android/app/Fragment.html#onAttach(android.app.Activity). Do we still want to use `onAttach(Activity activity)`?

barbeau: Ah, good catch.  No, you're right, we should change to `onAttach(Context context)`.  So this PR should be good as-is.  Thanks!

barbeau:
Thanks @cagryInside!  Per my other comment this looks good as-is!

PR #524, opened 2016-06-03
Fix #523 - Immediately populate stop name in arrivals header from values in Intent

### LINKED ISSUE #523: Stop name in arrivals header sliding panel is incorrect on initial load
**Summary:** 

Reported on oba-dev list - https://groups.google.com/forum/#!topic/onebusaway-developers/bq8y7m6dykE

If the user is viewing arrivals in ArrivalsListActivity and taps on "View on Map" option in the Action Bar, the stop name is not  populated immediately, showing "University Area Transit Center" layout placeholder name instead of the actual stop name.

**Steps to reproduce:** 

From above post:

>1) I displayed next departures from a favourite stop (text only view)
2) I clicked on map icon

My steps:

1. Go to Starred Stops
2. Tap on a starred stop
3. Tap on map icon in Action Bar

You can also reproduce this by:
* Rotating the orientation of the device in the map view when the arrivals header is open
* Minimizing the app when the sliding panel header is visible, browsing other apps until OBA gets suspended in the background, and then returning to OBA so Android re-loads it

**Expected behavior:** 

Stop location for the selected stop should be focused on, and the stop name for this stop should be shown in the sliding panel header in HomeActivity.

**Observed behavior:** 

>3) I correct location on map showed up, but instead of correct name of stop and list of next departures, "University Area Transit Center" showed up with no departures

This is more noticeable in slower internet connections when the stop name doesn't populate quickly from the server response.

Unknown

**Screenshots:** 

image

barbeau:
@cagryInside Could you please take a look at this to make sure I'm not missing any case when the stop name should be immediately populated (e.g., orientation changes, etc)?  See #523 for more details, including how to reproduce.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -96,6 +96,8 @@ public class ArrivalsListFragment extends ListFragment
 
     public static final String STOP_NAME = ".StopName";
 
+    public static final String STOP_CODE = ".StopCode";
+
     public static final String STOP_DIRECTION = ".StopDir";
 
     /**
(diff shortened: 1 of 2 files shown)


cagryInside:
> @cagryInside Could you please take a look at this to make sure I'm not missing any case when the stop name should be immediately populated (e.g., orientation changes, etc)? See #523 for more details, including how to reproduce.

Tested this one and it seems ok. I put a breaking point when it was doing a network request, then I was able to clearly see the result.

barbeau:
Thanks @cagryInside!

PR #527, opened 2016-06-07
WIP - Fix #525 - Support User Location Via ZIP Code

### LINKED ISSUE #525: Support user location via manual zip code entry
From user feedback on Google Play:

>Really like it, except... I do not turn on the location on my phone. It eats my battery and I don't like the idea of my phone advertising my location all the time. Without location on, this app seems to think I'm in the middle of Africa. It also seems to glitch out and lose my spot every time I open it and while I'm actively using it. How come this app's settings don't allow me to enter my zip code? I think that would be a simple and common sense solution for users.

This is a completely fair request, especially from a privacy perspective.  Shouldn't be too hard to prompt the user for a zip code if they choose not to turn location on, and then use the Google Geocoder to get the location from the zip code.

axismb:
Opening a PR for Issue #525, still a WIP but opening for the sake of code review & advice if that is okay. I will squash and re-upload when complete.

### DIFF
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -1106,8 +1106,7 @@ public class BaseMapFragment extends SupportMapFragment
         private Dialog createZipCodeDialog() {
             Drawable icon = getResources().getDrawable(android.R.drawable.ic_dialog_map);
             DrawableCompat.setTint(icon, getResources().getColor(R.color.theme_primary));
-            final EditText zipInput = new EditText(this.getContext());
-            zipInput.setInputType(InputType.TYPE_CLASS_NUMBER);
+            final EditText zipInput = (EditText)this.getActivity().findViewById(R.id.zipText);
 
             final AlertDialog.Builder builder = new AlertDialog.Builder(getActivity())
                     .setTitle(R.string.main_ziplocation_title)
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -27,7 +27,31 @@
  */
 package org.onebusaway.android.map.googlemapsv2;
 
+<<<<<<< 8fb7ada7860d56925ded95ad73321303969d7d55
 import com.google.android.gms.common.api.GoogleApiClient;
+=======
+import android.app.Activity;
+import android.app.Dialog;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.graphics.drawable.Drawable;
+import android.location.Location;
+import android.os.Bundle;
+import android.os.Handler;
+import android.provider.Settings;
+import android.support.v4.graphics.drawable.DrawableCompat;
+import android.support.v7.app.AlertDialog;
+import android.text.InputType;
+import android.text.TextUtils;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.EditText;
+import android.widget.Toast;
+
+>>>>>>> Fix #525 - Support User Location Via ZIP Code
 import com.amazon.geo.mapsv2.CameraUpdateFactory;
 import com.amazon.geo.mapsv2.LocationSource;
 import com.amazon.geo.mapsv2.OnMapReadyCallback;
@@ -58,25 +82,6 @@ import org.onebusaway.android.util.LocationHelper;
 import org.onebusaway.android.util.LocationUtils;
 import org.onebusaway.android.util.PreferenceUtils;
 import org.onebusaway.android.util.UIUtils;
-
-import android.app.Activity;
-import android.app.Dialog;
-import android.content.Context;
-import android.content.DialogInterface;
-import android.content.Intent;
-import android.graphics.drawable.Drawable;
-import android.location.Location;
-import android.os.Bundle;
-import android.os.Handler;
-import android.provider.Settings;
-import android.support.v4.graphics.drawable.DrawableCompat;
-import android.support.v7.app.AlertDialog;
-import android.text.TextUtils;
-import android.util.Log;
-import android.view.LayoutInflater;
-import android.view.View;
-import android.view.ViewGroup;
-import android.widget.Toast;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -378,6 +383,25 @@ public class BaseMapFragment extends SupportMapFragment
         MapDialogFragment.newInstance(id, this).show(getFragmentManager(), MapDialogFragment.TAG);
     }
 
+    private void setLocationByZip(final String zip) {
+        final Handler h = new Handler();
+        h.post(new Runnable() {
+            @Override
+            public void run() {
+                Location l = LocationUtils.getLocationFromZip(
+                        Application.get().getApplicationContext(), zip);
+                if (l != null && mLocationHelper != null) {
+                    mLocationHelper.onLocationChanged(l);
+                    setMyLocation(l, true, true);
+                } else {
+                    Toast.makeText(Application.get().getApplicationContext(),
+                            getString(R.string.main_ziplocation_error),
+                            Toast.LENGTH_SHORT).show();
+                }
+            }
+        });
+    }
+
 //    @Override
 //    public void onActivityResult(int requestCode, int resultCode, Intent data) {
 //        super.onActivityResult(requestCode, resultCode, data);
@@ -592,6 +616,7 @@ public class BaseMapFragment extends SupportMapFragment
      */
     @Override
     @SuppressWarnings("deprecation")
+<<<<<<< 8fb7ada7860d56925ded95ad73321303969d7d55
     public boolean setMyLocation(boolean useDefaultZoom, boolean animateToLocation) {
         if (!LocationUtils.isLocationEnabled(getActivity()) && mRunning && UIUtils.canManageDialog(
                 getActivity())) {
@@ -605,6 +630,23 @@ public class BaseMapFragment extends SupportMapFragment
         }
 
         Location lastLocation = Application.getLastKnownLocation(getActivity(), apiClient);
+=======
+    public void setMyLocation(boolean useDefaultZoom, boolean animateToLocation) {
+        Location lastLocation = Application.getLastKnownLocation(this.getActivity(), null);
+        if (!LocationUtils.isLocationEnabled(getActivity()) && mRunning && UIUtils.canManageDialog(
+                getActivity())) {
+
+            String zipCodeKey = getString(R.string.preference_key_zip_code);
+            String zip = PreferenceUtils.getString(zipCodeKey);
+            if (zip != null) {
+                setLocationByZip(zip);
+                return;
+            } else {
+                showDialog(MapDialogFragment.NOLOCATION_DIALOG);
+            }
+        }
+
+>>>>>>> Fix #525 - Support User Location Via ZIP Code
         if (lastLocation == null) {
             Toast.makeText(getActivity(),
                     getResources()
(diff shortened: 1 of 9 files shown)


barbeau:
Sure, WIPs are great!  Will try to look at this tomorrow.

azizmb9494:
Thanks @barbeau!

barbeau:
For UI, I think we should work this into the "Location is turned off" prompt.  If they decline to turn on location, then prompt them to enter zip.

On a related note, there might be issues with caching map location via bundle extras - I don't recall testing this in a while and user reported that map location is reset when restarting.  If location isn't available it should start map at last known location.

azizmb9494:
Re: The Map, I checked earlier today and iirc, that's exactly what it does. If last known location isn't null, resets to it.

My thinking for going about this was on the HomeActivity, check if location services is enabled/available, if not, check if the user has a zip code stored from previous uses. If not, prompt them to enter it. Once entered, attempt to retrieve location object and call LocationHelper.onLocationChanged on it.

azizmb9494:
Just saw that an isLocationEnabled function already exists, reverting now.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -1030,17 +1030,21 @@ private Dialog createZipCodeDialog() {
...
                                 public void onClick(DialogInterface dialog, int which) {
-                                    String zip = zipInput.getText().toString();
-                                    if (zip != null) {
+                                    String zip = zipInput.getText().toString().trim();
+                                    if (zip != null && LocationUtils.isValidZipCode(zip)) {
                                         String zipCodeKey = getString(R.string.preference_key_zip_code);
                                         PreferenceUtils.saveString(zipCodeKey, zip);
+                                        Location l = LocationUtils.getLocationFromZip(getActivity(), zip);
+                                        if (l != null) {
+                                            Application.setLastKnownLocation(l);
```
azizmb9494: @barbeau: Wouldn't I need to call onLocationChanged somewhere for this to fire off a refresh?

barbeau: Hmmm, good question.  Location has always come from an Android `LocationListener` in the past, so we don't have precedence for a location generated by ourselves.

There are a few ways to do this, including injecting the location into `mLocationHelper.onLocationChanged(l)` (which will in turn call `Application.setLastKnownLocation(l)`), or just recentering the map using by setting location `Application.setLastKnownLocation(l)` and then `BaseMapFragment.setMyLocation(true, true`)`.

Let's try injecting the location into the `LocationHelper` in `BaseMapFragment` using `mLocationHelper.onLocationChanged(l);`.  It might just work.

azizmb9494:
@barbeau: If there's no issues with this, I will squash and push! Currently, when you tap location button on map, it shows the usual "Location Services disabled" prompt, if you hit cancel, it'll prompt you to enter your zip code (or cancel again). Once you do that, it'll take you to your approximate location, and on future clicks, it'll automatically do that. In the future, the user can change the ZIP code by going to preferences.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -346,6 +349,23 @@ protected void showDialog(int id) {
...
+    private void setLocationByZip(final String zip) {
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                Location l = LocationUtils.getLocationFromZip(getContext(), zip);
+                if (l != null) {
+                    mLocationHelper.onLocationChanged(l);
+                    setMyLocation(l, true, true);
+                } else {
+                    Toast.makeText(getContext(), getString(R.string.main_ziplocation_error),
```
barbeau: `getContext()` should be `getApplicationContext()`, as the Toast could trigger when this Activity is destroyed.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -346,6 +349,23 @@ protected void showDialog(int id) {
         MapDialogFragment.newInstance(id, this).show(getFragmentManager(), MapDialogFragment.TAG);
     }
 
+    private void setLocationByZip(final String zip) {
+        Runnable r = new Runnable() {
```
barbeau: This should be posted to an Android `Handler` instead of just running a Thread - for example, something like:

~~~
final Handler h = new Handler();
h.post(new Runnable() {...
~~~

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -1009,6 +1036,42 @@ public void onClick(DialogInterface dialog, int which) {
...
+            DrawableCompat.setTint(icon, getResources().getColor(R.color.theme_primary));
+            final EditText zipInput = new EditText(this.getContext());
+            zipInput.setInputType(InputType.TYPE_CLASS_NUMBER);
+
+            final AlertDialog.Builder builder = new AlertDialog.Builder(getActivity())
+                    .setTitle(R.string.main_ziplocation_title)
+                    .setIcon(icon)
+                    .setCancelable(false)
+                    .setMessage(R.string.main_ziplocation)
+                    .setView(zipInput)
```
barbeau: This needs margins on the sides of the EditText, so it aligns with the rest of the dialog content and meets the material design guidelines:

image

You may be able to do this programmatically, but it's probably easier just to use a layout.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -998,6 +1025,42 @@ public void onClick(DialogInterface dialog, int which) {
...
                                 @Override
                                 public void onClick(DialogInterface dialog, int which) {
+                                    createZipCodeDialog().show();
+                                }
+                            }
+                    );
+            return builder.create();
+        }
+
+        private Dialog createZipCodeDialog() {
```
barbeau: Let's move this to `UIUtils`, and have it return the `Dialog` so we just show it here - it may be useful elsewhere, and `BaseMapFragment` is already pretty huge.

onebusaway-android/src/main/java/org/onebusaway/android/util/LocationUtils.java
```diff
@@ -214,6 +219,37 @@ public static GoogleApiClient getGoogleApiClientWithCallbacks(Context context) {
     }
 
     /**
```
barbeau: Could you add the parameters and return values in the Javadocs here?  Easy way to do this is copy the current text to clipboard, remove the comment, and above the method in Android Studio type `/**` and then hit enter.  It should automatically generate the @param and @return template.

onebusaway-android/src/main/java/org/onebusaway/android/util/LocationUtils.java
```diff
@@ -214,6 +219,37 @@ public static GoogleApiClient getGoogleApiClientWithCallbacks(Context context) {
...
+     * Checks if string is valid zip code.
+     */
+    public static boolean isValidZipCode(String zip)
+    {
+        Pattern pattern = Pattern.compile("^\\d{5}(?:-\\d{4})?$");
+        return pattern.matcher(zip).matches();
+    }
+
+    /**
```
barbeau: For this Javadoc, please add that this method makes a call to a server (in platform in `geocoder.getFromLocationName()`), and could be a long running operation, and therefore shouldn't be done on the UI thread.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
barbeau: You can actually use the ZIP code preference dialog as guidance for what the padding should be:

image

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/LocationUtilsTest.java
```diff
@@ -201,6 +201,16 @@ public void testLocationServices() {
         }
     }
 
+    public void testZipCodes() {
```
barbeau: Thanks for the test! :+1:

onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
```diff
@@ -112,6 +114,10 @@ public void onCreate(Bundle savedInstanceState) {
         aboutPref = findPreference(getString(R.string.preferences_key_about));
         aboutPref.setOnPreferenceClickListener(this);
 
+        zipPref = findPreference(getString(R.string.preference_key_zip_code));
```
barbeau: You only need this code if we want to track taps on this preference in analytics via  `onPreferenceClick()` - which I think we do.  Could you add an IF statement and analytics labels for this?  You can use the `aboutPref` as a template.

barbeau:
@azizmb9494 Thanks for this!  I got to do an initial pass at a review, and have some comments inline.  I need to look at some of the location logic a little more closely - it's complicated with various callbacks, and getting/injecting location from an external source is new, and I want to make sure nothing strange is happening.  I also need to think about the text for the dialog/preference a bit.  So I'll need a little more time to review, but in the mean time feel to address the comments I already made.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -357,6 +360,23 @@ protected void showDialog(int id) {
         MapDialogFragment.newInstance(id, this).show(getFragmentManager(), MapDialogFragment.TAG);
     }
 
+    private void setLocationByZip(final String zip) {
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                Location l = LocationUtils.getLocationFromZip(getContext(), zip);
+                if (l != null) {
+                    mLocationHelper.onLocationChanged(l);
```
barbeau: We may need to check if `mLocationHelper == null` here as well - `LocationUtils.getLocationFromZip()` does a network call and could potentially be a long-running method, and the user may have backgrounded or killer the Fragment/Activity in the mean time.  Please do some testing where you enter the zip and then immediately kill the app to make sure it doesn't crash.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -357,6 +360,23 @@ protected void showDialog(int id) {
         MapDialogFragment.newInstance(id, this).show(getFragmentManager(), MapDialogFragment.TAG);
     }
 
+    private void setLocationByZip(final String zip) {
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                Location l = LocationUtils.getLocationFromZip(getContext(), zip);
```
barbeau: This should probably be `getApplicationContext()` as well.

azizmb9494:
Thanks @barbeau! Addressed all but 1 of them, will do that one of the next week (hopefully this weekend!).

barbeau:
As I mentioned in #551, I think the following is the behavior we want for this feature:

Initial setup:
1. When the user taps on the "My Location" button, if Location Services are disabled in system settings ask them if they want to enable them
1. If they choose no, then ask if they want to enter a zip code to be used as their location when Location Services are disabled.  If yes, prompt them to enter zip code.  If no, then do 1) above again if the user taps on the "My Location" button again.

Ongoing behavior:
1. If the user taps the My Location button, if Location Services are disabled then recenter on the zip code they entered.  Show the zip code as the user's location.  I think we should also set the accuracy of the location we create to a very large value, something like 1000 meters, to indicate that it's not very accurate.
1. If the user taps the My Location button, if Location Services are enabled then use the user's actual real location, as if this feature isn't implemented.
1. Allow the user to clear the zip code from Settings

azizmb9494:
I'll have to double check but I think that's how it works right now.

barbeau:
@azizmb9494 Just trying to catch up on open PRs - is this ready for review, or did you still need to clean up a few items first?

azizmb9494:
@barbeau: Still need to clean up, have to re-factor dialog creation code & fix XML for padded editText for ZIP Entry. Sorry for the delay.

barbeau:
No worries, just wanted to make sure I wasn't holding things up.

aaronbrethorst:
Closing due to age. Please feel free to reopen with merge conflicts fixed.

PR #528, opened 2016-06-07
Fix #478 - Force update oba regions on restore from storage

### LINKED ISSUE #478: Restoring an old database version results in empty Open311 tables
**Summary:** 

If you get a backup of DB with the lower version of the application, and restore this DB after you upgrade the application, it will effectively delete the data in new created tables.

**Steps to reproduce:** 

1. Download the lower version of the app. 
1. Go to settings and click `Save to storage`. 
1. Remove the app. 
1. Download the latest version of the app. 
1. Go to settings and click `Restore from storage`. 

The latest version of the app has `Open311 url` data in the tables. After you do the above operations, the Open311 servers table will be empty.

***Shortcut steps to reproduce:*** 
Download the latest version of the app. Download [db.backup.zip](https://github.com/OneBusAway/onebusaway-android/files/247113/db.backup.zip) (taken from an older version of OBA) and unzip it. Put it under `\Internal storage\OBABackups`. Go to Settings and click `Restore from storage`. 

**Expected behavior:** 

It should update the data in the old tables. If there is new table, it repopulate it if possible.

**Observed behavior:** 

When you restore an older version of DB, the Open311 servers table will be empty. 

Stock Nexus 6 w/ Android 6.0.1 I don't believe this is device or version specific bug. It can occur with any device or version.

**Screenshots:** 

Before DB dump It has open 311 enpoint:
device-2016-05-03-105933

After DB dump, the open311 endpoints are no longer in the DB:
device-2016-05-03-110110

cagryInside:
We are calling `ObaRegionsTask` with `force update` option when restoring a database from the storage.

**Test Case**
1) Put it under the new [db.backup.zip](https://github.com/OneBusAway/onebusaway-android/files/303162/db.backup.zip) under `\Internal storage\OBABackups` which has different region name:

device-2016-06-07-130034

2) Click `restore from storage` :

device-2016-06-07-130642

3) Make sure `ObaRegionsTask` was executed successfully by looking the region names:

device-2016-06-07-133044

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/io/backup/RestorePreference.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/io/backup/RestorePreference.java
@@ -16,7 +16,9 @@
 package org.onebusaway.android.io.backup;
 
 import org.onebusaway.android.R;
+import org.onebusaway.android.app.Application;
 import org.onebusaway.android.io.ObaAnalytics;
+import org.onebusaway.android.region.ObaRegionsTask;
 
 import android.app.AlertDialog;
 import android.content.Context;
@@ -56,7 +58,7 @@ public class RestorePreference extends Preference {
         final String state = Environment.getExternalStorageState();
         // Also, this is only enabled if there's a backup file
         return (Environment.MEDIA_MOUNTED_READ_ONLY.equals(state) ||
-                        Environment.MEDIA_MOUNTED.equals(state)) &&
+                Environment.MEDIA_MOUNTED.equals(state)) &&
                 Backup.isRestoreAvailable(getContext());
     }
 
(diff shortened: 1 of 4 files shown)


onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -522,6 +522,7 @@
     <string name="preferences_db_save_error">Unable to save: %1$s</string>
     <string name="preferences_db_restore_warning">This will wipe out any existing data! Go ahead?
     </string>
+    <string name="preferences_restore_loading">Restoring from storage</string>
```
barbeau: Could you please add an ellipsis here?

So:
`<string name="preferences_restore_loading">Restoring from storage…</string>`

Also, please make sure it's the unicode character for the ellipsis, not just three periods.

Thanks!

barbeau:
Thanks @cagryInside!  Looks good - one small comment inline.

cagryInside:
@barbeau latest commit fixes the inline comment

onebusaway-android/src/main/java/org/onebusaway/android/io/backup/RestorePreference.java
```diff
@@ -86,15 +87,23 @@ public void onClick(DialogInterface dialog, int which) {
...
         try {
             Backup.restore(context);
-            Toast.makeText(context,
-                    context.getString(R.string.preferences_db_restored),
-                    Toast.LENGTH_LONG).show();
+
+            ObaRegionsTask task = new ObaRegionsTask(context, new ObaRegionsTask.Callback() {
+                @Override
+                public void onRegionTaskFinished(boolean currentRegionChanged) {
+                    Toast.makeText(context,
```
barbeau: In the Toast example [here](https://developer.android.com/guide/topics/ui/notifiers/toasts.html), they use `getApplicationContext()` - I think that's a better option here as the Activity context could have been destroyed.

Also, instead of `context.getString(R.string.preferences_db_restored)`, I believe we can just pass in the resource ID:
`R.string.preferences_db_restored`

Sorry, I missed this in the first review.

barbeau:
Thanks!  Sorry, one last item inline.

cagryInside:
It should be ok now.

barbeau:
Thanks @cagryInside!

PR #540, opened 2016-06-14
WIP Fix #25 - OTP trip planning feature

sdjacobs:
Trip planning feature using OpenTripPlanner. See issue #25.

FYI, I rebased to master - not sure if that means 25-tripPlan should be updated first, or this should be merged to master.

TODO
- UI improvements
- define realtime feature a little better

PR #541, opened 2016-06-14
WIP - Fix #25 - OpenTripPlanner-based trip planning feature

### LINKED ISSUE #25: Trip planning
There's no longer an OBA API for it, but people continue to ask for it. Frankly I doubt I will ever have time to do this, but I want to keep something open because it's such a popular feature request.

To be clear, this would involve either bringing back Brian's API, integrating OpenTripPlanner, or running my own server. If it were a different server it would have to be kept updated with GTFS updates just like the main OBA server is.

Moved from: http://code.google.com/p/seattle-bus-bot/issues/detail?id=83

sdjacobs:
- Code was integrated from York OBA app and OpenTripPlanner-for-Android.
- New TripPlanActivity with TripPlanFragment and TripResultsFragment.
- If a region has an OpenTripPlanner instance specified in otpBaseUrl field, that instance will be queried from the trip planner.
- Realtime feature: after a trip is planned and an itinerary is selected, OTP will be requeried every minute. If the rank of the selected itinerary changes in a returned list of itineraries, or the selected itinerary is no longer present, the user will be notified and can replan the trip.

TODO
- UI improvements
- realtime feature changes?

Replaces PR #540


barbeau:
Thanks @sdjacobs, this is great!

So based on our discussion off-line for the real-time reminder feature:

>you could also get notified if your trip became the best trip – if you pick the trip ranked 2 and then the start time for trip rank 1 passes, on the next request your trip can become the top ranked. Also, we’ve seen some behavior where the order of trips that OTP returns isn’t what we would consider best-to-worst. 

So the way we implemented the real-time reminder feature in OTP Android was to alert the user if the real-time prediction for the trip plan they chose changed.  So, if it was running on-time, and then it became delayed by 3 minutes, we would alert them with the update, and if the user tapped on the notification we'd take them to the trip planning screen where they could re-plan the trip if they wanted.  The user could then see if there was a better trip plan available.

Given the above comment, I think the OTP Android implementation based on predictions changing for a single trip_id is a less prone to false negatives, because it pushes the responsibility of figuring out if there is a "better trip" back on the user (although that's probably the main downside of this implementation as well).  But, as you say above, automatically determining when a trip is "better" probably isn't simple.

Were there other drawbacks to this implementation that influenced your choice to change the reminder feature to be based on the trip_ids in the itinerary list instead of delays for a particular trip?

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -696,9 +699,10 @@ public float getZoomLevelAsFloat() {
     }
 
     @Override
-    public void setRouteOverlay(int lineOverlayColor, ObaShape[] shapes) {
+    public void setRouteOverlay(int lineOverlayColor, ObaShape[] shapes, boolean clear) {
         if (mMap != null) {
-            mLineOverlay.clear();
+            if (clear)
```
barbeau: Please always use brackets `{` for IF statements, even if it's just a one-liner.

barbeau:
`src/main/res/drawable-hdpi/icon.png`, and any similar boilerplate icons, should be removed.

barbeau:
All PNGs should be added as black icons with a transparent background, and then colored dynamically, like:

`iconView.setColorFilter(
                        getResources().getColor(R.color.navdrawer_icon_tint_selected));`

The above example is from the [NavigationDrawerFragment](https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java#L450), and shows how the black trip plan icon is colored.

This is required for the build variants based on different theme colors to work.

The easiest way to generate these icons is to use the Android clipart icon generator, where you can choose the material design icon from clipart and output new black icons for all the screen densities, then just drop the folder into OBA Android:
https://romannurik.github.io/AndroidAssetStudio/icons-generic.html#source.type=clipart&source.space.trim=0&source.space.pad=0&size=24&padding=0&color=000%2C100&name=ic_example

When generating icon, use a padding of `0%` and `0dip` and start with a size of `24dip`.

sdjacobs:
The OTP Android implementation determines whether a trip is running on-time or not based on the "/trips/{tripId}/stoptimes" call in OTP. However this call returns scheduled stop times, not realtime stop times. @vreixo made a PR at opentripplanner/OpenTripPlanner#1495 to change this behavior, but it was unfortunately closed (I believe mistakenly).

OTP also has a call "/stops/{stopId}/stoptimes" which does seem to use realtime info (based on the code, I haven't tested it). But in any case, OTP's IndexAPI is pretty brittle - it seems to change quite a bit between versions and I'd rather not be tied to one specific version of OTP. 

Another option could be to get the stop times from OBA instead. Do you think that's reasonable? Presumably we could leverage another part of onebusaway-android to do that.

barbeau:
>But in any case, OTP's IndexAPI is pretty brittle - it seems to change quite a bit between versions and I'd rather not be tied to one specific version of OTP.

Yeah, not having a v1.0 release in OTP that freezes the API is one of my major complaints about OTP.  Just posted here asking for v1.0 release timeline - https://groups.google.com/forum/#!topic/opentripplanner-dev/EaVlEnSpcKY.

>Another option could be to get the stop times from OBA instead. Do you think that's reasonable? 

The per-trip/per-vehicle schedule deviation could be retrieved for a specific trip via [trip-details API](http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/methods/trip-details.html) - this is the `TripDetailsRequest`.

I'm not sure that cross OBA/OTP queries would be guaranteed to work going forward, as I understand that OTP has changed the way it identifies transit elements along with the agency.  The version you're currently using make still use the OBA format of `agency_id`_`trip_id`, as earlier versions used the OBA GTFS library.  But I think this was changed in newer OTP versions - I recall seeing posts that Conveyal was moving away from OBA GTFS library to their own.  So I'm not sure for forwards-compatibility reasons with OTP this option is much better than using the OTP API.

sdjacobs:
> I'm not sure that cross OBA/OTP queries would be guaranteed to work going forward, as I understand that OTP has changed the way it identifies transit elements along with the agency.

OTP is actually still using OBA's GTFS library, though the interpretation of AgencyAndId has changed - the agencyId String is now a "feed ID", an ID for the GTFS bundle which the entity was read from. However, the Leg object in an Itinerary has its own agencyId field (for transit Legs) which does match the gtfs and OBA agency. So Leg.agencyId should allow us to identify transit elements such that they're identifiable in OBA too.

barbeau:
>So Leg.agencyId should allow us to identify transit elements such that they're identifiable in OBA too.

Does OTP `trip_id` still follow the same structure as OBA, like "1_1234", or "Hillsborough Area Regional Transit_1234" if `agency_id` in GTFS isn't defined?

If so, then we could try using OBA `trip-details` API.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/ConversionUtils.java
```diff
@@ -0,0 +1,464 @@
...
+        } else {
+            oldTimeString = new SpannableString(" ");
+        }
+        timeUpdatedString = TextUtils.concat(beforeDateString, newDateString, oldDateString,
+                beforeTimeString, oldTimeString, newTimeString, timezone);
+        return timeUpdatedString;
+    }
+
+
+    public static int getDelayColor(long delay, Context applicationContext) {
```
barbeau: This method can be removed, and `ArrivalInfo.computeColorFromDeviation()` can be used instead.  Just be aware that that method returns the resource ID, which would still need to be converted to a color via `getResources().getColor(R.color.xxx)` (the below method returns the color integer directly).

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/ConversionUtils.java
```diff
@@ -0,0 +1,464 @@
+/*
```
barbeau: I'd like to see unit tests added for any utility methods.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/ConversionUtils.java
```diff
@@ -0,0 +1,464 @@
...
+
+/**
+ * @author Khoa Tran
+ * @author Simon Jacobs - conversion for OBA. Added Imperial measurements
+ */
+
+public class ConversionUtils {
+
+    public static double getDuration(String startTimeText, String endTimeText,
+                                     Context applicationContext) {
```
barbeau: Could you please add better Javadocs to these methods?  Should at a minimum include what the method does, along with `@param` and `@return` for parameters and return variables, respectively.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+                }
+            }
+        }
+
+        final boolean unitsAreImperial = !RegionUtils.getUnitsAreMetric(getContext());
+
+        Double maxWalk = builder.getMaxWalkDistance();
+        if (maxWalk != null && Double.MAX_VALUE != maxWalk) {
+            if (unitsAreImperial)
+                maxWalk *= 3.281;
```
barbeau: Can you please refactor this magic number out to a `final static` variable with a name for what is it?  Even better would be moving the conversion to it's own utility method in the `ConversionUtils` class, with a method signature like:

`public static double metersToFeet(double meters) {`

onebusaway-android/src/main/java/org/onebusaway/android/util/RegionUtils.java
```diff
@@ -277,6 +279,51 @@ public static boolean isRegionUsable(ObaRegion region) {
...
+
+        // Trim trailing slash if it exists.
+        return otpBaseUrl.replaceFirst("/$", "");
+    }
+
+    /**
+     *
+     * @return returns true if preferred units are metric, false if Imperial
+     */
+    public static boolean getUnitsAreMetric(Context context) {
```
barbeau: Could you please move this to `PreferenceUtils`?  It really doesn't have anything to do with the region, as it's a user preference.  It should have a name like `getUnitsAreMetricFromPreferences()`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
```diff
@@ -89,7 +91,8 @@
             R.string.navdrawer_item_my_reminders,
             R.string.navdrawer_item_settings,
             R.string.navdrawer_item_help,
-            R.string.navdrawer_item_send_feedback
+            R.string.navdrawer_item_send_feedback,
+            R.string.navdrawer_item_plan_trip
```
barbeau: We should change to dynamically generating this array, and only add the "Plan a Trip" navigation drawer element for regions where `region.getOtpBaseUrl()` isn't null.

onebusaway-android/src/main/java/org/onebusaway/android/util/RegionUtils.java
```diff
@@ -277,6 +279,51 @@ public static boolean isRegionUsable(ObaRegion region) {
     }
 
     /**
+     *
+     * @return OTP server URL with trimmed trailing slash, or null if none defined for region.
+     */
+    public static String getOtpBaseUrl() {
```
barbeau: Could you please change the method name and signature to:

`public static String formatOtpBaseUrl(String baseUrl) {`

...and don't reference `Application.get()` internally?

In other words, the utility method should do an operation on the string.  I've been trying to keep the Utils classes/methods free from internal statefulness.  This also makes it simpler to write unit tests for various OTP URL formats.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -735,4 +736,247 @@
...
+
+    <!-- Obtaining real time updates -->
+    <string name="toast_realtime_updates_fail">Realtime updates not available for this trip, times might be outdated</string>
+
+    <!-- Units -->
+    <string name="time_short_hours">h</string>
+    <string name="time_short_minutes">m</string>
+    <string name="time_short_seconds">s</string>
+    <string name="time_long_hours">h</string>
+    <string name="time_long_minutes">min</string>
```
barbeau: For time units, we already have:

    <string name="minutes_abbreviation">min</string>
    <string name="minutes_full">minutes</string>

Unless there is a reason not to, let's use these to be consistent throughout OBA.

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -735,4 +736,247 @@
...
+    <string name="tripplanner_error_geocode_to_ambiguous">Destination  is uncertain</string>
+    <string name="tripplanner_error_geocode_from_to_ambiguous">Origin and destination are uncertain</string>
+    <string name="tripplanner_error_triangle">Something went wrong.  Change your bike options and try again</string>
+    <!-- When routing from my location -->
+    <string name="tripplanner_current_location">Current Location</string>
+    <string name="tripplanner_error_no_location">Please enter a location for the start and end of your trip.</string>
+
+    <string name="tripplanner_reverse">Reverse</string>
+
+    <!-- Units -->
```
barbeau: For distance units, we already have:

    <string name="feet_abbreviation">ft</string>
    <string name="miles_abbreviation">mi</string>
    <string name="meters_abbreviation">m</string>
    <string name="kilometers_abbreviation">km</string>

Unless there is a reason not to, let's use these to be consistent throughout OBA.

onebusaway-android/src/main/res/layout/fragment_trip_plan_results.xml
```diff
@@ -0,0 +1,273 @@
...
+                android:id="@+id/textView2"
+                android:gravity="center_vertical"
+                android:layout_marginLeft="10dp"/>
+
+        <ImageView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:id="@+id/switchViewImageView"
+                android:layout_gravity="center_horizontal"
+                android:src="@drawable/ic_action_map"/>
```
barbeau: We already have `@drawable/ic_arrivals_styleb_action_map`, which is the same icon (but black and dynamically colored) - you should be able to use this existing icon rather than including another.

onebusaway-android/build.gradle
```diff
@@ -314,4 +321,7 @@ dependencies {
     googleCompile 'com.google.android.gms:play-services-maps:9.0.1'
     // Amazon Maps (only for Amazon flavor)
     amazonCompile 'com.amazon.android:amazon-maps-api:2.0'
+    // POJOs used for full data-binding via Jackson
```
barbeau: Can you please move this dependency above the `googleCompile`? line?  I've been trying to keep the dependencies used for all build variants above the ones used for only specific build variants.

onebusaway-android/src/main/java/org/onebusaway/android/provider/ObaProvider.java
```diff
@@ -225,6 +225,9 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                                 ObaContract.RegionOpen311Servers.API_KEY + " VARCHAR NOT NULL, " +
                                 ObaContract.RegionOpen311Servers.BASE_URL + " VARCHAR NOT NULL " +
                                 ");");
+                db.execSQL(
```
barbeau: This `ALTER TABLE`  statement needs to be moved to its own:

`if (oldVersion == 23) {`

...IF block, and a `++oldVersion;` needs to be added to the end of the `if (oldVersion == 22) {` IF block.

Otherwise the database won't upgrade properly.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -425,6 +426,11 @@ private void goToNavDrawerItem(int item) {
                             getString(R.string.analytics_label_button_press_reminders));
                 }
                 break;
+            case NAVDRAWER_ITEM_PLAN_TRIP:
+                Intent planTrip = new Intent(HomeActivity.this, TripPlanActivity.class);
+                startActivity(planTrip);
+                // TODO - Analytics
```
barbeau: Still need to add analytics here.  Would be nice to add them elsewhere in the trip planning portion as well, although I'll need to look at that more carefully.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanActivity.java
```diff
@@ -0,0 +1,122 @@
...
+            builder.copyIntoBundle(bundle);
+    }
+
+    public void route() {
+        builder.setListener(this).execute(this);
+    }
+
+    @Override
+    public void onTripRequestComplete(List<Itinerary> itineraries) {
+        Log.i(TAG, "Successfully routed. Itineraries are " + itineraries);
```
barbeau: All `Log` statements should be `Log.d()`, unless they are in a try/catch exception block, in which case they should be `Log.e()`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+
+                boolean wheelchair = wheelchairCheckbox.isChecked();
+
+                String maxWalkString = maxWalkEditText.getText().toString();
+                double maxWalkDistance;
+                if (TextUtils.isEmpty(maxWalkString)) {
+                    maxWalkDistance = Double.MAX_VALUE;
+                } else {
+                    double d = Double.parseDouble(maxWalkString);
+                    maxWalkDistance = unitsAreImperial ? d / 3.281 : d;
```
barbeau: Same - magic number.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+        address.setAddressLine(0, getString(R.string.tripplanner_current_location));
+        address.setLatitude(loc.getLatitude());
+        address.setLongitude(loc.getLongitude());
+        return address;
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent intent) {
+        if (resultCode != -1) {
+            Log.e(TAG, "Error getting geocoding results");
```
barbeau: This is fine as a `Log.e` as it's indicating an error.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
...
+
+                if (!RegionUtils.isLocationWithinRegion(loc, region)) {
+                    it.remove();
+                }
+            }
+        }
+        return addresses;
+    }
+
+    public static LatLngBounds getRegionBounds(ObaRegion region) {
```
barbeau: Any code that references Maps API v2 objects like `LatLngBounds` must be in the `org.onebusaway.android.map.googlemapsv2` package, otherwise it will won't work on the Amazon build variant (and any other future non-Google variant).  I'd suggest trying to leave the utility methods where they are but work around the Maps v2 objects.  You can move utilities that specifically deal with Maps API v2 objects like this to `MapHelpV2`.  This might prove more difficult for some methods than others - if you hit a barrier let me know.

barbeau: An aside - the reason the code needs to stay in `org.onebusaway.android.map.googlemapsv2` package is because the Amazon build variant code is dynamically generated based on those classes in that package.  So it won't pick up any classes to generated outside that package.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
...
+ * Various utilities related to location data
+ *
+ * @author Khoa Tran
+ * @author Simon Jacobs (integration for onebusaway-android)
+ */
+
+
+public class LocationUtil {
+
+    private static final String TAG = "LocationUtil";
```
barbeau: This TAG needs to change, as it currently collides with the `LocationUtils` class.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
...
+    private static final int GEOCODER_MAX_RESULTS = 5;
+    //in meters
+    private static final int GEOCODING_MAX_ERROR = 100;
+
+    public static ArrayList<CustomAddress> processGeocoding(Context context, ObaRegion region,
```
barbeau: Looks like this method isn't used, it can be removed.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
...
+                                                            String... reqs) {
+        return processGeocoding(context, region, false, reqs);
+    }
+
+    public static ArrayList<CustomAddress> processGeocoding(Context context, ObaRegion region, boolean geocodingForMarker, String... reqs) {
+        ArrayList<CustomAddress> addressesReturn = new ArrayList<CustomAddress>();
+
+        String address = reqs[0];
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
```
barbeau: `prefs` isn't used here, it can be removed.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
...
+        boolean latLngSet = false;
+
+        try {
+            if (reqs.length >= 3) {
+                latitude = Double.parseDouble(reqs[1]);
+                longitude = Double.parseDouble(reqs[2]);
+                latLngSet = true;
+            }
+        } catch (Exception e) {
+            Log.d(TAG, "Geocoding without reference latitude/longitude");
```
barbeau: Should be `Log.e()`.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
barbeau: Ah - unless we need it for geocoding on Amazon build variant, now that you're using the Places widget, which I think is the case.

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/LocationUtil.java
```diff
@@ -0,0 +1,215 @@
+/*
```
barbeau: Let's get rid of this class and move the methods (minus `getRegionBounds(ObaRegion region)`) to the existing `LocationUtils` class.  I think they are generally useful and there isn't a good reason to split them into a separate utility class.

onebusaway-android/src/main/java/org/onebusaway/android/util/UIUtils.java
```diff
@@ -101,7 +101,6 @@ public static void setupActionBar(AppCompatActivity activity) {
         if (!(activity instanceof HomeActivity)) {
             bar.setDisplayHomeAsUpEnabled(true);
         }
-
```
barbeau: Extra whitespace change that can be removed.

onebusaway-android/src/main/java/org/onebusaway/android/directions/tasks/TripRequest.java
```diff
@@ -0,0 +1,349 @@
...
+import java.io.IOException;
+import java.lang.ref.WeakReference;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Iterator;
+import java.util.List;
+import java.util.logging.Logger;
```
barbeau: There are some unused imports here and in `DirectionsMapController`.  If you're using Android Studio, just check "Optimize imports" when you go to commit the code, and it will automatically remove unused imports.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+            return;
+        }
+
+        Place place = PlaceAutocomplete.getPlace(getActivity(), intent);
```
barbeau: Per https://developers.google.com/android/reference/com/google/android/gms/location/places/ui/PlaceAutocomplete.html#public-methods, this should use the Application Context, not the activity context:

`Place place = PlaceAutocomplete.getPlace(Application.get().getApplicationContext(), intent);`

onebusaway-android/src/main/res/layout/fragment_trip_plan.xml
```diff
@@ -0,0 +1,323 @@
...
+                                    android:layout_marginLeft="2dp"
+                                    android:layout_marginRight="2dp"/>
+
+                        </LinearLayout>
+
+                    </LinearLayout>
+                </FrameLayout>
+
+                <ImageButton
+                        android:background="@color/theme_primary"
```
barbeau: Please remove all the styling on this button, including the background, padding, and src image (and remove the `@drawable/plan_my_trip` from the project) - it should just be a normal material design button with the text "Plan my trip".

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+import java.util.Date;
+import java.util.Locale;
+
+
+public class TripPlanFragment extends Fragment {
+
+    private static final int FROM_ADDRESS_GEOCODE = 1;
+    private static final int TO_ADDRESS_GEOCODE = 2;
+
+    private TextView fromAddressTextArea;
```
barbeau: To following the coding conventions of the rest of the project, all class member variables should have the prefix `m` - should this would be `mFromAddressTextArea`.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanFragment.java
```diff
@@ -0,0 +1,464 @@
...
+            handlePlaceResult(place, toAddress, toAddressTextArea);
+        }
+
+    }
+
+    private void handlePlaceResult(Place place, CustomAddress address, TextView tv) {
+        address.setAddressLine(0, place.getName().toString());
+        address.setAddressLine(1, place.getAddress().toString());
+
+        LatLng loc = place.getLatLng();
```
barbeau: `Location` should be used here, again to keep all the Google imports specific to Maps API v2 in the `org.onebusaway.android.map.googlemapsv2` package.  You'll probably need to abstract the implementation of the Places widget into `ProprietaryMapHelpV2` there, and use the Android `Geocoder` for the Amazon build variant in the `ProprietaryMapHelpV2` under `src/amazon/java/org/onebusaway/android/map/googlemapsv2`.

barbeau: An aside - `ProprietaryMapHelpV2` contains the implementation related to the Maps API v2 or and/or Google Play Services that's specific to Google or another platform, and can't be shared simply by changing imports (like the rest of the classes in `org.onebusaway.android.map.googlemapsv2`).

onebusaway-android/src/main/res/layout/trip_plan_advanced_settings_dialog.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8"?>
```
barbeau: This dialog styling doesn't match the material design guidelines:

image

Background shouldn't be the theme color, and the button, combo box, text boxes, and fields should all use the default material design styling.  In many cases you should be able to get the default material design imply by removing the color attributes.  See https://material.google.com/components/dialogs.html#dialogs-specs for examples.

sdjacobs:
Thanks, Sean, I'm working on incorporating your comments. A couple questions:

Re ConversionUtils tests - most of ConversionUtils comes from OTP Android. Are tests for that class in development there? Would rather not duplicate effort if possible.

The mode and direction icons that are used in the TripResultsPage (see image below) are also taken from OTP Android. They are in grayscale rather than black. Should these also be changed or are they OK? If they should be changed, do you have suggestions for where I could find replacements?

screen shot 2016-06-15 at 2 36 16 pm

Re the dynamic generation of the drawer element - I had some code to do this, but it didn't work because there's no region set when the application is first opened, and the navigation drawer is not redrawn after the region is selected. So, if the app is opened fresh, then Tampa is selected, the app will need to be closed and reopened in order to get the "Plan a trip" item to show up. Any suggestions for this?

I'm still thinking about what to do for Amazon autocomplete. It sounds like right now I'd have to put the Google Places widget in Google's ProprietaryMapHelpV2 and the original autocomplete-from-Android-geocoder code in Amazon's ProprietaryMapHelpV2. I think instead it may make sense to revert to the autocomplete code, and just add into (the Google version of) ProprietaryMapHelpV2 the ability to get suggestions from Google Places [programmatically](https://developers.google.com/places/android-api/autocomplete#get_place_predictions_programmatically). The downside here is that Google Places' suggestions have Place IDs but not lat/longs, so another round trip to Google would be required to get the lat/long.

barbeau:
>Re ConversionUtils tests - most of ConversionUtils comes from OTP Android. Are tests for that class in development there? Would rather not duplicate effort if possible.

Unfortunately there isn't any active development for OTP Android going on at this time.

barbeau:
>The mode and direction icons that are used in the TripResultsPage (see image below) are also taken from OTP Android. They are in grayscale rather than black. Should these also be changed or are they OK? If they should be changed, do you have suggestions for where I could find replacements?

These should be black as well, so if we need the same icon elsewhere we can reuse it and dynamically recolor it.  We should have a bus icon you can use that's currently displayed in the `AgenciesFragment` for supported agencies (under Help) - `R.drawable.ic_maps_directions_bus`.  You should be able to find the other mode icons at https://design.google.com/icons/ or the older [Android Icon Generator site](https://romannurik.github.io/AndroidAssetStudio/icons-generic.html#source.space.trim=1&source.space.pad=0&size=24&padding=0&color=000%2C100&name=ic_example).  From a quick look I don't see direction arrows specifically on the material design icon site, although you could probably flip something like "subdirectory arrow" ones.

The other major site would be the [Noun Project](https://thenounproject.com/), which is probably where the original direction arrow icons came from.  We'd just need to be sure to add proper attribution in the "About" fragment to credit sources.  If you can find an SVG on the Noun Project, you can upload that to the Android Icon Generator site to generate the PNGs at different densities to drop into the project.

barbeau:
>Re the dynamic generation of the drawer element - I had some code to do this, but it didn't work because there's no region set when the application is first opened, and the navigation drawer is not redrawn after the region is selected. So, if the app is opened fresh, then Tampa is selected, the app will need to be closed and reopened in order to get the "Plan a trip" item to show up. Any suggestions for this?

Good question - I'd probably need to play with this.  My guess is that you need a callback from the `ObaRegionsTask` into the NavigationDrawerFragment that calls `.requestLayout` on the View that contains the items (which I think is `mDrawerItemsListContainer` from a quick look).

barbeau:
>I'm still thinking about what to do for Amazon autocomplete. It sounds like right now I'd have to put the Google Places widget in Google's ProprietaryMapHelpV2 and the original autocomplete-from-Android-geocoder code in Amazon's ProprietaryMapHelpV2. I think instead it may make sense to revert to the autocomplete code, and just add into (the Google version of) ProprietaryMapHelpV2 the ability to get suggestions from Google Places programmatically. The downside here is that Google Places' suggestions have Place IDs but not lat/longs, so another round trip to Google would be required to get the lat/long.

I'd prefer to avoid making several calls to the server - this can be painful on slow connections and increases the chances for failure.

After looking at this more, the `PlaceAutocomplete` class is stubbed out in the Google Play Services library, so the app does build and run with the class included on the Amazon Fire Phone.  So this means that you could implement this in `MapHelpv2` without needing a proprietary implementation for Google/Amazon variants.

Currently, when you click on the "From" or "To" boxes, the `StartAutocompleteOnClick.onClick()` listener is fired:

~~~
    private class StartAutocompleteOnClick implements View.OnClickListener {

        int requestCode;

        StartAutocompleteOnClick(int requestCode) {
            this.requestCode = requestCode;
        }

        @Override
        public void onClick(View v) {
            try {
                ObaRegion region = Application.get().getCurrentRegion();
                LatLngBounds bounds = LocationUtil.getRegionBounds(region);

                Intent intent =
                        new PlaceAutocomplete.IntentBuilder(PlaceAutocomplete.MODE_OVERLAY)
                                .setBoundsBias(bounds)
                                .build(getActivity());

                startActivityForResult(intent, requestCode);
            } catch (Exception e) {
                Log.e(TAG, "Error getting autocomplete: " + e.toString());
            }

        }
    }
~~~

So on Fire Phone I see the exception in the catch block:

`E/TripPlanFragment: Error getting autocomplete: com.google.android.gms.common.GooglePlayServicesRepairableException: Google Play Services not available`

I think we can move this code to `MapHelpV2` (mainly because of using `LatLngBounds`), and just guard the call to `PlaceAutocomplete.IntentBuilder` with a call to `GooglePlayServicesUtil.isGooglePlayServicesAvailable() `.  If this returns false, then we should use the Android Geocoder and the original autocomplete code (although I haven't seen the two implementations side-by-side).

Also FYI, looks like `GooglePlayServicesUtil.isGooglePlayServicesAvailable()` is now deprecated - see https://github.com/OneBusAway/onebusaway-android/issues/542.

sdjacobs:
The direction icons get pretty specific -- hard left, hard right, slightly left, slightly right, etc -- and I'm having trouble finding replacements on either Google Design or the Noun Project. It seems that these icons originally came from OTP: https://github.com/opentripplanner/OpenTripPlanner/tree/master/src/client/images/directions

I've also found another set of open source directions icons from [Mapbox](https://www.mapbox.com/blog/directions-icons/), and there's a request to google to release their directions icons: google/material-design-icons#341

What do you think we should do at this point? I prefer keeping the original OTP icons. I can create differently sized versions if necessary. Both sets of icons include gray parts as well as black parts - how should we handle this given that the icons should be colored dynamically?

Happy to investigate further if you have other ideas for where to find directions icons.

barbeau:
I think I'd prefer the Mapbox icons (and they were just released yesterday - good find!) - the original OTP icons aren't high enough resolution to look decent at larger sizes on higher resolution phones.  I think resizing the original OTP icons and making them look decent will be more work than just using the Mapbox icons (by running them through the Android Icon Generator).  What do you see the major downside to the Mapbox icons being?

For recoloring with more than one color, we could re-use what I implemented for dynamically re-coloring the vehicle marker icons - see `VehicleOverlay.getVehicleIcon()` and `UIUtils.colorBitmap()`.  This changes the black color to a different color, but leaves any other color alone.  It works reasonably well for re-coloring the vehicle marker icons to red, blue, green, and gray, while leaving the vehicle inside the marker white - if you look close you can see some artifacts on the edges where the black turns slightly to gray, which isn't replaced by the target color.  

Looking at the Mapbox direction arrow icon coloring:

image

...and the icons used by Google in Google Maps:

image

...they are pretty close - I think the Google arrows are a very dark gray, and the Mapbox ones black, but they are close enough that the Mapbox ones could pass for the Google material design ones.  So we might not need to re-color the direction arrows in their current state.

sdjacobs:
Ok, you've convinced me that the Mapbox icons are better. I'll put those in, and I'll see how it looks before I worry about whether or how to recolor them.

barbeau:
Cool!  Just drag/drop Mapbox SVG into Android Icon Generator and you should get the zip file with the directory structure to drop into OBA Android folders.  In generator I'd use trim = yes, a padding of 0% and 0dip and start with a size of 24dip - I think 24dip should be large enough for how big we'd want these icons.  You can always resize the ImageView in XML w/ max/min/Width/Height to tweak based on the exact presentation.

barbeau:
@sdjacobs One item I forgot to mention but is very important, especially if we keep the two fragments - TripPlanFragment and the TripResultsFragment.

If you plan a trip, and get the trip results, and then change the screen orientation from portrait to landscape, we lose the TripResultsFragment and it resets to the TripPlanFragment - in other words, the trip results disappear and you're back to the trip planning screen.  The state of the fragment should be maintained via `onSaveInstanceState()` and reconstructed after the orientation change.

barbeau:
In general, it's also good to test changing the orientation on every screen, including during when a trip plan is processing, to make sure nothing unexpected happens.  For example, I just tested the orientation change during the trip plan, and while the app doesn't crash, it does lose the progress dialog, and eventually pops up a "Planning not possible" prompt, and I see this stack trace in Android Studio:

~~~
E/WindowManager: android.view.WindowLeaked: Activity org.onebusaway.android.ui.TripPlanActivity has leaked window com.android.internal.policy.PhoneWindow$DecorView{1e2bfb0 V.E...... R....... 0,0-1368,764} that was originally added here
		at android.view.ViewRootImpl.<init>(ViewRootImpl.java:375)
		at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:299)
		at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:85)
		at android.app.Dialog.show(Dialog.java:322)
		at org.onebusaway.android.directions.tasks.TripRequest.onPostExecute(TripRequest.java:218)
		at org.onebusaway.android.directions.tasks.TripRequest.onPostExecute(TripRequest.java:60)
		at android.os.AsyncTask.finish(AsyncTask.java:651)
		at android.os.AsyncTask.access$500(AsyncTask.java:180)
		at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:668)
		at android.os.Looper.loop(Looper.java:148)
		at android.app.ActivityThread.main(ActivityThread.java:5525)
		at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
		at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
06-16 16:53:02.633 4677-4677/com.joulespersecond.seattlebusbot E/WindowManager: android.view.WindowLeaked: Activity org.onebusaway.android.ui.TripPlanActivity has leaked window com.android.internal.policy.PhoneWindow$DecorView{fa84336 V.E...... R......D 0,0-1368,464} that was originally added here
		at android.view.ViewRootImpl.<init>(ViewRootImpl.java:375)
		at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:299)
		at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:85)
		at android.app.Dialog.show(Dialog.java:322)
		at android.app.ProgressDialog.show(ProgressDialog.java:116)
		at android.app.ProgressDialog.show(ProgressDialog.java:99)
		at org.onebusaway.android.directions.tasks.TripRequest.onPreExecute(TripRequest.java:112)
		at android.os.AsyncTask.executeOnExecutor(AsyncTask.java:604)
		at android.os.AsyncTask.execute(AsyncTask.java:551)
		at org.onebusaway.android.directions.util.TripRequestBuilder.execute(TripRequestBuilder.java:255)
		at org.onebusaway.android.ui.TripPlanActivity.route(TripPlanActivity.java:70)
		at org.onebusaway.android.ui.TripPlanFragment$9.onClick(TripPlanFragment.java:235)
		at android.view.View.performClick(View.java:5201)
		at android.view.View$PerformClick.run(View.java:21215)
		at android.os.Handler.handleCallback(Handler.java:739)
		at android.os.Handler.dispatchMessage(Handler.java:95)
		at android.os.Looper.loop(Looper.java:148)
		at android.app.ActivityThread.main(ActivityThread.java:5525)
		at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
		at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
~~~

barbeau:
One item that @sdjacobs and @cagryInside and I talked about yesterday - we need some mechanism to report problems with trips per region.  I'm guessing this will be another email contact per region that goes straight to the OTP maintainer.  I think the biggest source of feedback will be failed trips, when OTP can't find a plan between the provided origin and destination.  This can happen quite frequently when using "Current location" as trip origin/destination, as it's not snapped to the street/sidewalk/transit network.

In the dialog that reports the planned trip failure:

image

...we should provide an immediate feedback link that sends the failed trip plan URL request to this contact point - something like the "Report" button in the ANR dialog that Android uses:

image

We may want to add another entry for "Trip planner" under "Send feedback" as well, that directs the user to plan the trip and then send the feedback via this mechanism, so we get the exact trip plan URL.

barbeau:
For the "better trip" reminder feature, can you please also add a preference to turn it on and off?

In testing I'm getting a lot of notifications, and I could see some users getting annoyed by these and wanting to turn them off.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripResultsFragment.java
```diff
@@ -0,0 +1,296 @@
...
+
+            DirectionExpandableListAdapter adapter = new DirectionExpandableListAdapter(
+                    getActivity(),
+                    R.layout.list_direction_item, R.layout.list_subdirection_item, direction_data);
+
+            mDirectionsListView.setAdapter(adapter);
+
+            mDirectionsListView.setGroupIndicator(null);
+
+            mRealtimeService.onItinerarySelected(itinerary, rank);
```
barbeau: We may want to move this into `onPause()`, so the notification only start when the user leaves the view of the trip planning results.  In some cases, I'm currently getting notified each time I tap on the different tabs for each trip plan result, which is confusing.

sdjacobs:
> For the "better trip" reminder feature, can you please also add a preference to turn it on and off?

Yes, I'm planning on doing this.

I'm looking at the "better trip" feature right now. I've implemented the following changes locally:
- When we requery, we don't update the time of the request to the current time, we use the original time that was set. The assumption is that the user leaves when the trip plan tells them too, and if they don't, they'd make a new query themselves.
- We notify the user if their selected itinerary doesn't appear in the returned itineraries, OR it does appear but it has a worse rank. (The previous behavior was to notify if it has a _different_ rank).
- Querying will end at the end time of the itinerary.

I think these changes reduce the false positives and are a lot closer to what we mean by "better trip." It's actually somewhat hard to find instances where the notification fires! I found one instance where I was able to find a short trip in downtown Boston that had one low-ranked option to take a bus, then a few minutes later that option disappeared. But it's hard, in general, to find reproducible cases. It's still a little worrying that the way OTP ranks the trips is opaque (to me, at least).

sdjacobs:
Here's another possibility: If your selected itinerary does appear in the list of returned itineraries, check the start times and end times of the transit legs. If they've changed by {2 minutes?} notify the user.

This would be a little closer to OTP Android. Again, though, it might be difficult to test.

barbeau:
>I think these changes reduce the false positives and are a lot closer to what we mean by "better trip." It's actually somewhat hard to find instances where the notification fires!

I think this is ok - I'd err towards more false negatives (notification didn't fire when it "should have") than false positives.

barbeau:
>If your selected itinerary does appear in the list of returned itineraries, check the start times and end times of the transit legs. If they've changed by {2 minutes?} notify the user.

This is a good idea - there are likely two situations where the user wants to be notified after planning a trip:

1.  Something happened on their trip, and they should take a different route
1.  Something happened on their trip, and they should take the same route (e.g., there is no other option), but the bus is running earlier or later than normal

The current OBA Android implementation covers the 1st but not the second, and the OTP Android implementation covers the 2nd but not the first.  Ideally we should cover both of these with a low false positive notification rate.  +/- 2 min is probably a good diff for 2nd.

sdjacobs:
I spend a little bit of time seeing if I could implement what would be minimally required for the trip plan fragment and trip results fragment to appear on the same page:
screen shot 2016-06-20 at 11 31 09 am

When any of the controls are updated, the trip will be replanned and the results redrawn.

The downside here is that there's much less screen real estate to see the map/list view. It would be nice if the results could be dragged over the trip plan controls, but I'm not sure of the best way to implement that. Any suggestions here? I only have today to add new UI code, so if there isn't a straightforward way to get more screen real estate, I'll revert to the old version and simply fix the margins and tabs.

barbeau:
Not sure about dragging over, but I think a more natural interaction would
be sliding whole screen up via scrollview.

@cagryInside did something similar for issue reporting, McCune he can chime
in.  I'm traveling today so probably won't be able to give more details
myself.

cagryInside:
> Not sure about dragging over, but I think a more natural interaction would
> be sliding whole screen up via scrollview.

I agree with @barbeau. IMO putting everything under a scrollview might be the best option. However, there is a challenge when you show the mapview under the scrollview. Mapview will get the focus and the screen won't scroll. You might need to use something like [this](http://www.londatiga.net/it/programming/android/how-to-make-android-map-scrollable-inside-a-scrollview-layout/) to make mapview scrollable.

sdjacobs:
I just pushed a commit with most of the UI changes we've been talking about. I used the [SlidingUpPanelLayout](https://github.com/umano/AndroidSlidingUpPanel) that's used elsewhere in OBA Android to be able to expand the trip results page (I found this before I saw the comments about the scroll view). I would appreciate your comments - there are still a few quirks with the UI, but I'd like to get some input before changing anything else. Depending on your comments, we can switch to a ScrollView or revert to the earler 2-page format.

barbeau:
@sdjacobs Thanks, looking at this now.

BTW, the OTP server seems very slow to respond for some trip plans.  Is the current OTP server the same server that will be used in production, or just a test server?

barbeau:
In my experience, using "Current Location" to set origin/destination frequently results in the trip plan failing as it's not close enough to a routable point in the OTP graph.  I went back to OTP Android and did a quick look at the code for the "Intelligent markers" feature (see `MainFragment` around line 702).  IIRC to avoid failing trips based on "Current Location" in OTP Android we reverse geocode the "Current Location" lat/long to an address, and then geocode this back to a lat/long for OTP trip plan request.  This way we get a location that's on the network, versus a random lat/long.  We may want to do something similar here, although we'd also need an option to turn it on or off (just like OTP Android) in case this pushes pedestrian start/depart locations away from pedestrian paths.

barbeau:
So I think the single-screen version is definitely an improvement over the 2 screen version.  My only concern is the map within a sliding panel or scroll view - I don't think this is going to work well because of the conflicting touch events.  I do like the sliding panel in general though - so I think the main question is how to move the map out of the sliding panel but still keep it in the UI.

@cagryInside and I spent some time talking about how to fix this without needing to do a lot of code changes, and we came up with the below mockup for flow, using the sliding panel:

image

This implementation is almost exactly the same as the HomeActivity and "Show route on map" view, in terms of the map being the main background view with other information laid on top of it:

1. Left screenshot - There is a top view (like the `RoutePopup` view in `RouteMapController`) that contains the trip planning fields, with the map being the main view behind it.  Ideally the user would be allowed to long press on the map here to set origin and destination too (like OTP Android)
1. Right screenshot - When trip planning is triggered, the trip planning header is hidden, and the results are shown, with the map on top holding the map result, and the sliding panel header the trip options, and the sliding panel contents holding the list of directions.  Here's how the panel would look slid up and down:

image

Note that the "Switch view" ribbon here is eliminated.  Hitting back button here in the Right screenshot would take you back to the Left screenshot.  I believe the sliding panel implementation itself (header and contents) for the above is exactly the same as your current implementation.

Please take a look at this and let us know what you think.

On a more minor note, there are still some alignment issues (which I think you already see) on the trip planning view:

image

sdjacobs:
> BTW, the OTP server seems very slow to respond for some trip plans. Is the current OTP server the same server that will be used in production, or just a test server?

I'm not sure yet. The server currently being used has a fairly old version of OTP (0.11.1) and I'm hoping we can upgrade it, as I think there have been some performance improvements with newer versions of OTP.

> In my experience, using "Current Location" to set origin/destination frequently results in the trip plan failing as it's not close enough to a routable point in the OTP graph

Can you give me a sample lat/long? I haven't noticed this with Boston, which I've been using on my phone. Would be very nice if this was also something that was fixed in a later OTP version...

sdjacobs:
I like your layout (and I'm glad I hadn't yet put time into making list view/map view tabs!). Am I correct that from the results page there is no way to see the trip plan information (origin, destination, time) without hitting the back button? A nice feature of the currently committed UI is that the trip selection controls can essentially double as a header for the results.

barbeau:
>Am I correct that from the results page there is no way to see the trip plan information (origin, destination, time) without hitting the back button? A nice feature of the currently committed UI is that the trip selection controls can essentially double as a header for the results.

Yes, with the current mockup, although I'm open to keeping the origin/destination at the top.  One possible implementation is starting with the sliding panel in the collapsed position - in this case, when the user scrolled up the top origin/destination view would push up, like the current routes view.  I'm open to other ideas too.  In keeping the origin/destination on top, though, we need to be very careful about the size of the map, and keeping it usable on small screen devices - that was my main motivation in dropping it from the above design.

@cagryInside and I both like Google Maps presentation of transit trips, which is partially what this is modeled after (although with some concessions to reduce implementation changes from the current implementation).  Here's how it works in Google Maps - main difference is that we don't present the trip options as a tappable list initially:

image

image

I'd welcome any design/implementation changes that get us closer to this presentation.

barbeau:
>Can you give me a sample lat/long?

Of course now I can't reproduce it - I'll post again when I get a lat/long that does this.

barbeau:
Here's an example of a failing URL for a trip plan, with "Current Location" as origin and "Tampa International Airport" as destination:
http://otp-fl.camsys-apps.com:8080/otp/routers/default/plan?wheelchair=false&showIntermediateStops=true&toPlace=27.9835,-82.5371&time=12:28PM&fromPlace=28.0587,-82.4165&date=06-24-2016&optimize=QUICK&arriveBy=false

sdjacobs:
> Here's an example of a failing URL for a trip plan.

I actually think the issue here is with the destination, not your Current Location origin.

I reproduced this on the emulator -  a trip could not be planned if I selected "Tampa International Airport, George J Bean Parkway..." as the destination, but it _could_ be planned if I used "Tampa International Airport, Tampa, FL, United States." (Both results appear as autocomplete suggestions while typing "Tampa International..."). Trips also failed with other origins and that Bean Parkway destination.

I'm not sure what we can do on the app itself to fix this behavior - I'd guess that the underlying issue is either with Google data or OSM data.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanActivity.java
```diff
@@ -201,4 +236,95 @@ public void onTripRequestComplete(List<Itinerary> itineraries) {
...
+            public void onClick(DialogInterface dialog, int which) {
+                clearBundleErrors();
+            }
+        });
+
+        feedback.setNegativeButton(R.string.report_problem_report, new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                // This will become a separate OTP email address
+                String email = Application.get().getCurrentRegion().getContactEmail();
```
sdjacobs: This line should be changed once a field for OTP maintainer is added.

barbeau:
@sdjacobs Some quick UI feedback looking at UI alone (not code - I'll try to elaborate more tomorrow):

* Scrolling map up/down sticks - If I plan a trip, I can't scroll the map up or down.  If I try to scroll up or down it "sticks" as the movement is consumed by the sliding panel instead.
* My location and clock icons - These should be tinted gray - see "Send feedback->Report Stop problem" for an example.
* Map directions list - I believe you need to set the scrollable view to the actual ListView here for the map directions list to interact properly with the sliding panel - see the HomeActivity/ArrivalsListFragment for a callback that passes the ListView from the ArrivalsListFragment to the HomeActivity after it's initiated.
* Report problem - Tapping on "Report" a trip planner problem doesn't seem to do anything?  Or is this still waiting on adding the actual `otpContactEmail` field?

sdjacobs:
> My location and clock icons - These should be tinted gray

Done.

> Report problem - Tapping on "Report" a trip planner problem doesn't seem to do anything? Or is this still waiting on adding the actual otpContactEmail field?

The code is there, but no email is sent if nothing is populated in otpContactEmail. I just added a message in a toast for when this occurs. I was able to get this code path to run properly by forcing `RegionUtils.getRegions` to load regions from the local regions resource. When the regions on the server are updated with otpContactEmail, this should work properly.

sdjacobs:
> Scrolling map up/down sticks - If I plan a trip, I can't scroll the map up or down. If I try to scroll up or down it "sticks" as the movement is consumed by the sliding panel instead.

Just added a fix for this. I updated the Umano dependency to the latest version, and then added a custom ScrollableViewHelper (new feature in Umano) to either a) proxy the call to the default implementation if the list view is showing, or b) return a positive position (ie more to scroll) if the map is showing. The upshot of this is that the list view and the map view will have the same scrolling behavior within the panel. Let me know what you think about this.

barbeau:
>Just added a fix for this. I updated the Umano dependency to the latest version, and then added a custom ScrollableViewHelper (new feature in Umano) to either a) proxy the call to the default implementation if the list view is showing, or b) return a positive position (ie more to scroll) if the map is showing. The upshot of this is that the list view and the map view will have the same scrolling behavior within the panel. Let me know what you think about this.

I'm still not a huge fan of the map within the sliding drawer, but this is definitely a big improvement in usability over the previous iteration.  I'll need to take a closer look at the HomeActivity changes, as previous library updates have broken some of the sliding panel behavior.  From a quick runs through the UI though it appears everything is still functioning correctly.

barbeau:
Not related to the sliding panel itself - I've noticed that when switching to the map view that I can start to move the map, but then it snaps back to the original trip plan bounds once.  After that it seems to let me pan freely.

barbeau:
A few more things that should be quick fixes:

* The tab headers should be white text on the dark theme color (same as Android system status bar at top of screen.  I probably use solid white (`header_text_color`) for the route names, and then the `header_text_faded_color` for the times and durations.

image

* The highlighted tab color should match the theme colors - `theme_accent` will likely work here, if that's not obvious enough then `theme_muted` or another green should work.

image

barbeau:
Actually, on second thought - to be consistent with the arrivals sliding panel, let's make the tab background `theme_primary`, and the selected tab color would be `theme_primary_dark`.

barbeau:
@sdjacobs Are you still planning on transitioning the strange "Switch view" button to tabs?

image

This is probably the last major design issue I have with this layout that really goes against Android guidelines.  Ideally we'd avoid these nested tabs altogether (and the map in the sliding panel) via a design like https://github.com/OneBusAway/onebusaway-android/pull/541#issuecomment-228109780, but we can always move towards that later.

sdjacobs:
I pushed some commits that address these issues. I agree that ultimately we should move to a different layout, but hopefully this will suffice for now.

It seems like tab layouts can either have icons or text, but not both (or at least not without some more involved configuration). The version I committed has text- which do you prefer? Or do you think it's worthwhile to try to have both? I can't decide which I like better.

screen shot 2016-07-14 at 2 11 01 pm
screen shot 2016-07-14 at 2 13 38 pm

sdjacobs:
I was able to add the icon to the list/map tabs:

screen shot 2016-07-15 at 11 56 11 am

Unfortunately I'm having a lot of trouble getting the images and text in the tab to align! Any suggestions? The resource file is [here](https://github.com/camsys/onebusaway-android/blob/25-tripPlan-all-changes/onebusaway-android/src/main/res/layout/tab_trip_results_view.xml). (For some reason, when layout_gravity=center for both ImageView and TextView it's even worse.)

barbeau:
Try:

~~~
    <ImageView
            android:layout_height="wrap_content"
            android:layout_width="wrap_content"
            android:id="@android:id/icon"
            android:layout_gravity="center_vertical"
            android:src="@drawable/ic_arrivals_styleb_action_map"
            android:tint="@color/navdrawer_icon_tint"
            android:layout_marginRight="2dp"/>
    <TextView
            android:id="@android:id/text1"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:layout_gravity="center"
            android:ellipsize="end"
            android:singleLine="true"
            android:textSize="@dimen/trip_results_tab_text_size"
            android:textColor="@color/body_text_1"
            android:textAllCaps="true"
            android:text="MAP VIEW"
    />
~~~

(minus the hard-coded image and text)

It should give you:

image

There are a lot of ways to do this (including `RelativeLayout`), but sometimes its fastest just to switch over to the design view and drag stuff where you want it, and see what the editor comes up with (that's how I got the above).  May need further manual tweaking (I manually added a bit of a margin to the right of map icon as well).

sdjacobs:
Did you try that layout in the app itself? When I try your version, it's centered in the design view, but not in the tab when I open the app.

barbeau:
No, didn't actually fire up app.  In app, is it the same as your previous
screenshots?

sdjacobs:
It's actually slightly more offset: 
screen shot 2016-07-15 at 1 07 02 pm

barbeau:
Hmmm...I'm not in front of my machine now but I'll try to take a look later.

In the meantime you might want to look at the ArrivalListHeader and layout
for the stop star and name at the bottom.  You might be able to use the
same nested layout design here.

sdjacobs:
I found a work-around by changing the icon to a nested drawable of the TextView. But please let me know if you find a better solution.

barbeau:
@sdjacobs Ok, that works for now.  IIRC that's the legacy tab solution for Recents right now.

Also, just talked to Brandon - we're going to slap a "beta" label on the trip planning feature in the app itself.  Please add this:

1. In the navigation drawer - "Plan a trip (beta)" - assuming it doesn't run into the margin
1. In the Action Bar - "Trip Planner (beta)"
1. In the preference description for the trip plan alerts

If you think of anywhere else it should be added, please let me know.  Also be on the heads up for further discussion with Brandon on rebase/merge schedule.

barbeau:
Also, please test rapid orientation changes (within 1-2 seconds) on the trip planning activity, including:

1. while it's actively planning a trip
1. while any dialogs are open

...to make sure behavior is reasonable and no crashes.  Users tend to see this when they drop their hand while walking, and then look at device again.

barbeau:
When this is ready to go, please squash down to one commit - thanks!

barbeau:
Usability/consistency comment - after you plan a trip and the sliding panel slides up to hide header and shows results, tapping the back button should slide the panel back down so you can see the origin/destination.  Tapping the back button a second time should then close the activity.

barbeau:
I'm also seeing this crash on this branch, which doesn't happen on `master` - I haven't had a chance to look and see why it's happening:

~~~
AndroidRuntime: FATAL EXCEPTION: main
   Process: com.joulespersecond.seattlebusbot, PID: 12970
   java.lang.RuntimeException: Unable to start activity ComponentInfo{com.joulespersecond.seattlebusbot/org.onebusaway.android.ui.HomeActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.onebusaway.android.ui.AlertList.getCount()' on a null object reference
	   at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2434)
	   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2494)
	   at android.app.ActivityThread.access$900(ActivityThread.java:157)
	   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1356)
	   at android.os.Looper.loop(Looper.java:148)
	   at android.app.ActivityThread.main(ActivityThread.java:5525)
	   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
	   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
	Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.onebusaway.android.ui.AlertList.getCount()' on a null object reference
	   at org.onebusaway.android.ui.ArrivalsListHeader.refreshError(ArrivalsListHeader.java:1224)
	   at org.onebusaway.android.ui.ArrivalsListHeader.refresh(ArrivalsListHeader.java:627)
	   at org.onebusaway.android.ui.HomeActivity$7.onPanelCollapsed(HomeActivity.java:1390)
	   at org.onebusaway.android.ui.HomeActivity$7.onPanelStateChanged(HomeActivity.java:1351)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.dispatchOnPanelStateChanged(SlidingUpPanelLayout.java:644)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.setPanelStateInternal(SlidingUpPanelLayout.java:1108)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.setPanelState(SlidingUpPanelLayout.java:1080)
	   at org.onebusaway.android.ui.HomeActivity.updateArrivalListFragment(HomeActivity.java:1067)
	   at org.onebusaway.android.ui.HomeActivity.setupMapState(HomeActivity.java:1500)
	   at org.onebusaway.android.ui.HomeActivity.onCreate(HomeActivity.java:309)
	   at android.app.Activity.performCreate(Activity.java:6272)
	   at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1107)
	   at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2387)
	   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2494) 
	   at android.app.ActivityThread.access$900(ActivityThread.java:157) 
	   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1356) 
	   at android.os.Looper.loop(Looper.java:148) 
	   at android.app.ActivityThread.main(ActivityThread.java:5525) 
	   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730) 
	   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620) 
~~~

Steps to repro:

1. Star a stop
1. Go to Starred Stops, long press on stop, create a home screen shortcut
1. Tap on the home screen shortcut to open arrivals for that stop
1. Tap on the map icon in the Action Bar

barbeau:
In the "Additional trip preferences" screen, please change "Wheelchair accessible" to "Avoid stairs".  I believe this is the only current consequence of checking this box, and I'd rather not imply that we consider other accessibility issues that the traveler might encounter.

barbeau:
Ok, production Regions API now has the `otpContactEmail`:
http://regions.onebusaway.org/regions-v3.json

onebusaway-android/src/main/java/org/onebusaway/android/report/ui/ReportActivity.java
```diff
@@ -16,13 +16,6 @@
 
 package org.onebusaway.android.report.ui;
```
barbeau: Looks like changes to this class are just import formatting - I'd prefer you remove this from the changeset.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/MapHelpV2.java
```diff
@@ -205,4 +205,5 @@ public static LatLng getClosestVehicle(ObaTripsForRouteResponse response,
 
         return makeLatLng(closestVehicleLocation);
     }
+
```
barbeau: Could you please remove this white space change to keep the changeset list of files changed clean?

barbeau:
I'm seeing some kind of race condition when planning a trip now.  Steps to repro:

1. Do clean install of this branch
1. Tap "Plan a trip"
1. Plan a trip from "Current location" to "Tampa International Airport"

I see the progress bar, and then it says "Planning not possible" in dialog, and keeps flashing this dialog and then closing it.  The screen is non-responsive as well.

barbeau:
I'm also seeing this crash on this branch and not on `master` - steps to repro:

1. Clean install of this branch
1. Tap on stop, then arrival time, then "Show route on map".
1. Change device orientation (e.g., portrait to landscape)

Here's the stacktrace:

~~~
E/AndroidRuntime: FATAL EXCEPTION: main
   Process: com.joulespersecond.seattlebusbot, PID: 32591
   java.lang.RuntimeException: Unable to start activity ComponentInfo{com.joulespersecond.seattlebusbot/org.onebusaway.android.ui.HomeActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.onebusaway.android.ui.AlertList.getCount()' on a null object reference
	   at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2434)
	   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2494)
	   at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:4120)
	   at android.app.ActivityThread.access$1000(ActivityThread.java:157)
	   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1362)
	   at android.os.Looper.loop(Looper.java:148)
	   at android.app.ActivityThread.main(ActivityThread.java:5525)
	   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
	   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
	Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.onebusaway.android.ui.AlertList.getCount()' on a null object reference
	   at org.onebusaway.android.ui.ArrivalsListHeader.refreshError(ArrivalsListHeader.java:1224)
	   at org.onebusaway.android.ui.ArrivalsListHeader.refresh(ArrivalsListHeader.java:627)
	   at org.onebusaway.android.ui.HomeActivity$7.onPanelCollapsed(HomeActivity.java:1390)
	   at org.onebusaway.android.ui.HomeActivity$7.onPanelStateChanged(HomeActivity.java:1351)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.dispatchOnPanelStateChanged(SlidingUpPanelLayout.java:644)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.setPanelStateInternal(SlidingUpPanelLayout.java:1108)
	   at com.sothree.slidinguppanel.SlidingUpPanelLayout.setPanelState(SlidingUpPanelLayout.java:1080)
	   at org.onebusaway.android.ui.HomeActivity.updateArrivalListFragment(HomeActivity.java:1067)
	   at org.onebusaway.android.ui.HomeActivity.setupMapState(HomeActivity.java:1485)
	   at org.onebusaway.android.ui.HomeActivity.onCreate(HomeActivity.java:309)
	   at android.app.Activity.performCreate(Activity.java:6272)
	   at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1107)
	   at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2387)
	   at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2494) 
	   at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:4120) 
	   at android.app.ActivityThread.access$1000(ActivityThread.java:157) 
	   at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1362) 
	   at android.os.Looper.loop(Looper.java:148) 
	   at android.app.ActivityThread.main(ActivityThread.java:5525) 
	   at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730) 
	   at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620) 
~~~

sdjacobs:
Okay, my recent commits address the issues you've raised so far, including the new crashes that were occurring on this branch but not on master. (They were because the newer Umano library dispatched events in more cases than the older code.)

> I'm seeing some kind of race condition when planning a trip now.

I couldn't reproduce this issue, but I pushed a commit this morning that corrected behavior relating to the planning-in-progress dialog. Could you try this again?

barbeau:
>I couldn't reproduce this issue, but I pushed a commit this morning that corrected behavior relating to the planning-in-progress dialog. Could you try this again?

I did a clean checkout, clean build, and I don't see this anymore - likely another casualty of Instant Run in Android Studio.

barbeau:
For later (likely) later server-side improvements - an example of a trip plan from "Current Location" that fails - I can change the start locatino to elsewhere and get a good trip:
http://otp-fl.camsys-apps.com:8080/otp/routers/default/plan?wheelchair=false&showIntermediateStops=true&toPlace=28.0286,-82.3930&time=03:02PM&fromPlace=28.0788,-82.4106&date=07-18-2016&optimize=QUICK&arriveBy=false

barbeau:
It looks like there is a layout issue on smaller screen/resolution devices, possibly related to Android 2.3 as well.

Here's what `TripPlanActivity` looks like on a Samsung Epic 4G from Sprint (SPH-D700 - basically a Galaxy S) w/ Android 2.3.6:

image

cagryInside:
> It looks like there is a layout issue on smaller screen/resolution devices, possibly related to Android 2.3 as well.

I see the same thing with 3.4" Nexus One emulator:
device-2016-07-18-232708

barbeau:
@cagryInside You beat me to it - I was trying to repo on a Nexus One emulator too, but arm is sooooo sloooow on Windows machine :)

barbeau:
When creating a Nexus One emulator w/ API 9 or 10 on Windows machine, when I try to start it I get:

~~~
Note: The environment variable $ANDROID_SDK_HOME is set, and the emulator uses that variable to locate AVDs.
This may result in the emulator failing to start if it cannot find the AVDs in the folder pointed to by the
given environment variable.
ANDROID_SDK_HOME=C:\APIs\android-sdk-windows

emulator: ERROR: This AVD's configuration is missing a kernel file! Please ensure the file "kernel-qemu" is in the same location as your system image.
emulator: ERROR: ANDROID_SDK_ROOT is undefined
~~~

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanActivity.java
```diff
@@ -0,0 +1,429 @@
...
+        // Set the height of the panel after drawing occurs.
+        final ViewGroup layout = (ViewGroup)findViewById(R.id.trip_plan_fragment_container);
+        ViewTreeObserver vto = layout.getViewTreeObserver();
+        vto.addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+            @Override
+            public void onGlobalLayout() {
+                layout.getViewTreeObserver().removeGlobalOnLayoutListener(this);
+                int viewHeight = findViewById(android.R.id.content).getHeight();
+                int height = layout.getMeasuredHeight();
+                mPanel.setPanelHeight(viewHeight - height);
```
sdjacobs: @cagryInside and @barbeau  - This is the code path that needs to be fixed for the lower res / older versions of Android. I need to set the panel height manually to make sure it doesn't cover the trip planning controls. This seems correct to me but clearly it isn't working for older Androids - any suggestions?

barbeau:
Looks like this is a function of the screen size / resolution, and not necessarily Android 2.3.  I can repro this on an Android emulator if I clone the default Nexus 5X emulator (with Android 6.0 w/ Google APIs), but then give it the screen dimensions/resolution of the older devices like the Nexus One - 3.7 inches, 480 x 800, hdpi - I gave it the Nexus One skin and 512MB RAM too, although I don't believe that matters:

image

image

barbeau:
Looks like this affects the "Additional Trip Preferences" dialog too - "ft" bleeds to the next line, just barely:

image

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripPlanActivity.java
```diff
@@ -0,0 +1,429 @@
...
+
+        // Set the height of the panel after drawing occurs.
+        final ViewGroup layout = (ViewGroup)findViewById(R.id.trip_plan_fragment_container);
+        ViewTreeObserver vto = layout.getViewTreeObserver();
+        vto.addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+            @Override
+            public void onGlobalLayout() {
+                layout.getViewTreeObserver().removeGlobalOnLayoutListener(this);
+                int viewHeight = findViewById(android.R.id.content).getHeight();
+                int height = layout.getMeasuredHeight();
```
barbeau: I thought that changing this to `layout.getHeight()` would resolve the problem (see http://stackoverflow.com/a/19524365/937715), but it doesn't seem to have any effect for me.  I'll keep looking...

sdjacobs:
I think I'm close... if I add in the action bar height, e.g.

<pre>
int abh = TripPlanActivity.this.getSupportActionBar().getHeight();
mPanel.setPanelHeight(viewHeight - height - abh);
</pre>

then the calculation is correct for the low-res version. Unfortunately then there's far too much space for the high-res version. Perhaps the action bar height isn't included in the content height in the high-res version, though I don't know why that would be the case.

Any idea what the right condition would be for when to add in the action bar height?

cagryInside:
@sdjacobs and @barbeau  the bug fix for the smaller sizes looks working except small clock icon is not showing properly.

3.4" Nexus One emulator screenshot:

screenshot_20160719-012511


Nexus 6 screenshot looks perfect:

device-2016-07-19-012204

barbeau:
@sdjacobs after resolving this small screen dimension/resolution issue please go ahead and rebase and squash.  I'd like to get this out to beta ASAP to catch other potential issues.

sdjacobs:
Ok - I rebased and created a new PR to preserve the history here. #602

barbeau:
Thanks! :+1: 

Closing this in favor of #602.

barbeau:
So the `onGlobalLayout()` fix improved the layout on the Samsung Epic, although it didn't completely resolve the issue:

image

It is usable now, although you can't set the arrive/depart by or depart time or date.  Here is what the trip plan results screen looks like now:

image

I'm ok pushing this into beta as-is, and we can continue to look to see if we can figure out a way to make the complete top section visible.

PR #547, opened 2016-06-17
Fix #544: Tapping on vehicle marker popup focuses on vehicle icon in TripDetailsListFragment.

### LINKED ISSUE #544: Tapping on vehicle marker popup should focus on vehicle icon in TripDetailsListFragment
**Summary:** 

Currently, if a user taps on an arrival time at a particular stop, the stop and vehicle icons both appear in TripDetailsListFragment and the stop is focused as the top visible record in the list - this is the desired behavior:

image

Prior to #463, we weren't showing the bus stop icon in TripDetailsListFragment if the user tapped on the vehicle marker popup balloon to enter TripDetailsListFragment.  In that case, the vehicle icon would take focus and appear as the first visible row in TripDetailsListFragment:

image

This is the desired behavior in terms of focusing on the vehicle as the first record, but the stop icon for the user's currently focused stop on the map was absent.  #463 fixes this to include the stop icon as well as the vehicle icon - however, a side effect of this is that if the stop icon is included, we always focus on the stop icon, and not the vehicle icon.  This isn't a problem with #463, but just an unintended consequence of the current code that sets the top visible list item, which always defaults to focusing on the stop if it's included.

IMO the behavior should be that if the user tapped on the vehicle marker popup balloon, we should focus on that vehicle as the first record (i.e., the same behavior prior to #463).

A fix for this is to allow the calling Activity to specify via Intent parameters whether the stop or vehicle should have focus as the first record.  If coming from ArrivalsListFragment and the user tapped on an arrival time, then use the Intent to set focus to the stop n TripDetailsListFragment.  If coming from the VehicleOverlay when a user tapped on the vehicle marker popup, then use the Intent to set focus to the vehicle in TripDetailsListFragment.

**Steps to reproduce:** 

1. Show a route
1. Tap on a vehicle marker
1. Tap on the vehicle marker balloon popup

**Expected behavior:** 

The vehicle icon should be the record at the top of the visible portion of the list:

image

**Observed behavior:** 

The stop icon is the record at the top of the visible portion of the list:

image

Stock LG G4 w/ Android 6.0

**Screenshots:** 

See above

azizmb9494:
If a vehicle marker popup is tapped, it attempts to focus on the stop with the bus icon first. If stop marker is selected, it attempts to focus on that selected stop.

Testing Done:
1. Selected stop, tapped "Show Route on Map", clicked vehicle overlay and tapped popup. Trip Details Activity opened with bus focused. Worked as wanted.
2. Selected stop, tapped "View Trip Details". Trip Details Activity opened with stop focused. Worked as wanted.
3. gradlew connectedObaGoogleDebugAndroidTest connectedObaAmazonDebugAndroidTest: Success.

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
@@ -45,6 +45,7 @@ import org.onebusaway.android.io.elements.ObaTripStatus;
 import org.onebusaway.android.io.request.ObaTripsForRouteResponse;
 import org.onebusaway.android.ui.ArrivalInfo;
 import org.onebusaway.android.ui.TripDetailsActivity;
+import org.onebusaway.android.ui.TripDetailsListFragment;
 import org.onebusaway.android.util.MathUtils;
 import org.onebusaway.android.util.UIUtils;
 
@@ -300,9 +301,11 @@ public class VehicleOverlay implements GoogleMap.OnInfoWindowClickListener {
         ObaTripStatus status = mMarkerData.getStatusFromMarker(marker);
 
         if (mController != null && mController.getFocusedStopId() != null) {
-            TripDetailsActivity.start(mActivity, status.getActiveTripId(), mController.getFocusedStopId());
+            TripDetailsActivity.start(mActivity, status.getActiveTripId(),
+                    mController.getFocusedStopId(), TripDetailsListFragment.SCROLL_MODE_VEHICLE);
         } else {
-            TripDetailsActivity.start(mActivity, status.getActiveTripId());
+            TripDetailsActivity.start(mActivity, status.getActiveTripId(),
+                    TripDetailsListFragment.SCROLL_MODE_VEHICLE);
         }
     }
 
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
@@ -16,18 +16,6 @@
  */
 package org.onebusaway.android.ui;
 
-import org.onebusaway.android.R;
-import org.onebusaway.android.io.ObaApi;
-import org.onebusaway.android.io.elements.ObaReferences;
-import org.onebusaway.android.io.elements.ObaRoute;
-import org.onebusaway.android.io.elements.ObaStop;
-import org.onebusaway.android.io.elements.ObaTrip;
-import org.onebusaway.android.io.elements.ObaTripSchedule;
-import org.onebusaway.android.io.elements.ObaTripStatus;
-import org.onebusaway.android.io.request.ObaTripDetailsRequest;
-import org.onebusaway.android.io.request.ObaTripDetailsResponse;
-import org.onebusaway.android.util.UIUtils;
-
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.drawable.GradientDrawable;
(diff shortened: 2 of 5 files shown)


onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -300,9 +300,10 @@ public void onInfoWindowClick(Marker marker) {
         ObaTripStatus status = mMarkerData.getStatusFromMarker(marker);
 
         if (mController != null && mController.getFocusedStopId() != null) {
-            TripDetailsActivity.start(mActivity, status.getActiveTripId(), mController.getFocusedStopId());
+            TripDetailsActivity.start(mActivity, status.getActiveTripId(),
+                    mController.getFocusedStopId(), "vehicle");
```
barbeau: "vehicle" string should be moved to a constant `SCROLL_MODE_VEHICLE` in `TripDetailsListFragment`.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -300,9 +300,10 @@ public void onInfoWindowClick(Marker marker) {
...
         } else {
-            TripDetailsActivity.start(mActivity, status.getActiveTripId());
+            TripDetailsActivity.start(mActivity, status.getActiveTripId(), null, "vehicle");
```
barbeau: I'd prefer creating a new overloaded `TripDetailsActivity.start()` method without stopId so we don't have to pass in `null` here.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
```diff
@@ -26,6 +26,7 @@
 import org.onebusaway.android.io.elements.ObaTripStatus;
 import org.onebusaway.android.io.request.ObaTripDetailsRequest;
 import org.onebusaway.android.io.request.ObaTripDetailsResponse;
+import org.onebusaway.android.tripservice.BootstrapService;
```
barbeau: Looks like this import can be removed - in Android Studio, when committing you can check "Optimize imports" to do this automatically.

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
```diff
@@ -228,19 +234,8 @@ private void setTripDetails(ObaTripDetailsResponse data) {
...
-                    listView.post(new Runnable() {
-                        @Override
-                        public void run() {
-                            listView.setSelection(mStopIndex);
-                        }
-                    });
-                }
-            } else {
-                // If we don't have a stop, then scroll to the current position of the bus
+            if (mScrollMode != null && mScrollMode.equals("vehicle")) {
```
barbeau: To simplify this section of code, this can be:

~~~
if (mScrollMode == null) {
    // Don't automatically scroll to stop or vehicle icon
    return;
}
~~~

So, if no scroll mode is provided, we do nothing.  And then do a `switch` for the two scroll modes.  This `switch` should probably be in it's own private method to avoid cluttering this one too much.

barbeau:
Thanks @azizmb9494!  This generally looks good, a few comments inline.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
barbeau: We'll also need to add a `SCROLL_MODE_STOP` in `TripDetailsListFragment`, and in `ArrivalsListFragment.goToTripDetails()` set the stop mode.

azizmb9494:
Thanks @barbeau, fixed!

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripDetailsListFragment.java
```diff
@@ -704,4 +689,65 @@ public Integer getNextStopIndex() {
...
+                    if (mStopIndex != null) {
+                        listView.post(new Runnable() {
+                            @Override
+                            public void run() {
+                                listView.setSelection(mStopIndex);
+                            }
+                        });
+                    }
+                }
+            }
```
barbeau: I think you're missing a `break` here - testing the PR w/ breakpoints it seems to execute both `case` blocks.

barbeau:
@azizmb9494 Sorry it's taken so long to review!.  Looks good in general, but I believe there is a missing `break` statement (see in-line comment), other than that it looks good.  If you can tweak that we can get this merged!

azizmb9494:
No worries, fixed @barbeau

barbeau:
Thanks @azizmb9494, looks good!

PR #556, opened 2016-06-24
Add ability to open Infrastructure Issue Activity with context

cagryInside:
Added a new method to open `InfrastructureIssueActivity` with a `context` instead of an `activity`.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
@@ -80,6 +80,8 @@ public class InfrastructureIssueActivity extends BaseReportActivity implements
         ReportProblemFragmentCallback, IssueLocationHelper.Callback,
         SimpleArrivalListFragment.Callback, GeocoderTask.Callback {
 
+    private static final int REQUEST_CODE = 0;
+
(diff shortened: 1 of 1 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/report/ui/InfrastructureIssueActivity.java
```diff
@@ -172,29 +174,41 @@
...
-     * @param activity The context of the activity.
+     * @param activity The parent activity.
      * @param intent   The Intent containing focusId, lat, lon of the map
      */
     public static void start(Activity activity, Intent intent) {
-        activity.startActivityForResult(makeIntent(activity, intent), 0);
+        activity.startActivityForResult(makeIntent(activity, intent), REQUEST_CODE);
     }
 
     /**
```
barbeau: @cagryInside Could you add a little more information to the Javadoc here of how `.startActivityForResult()` is used (e.g., to close the calling activity)?  For future developers it would be good to document the difference in behavior here from the one that takes a `Context` as a parameter.

barbeau:
@cagryInside Thanks, looks good!  Just one comment on adding more Javadoc description.

cagryInside:
> @cagryInside Thanks, looks good! Just one comment on adding more Javadoc description.

It should be good now.

barbeau:
Thanks @cagryInside!

PR #557, opened 2016-06-24
Fix #551 - Use previous map center as default map view

### LINKED ISSUE #551: Map doesn't automatically recenter on last map position on fresh startup
**Summary:** 

First reported in #525:

>Without location on, this app seems to think I'm in the middle of Africa. It also seems to glitch out and lose my spot every time I open it and while I'm actively using it.

Also reported by another user:

>every time I start the app it puts my location on the equator somewhere near Nigeria. (I assume that it's 0 lat, 0 long.)

>I'm in the Puget Sound region and the app remembers that so it asks if I want to move there or stay in the middle of the Atlantic Ocean. I take the offer to move and I then see the entire Puget Sound region and have to zoom in to find Seattle. Why can't it just remember my last location like the last version?

>I keep GPS turned off because a) it drains my battery, b) doesn't work in my office building anyway, c) I don't need to be tracked everywhere.
>I see the new 'starred stops' option, which is great, so I'm using that but simply remembering my last view in the app is more likely to be useful than 'somewhere off the coast of Africa'.  I think that your assumption that everyone has GPS or has it always turned on is OK but when it's not turned on you need a better fallback.

I think the behavior reported above is a bug, but I'm not yet sure if it's with OBA Android or Maps API v2.  It may be that we didn't sufficient test with location services off in the situation where the app completely gets killed by the user or platform.

So, separate from the enter being able to enter a home zip code to set their position, we should revert to the same map position as the user last saw in the app when they do a fresh start up of the app.

cc @azizmb9494 as he's been working on the implementation of #525 in PR #527.  I think the behavior we probably want in #527 is that if the user enters a zip code and hits the "My location" button, we re-center over the zip code.  If the user doesn't hit "My location," then we want to duplicate their previous view of the map when they last closed the app.

**Steps to reproduce:** 

Not certain, but I assume:

1. Kill the app completely
1. Turn off location services on the phone
1. Start up the app

**Expected behavior:** 

Map should center on the most recent map position, with the previous zoom level - this was the behavior of v1.x.

**Observed behavior:** 

Map re-centers on lat/long (0,0) with a world-level zoom level

* Device Galaxy Note4 (trltetmo), App version 62 (2.0.12), OS Android 5.1
* App Version: 2.0.13, Model: SM-G386T, OS Version: 4.4.2 / 19

barbeau:
Still need to do some further testing, and perhaps some code cleanup, but I think this should hopefully fix #551.  I'd welcome other's testing as well!

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -16,25 +16,6 @@
  */
 package org.onebusaway.android.map.googlemapsv2;
 
-import android.app.Activity;
-import android.app.Dialog;
-import android.content.Context;
-import android.content.DialogInterface;
-import android.content.Intent;
-import android.graphics.drawable.Drawable;
-import android.location.Location;
-import android.os.Bundle;
-import android.os.Handler;
-import android.provider.Settings;
-import android.support.v4.graphics.drawable.DrawableCompat;
-import android.support.v7.app.AlertDialog;
-import android.text.TextUtils;
-import android.util.Log;
-import android.view.LayoutInflater;
-import android.view.View;
-import android.view.ViewGroup;
-import android.widget.Toast;
-
 import com.google.android.gms.maps.CameraUpdateFactory;
 import com.google.android.gms.maps.LocationSource;
 import com.google.android.gms.maps.OnMapReadyCallback;
@@ -63,7 +44,27 @@ import org.onebusaway.android.map.StopMapController;
 import org.onebusaway.android.region.ObaRegionsTask;
 import org.onebusaway.android.util.LocationHelper;
 import org.onebusaway.android.util.LocationUtils;
+import org.onebusaway.android.util.PreferenceUtils;
 import org.onebusaway.android.util.UIUtils;
+
+import android.app.Activity;
+import android.app.Dialog;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.graphics.drawable.Drawable;
+import android.location.Location;
+import android.os.Bundle;
+import android.os.Handler;
+import android.provider.Settings;
+import android.support.v4.graphics.drawable.DrawableCompat;
+import android.support.v7.app.AlertDialog;
+import android.text.TextUtils;
+import android.util.Log;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Toast;
 
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


barbeau:
So this PR changes the behavior when locations services are turned on as well - now, the map will automatically be restored to the previous map center and zoom, even if location services are available.  Previously, if location services were turned on and available, the map would always center on the user's location.  IMHO this new behavior is correct - it matches the behavior of Google Maps.  If the user wants to re-center the map on their location, they can just hit the "My Location" floating action button.

@cagryInside Could you please test this PR a bit, just to make sure I didn't miss anything?  It should always recenter the map/zoom to the previous state, even when the app is completely killed, whether or not location services are enabled.  Please also double-check to make sure that centering the map on stops and routes still work, when selecting "Show on map" options from arrivals list and trip status screens.  This all appears to work fine for me, but a second pair of eyes would be helpful.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -233,10 +234,14 @@ private void initMap(Bundle savedInstanceState) {
...
             Bundle args = getActivity().getIntent().getExtras();
-            // The rest of this code assumes a bundle exists,
-            // even if it's empty
+            // The rest of this code assumes a bundle exists, even if it's empty
             if (args == null) {
                 args = new Bundle();
+                Location l = getMapCenterAsLocation();
+                if (l == null || l.getLatitude() == 0.0 || l.getLongitude() == 0.0) {
+                    // We don't have a map view location yet - try getting last one from prefs
+                    PreferenceUtils.maybeRestoreMapViewToBundle(args);
```
barbeau: @cagryInside Could you please try debugging this and seeing if the code makes it into this IF branch on your device?  It's possible that this could be initializing differently on different devices, or versions of Google Play Services.  We may want to completely remove the IF here and always execute PreferenceUtils.maybeRestoreMapViewToBundle().

barbeau:
@cagryInside I just remove the IF and force-pushed this to this PR - let me know if this fixes the issue now on your device.

barbeau:
Today @cagryInside and I tested this on his Nexus 6 w/ Android 6.0.1 and it doesn't work (i.e., there is no change in behavior from the `master` branch).  It does work on my LG G4 w/ Android 6.0.  I just tested this on a Nexus 5 API Level 23 Google APIs emulator and it doesn't work.  I'm not sure yet what the difference is...

barbeau:
Looks like this is an async/multithreading issue - on the Nexus 5 emulator, if I set a break point in `StopMapController.setState()` on line 180 `Location center = null;` and if I step through to the end of the `.setState()` method and then hit play to resume execution, the map correctly centers on the previous map center and zoom level.  So on the Nexus 5 and 6, something seems to either be pushing the map back to `(0,0)` after centering on the previous map location, or blocking the centering on the previous map location from happening at all.

barbeau:
The underlying cause of the different behavior on different devices is a strange one - apparently in `BaseMapFragment.initMap()`, calling `getActivity().getIntent().getExtras()` on Nexus 5 emulator w/ Android 6.0 and Nexus 6 w/ Android 6.0.1 returns a mysterious Bundle (which we're assuming is generated by the platform - it contains a `Parcelable` not created by our app) after killing the Activity.  On LG G4, `getActivity().getIntent().getExtras()` returns null after killing the Activity, which seems to be the behavior of previous Android devices as well, based on the design of the code here  - the code assumed that the Bundle would be null if the app didn't write anything to the Bundle, and used that to indicate that the default OBA map loading behavior should occur.

We changed our design to always write the lat/long to the Bundle if it doesn't exist (whether or not the Bundle is null to begin with), so the default behavior is now to center on the previous map position.  This works across the Nexus 5 emulator, Nexus 6, and LG G4.

PR #567, opened 2016-06-28
Fix #558 - Add AlarmReceiver for legacy reminders

### LINKED ISSUE #558: Reminders don't trigger/update unless app is opened
**Summary:** 

Reported by a user via email:

> Love the app.  However, after a restart of my phone, my reminders won't trigger until I open the app once.

**Steps to reproduce:** 

Unknown (I haven't tried to reproduce yet), but I'm assuming:
1. Set a reminder
2. Restart the phone
3. Do NOT start the app

**Expected behavior:** 

Reminder triggers based on when it was set

**Observed behavior:** 

Reminder doesn't trigger

> App Version: 2.0.13
> Model: Nexus 5
> OS Version: 6.0.1 / 23

barbeau:
~~*Please do not merge*~~

~~Initial WIP for fixing legacy reminders.  I haven't tested this yet, but will soon, and would welcome testing by others.~~

I don't believe a legacy BootReceiver is needed, as that is scheduled by the system.

cc @cagryInside

### DIFF
--- a/onebusaway-android/src/main/AndroidManifest.xml
+++ b/onebusaway-android/src/main/AndroidManifest.xml
@@ -230,6 +230,14 @@
                 <action android:name="android.intent.action.TIMEZONE_CHANGED"/>
             </intent-filter>
         </receiver>
+        <!-- Legacy Receiver for backwards-compatibility support with v1.x - See Issue #558 -->
+        <receiver android:name="com.joulespersecond.seattlebusbot.AlarmReceiver">
+            <intent-filter>
+                <!-- Should match constants for actions defined in TripService -->
+                <action android:name="com.joulespersecond.seattlebusbot.action.SCHEDULE"/>
+                <action android:name="com.joulespersecond.seattlebusbot.action.POLL"/>
+            </intent-filter>
+        </receiver>
 
         <meta-data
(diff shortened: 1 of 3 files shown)


barbeau:
Ok, this should be ready for merge (after Travis confirms it builds fine).

I've tested the following scenarios:

**Scenario A**:

1. Install v1.7.9
1. Create reminder
1. Install this PR, and do NOT start app

and:

**Scenario B:**

1. Install v1.7.9
1. Create reminder
1. Install this PR, do NOT start app, and immediately power cycle device.  Do NOT start app after device power cycles.

In both scenarios, the reminder triggers as intended without needing to open the app.  

In Scenario A, the first time the `AlarmReceiver` triggers based on the Intent set in v1.7.9, it's received by the legacy `AlarmReceiver` (i.e., the one added in this PR).  During this execution, the `TripService` runs and registers the next alarm with the current `AlarmReceiver`.  Subsequent Intents are then received by the current `AlarmReceiver`.

In Scenario B, the `BootReceiver` registers all reminders with the current `AlarmReceiver` via the `TripService` when boot is complete, so the legacy `AlarmReceiver` never receives any Intents (which is fine).

So, this PR addresses Scenario A, which previous failed, but with this PR now works as expected.

PR #576, opened 2016-07-01
Fix #561 - Consistently format trip headsign

### LINKED ISSUE #561: Some headsigns have inconsistent upper/lower case
**Summary:** 

Trip headsigns are treated inconsistently in term of case within the app.  For example, in the below screenshot, in the sliding panel header the headsign is shown as "UTC", but in the arrivals list below it's shown as "Utc".  There is a history of data issues in OBA, including ALL CAPS HEADSIGNS IN PUGET SOUND which is why title casing is currently applied.

image

Similar issue in OBA iOS - https://github.com/OneBusAway/onebusaway-iphone/issues/635.  Solution proposed there and in in Google Group discussion at https://groups.google.com/forum/#!topic/onebusaway-developers/ji4tXePwB10 is to only apply title case if all headsign characters are uppercase.  Ideally the transit agencies just fix their data, so we don't have to do this kind of post-processing (also discussed in above Google Group post).

**Steps to reproduce:** 

Look at arrivals at the above stop in San Diego, or Tampa stop 6497.

**Expected behavior:** 

Be consistent with case in headsigns, and all headsigns should be title case, but with acronyms like "UTC" remaining upper case.

**Observed behavior:** 

Trip headsigns have different case.

N/A

**Screenshots:** 

See above

barbeau:
~~**Please don't merge yet**~~

* If headsign is all caps, then convert to title case.  Otherwise, don't change it.

~~TODO - test!~~

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/MyTextUtilsTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/MyTextUtilsTest.java
@@ -27,4 +27,9 @@ public class MyTextUtilsTest extends AndroidTestCase {
         assertEquals("Seattle", MyTextUtils.toTitleCase("SEATTLE"));
         assertEquals("Seattle", MyTextUtils.toTitleCase("Seattle"));
     }
+
+    public void testIsAllCaps() {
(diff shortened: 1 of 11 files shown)


barbeau:
I've tested this and it works fairly well:

image

image

PR #584, opened 2016-07-06
Fix #568 - Color-code Style B real-time arrival/departure times

### LINKED ISSUE #568: Change color of schedule/real-time arrival times in Style B cards
**Summary:** 

We should change the scheduled/real-time text colors for Style B cards to match the rest of OBA.  The current colors are a hold-over from the original York design.

We use gray to indicate schedule arrivals in the classic Style A OBA arrival cards, and vehicle markers on the map:

image

In the Style B cards, we are showing schedule times in black and real-time arrival info in gray:

image

We should change the schedule times to the gray we use elsewhere in the app, and the real-time arrival times to either black or colored to match the status labels.

barbeau:
* Now matches Style A in color - scheduled is gray, late blue, on time green, and early red
* Add utility method and unit test for adding alpha value to an int color

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
@@ -16,6 +16,7 @@ import org.onebusaway.android.provider.ObaContract;
 import org.onebusaway.android.ui.ArrivalInfo;
 import org.onebusaway.android.util.UIUtils;
 
+import android.graphics.Color;
 import android.support.v4.util.Pair;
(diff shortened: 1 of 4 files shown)


barbeau:
This PR currently uses the slightly transparent colored times (as discussed in https://github.com/OneBusAway/onebusaway-android/issues/568#issuecomment-230849833):

image

barbeau:
Just force-pushed a new commit for this PR - I did a little more work to de-emphasize the schedule times by removing the bold of the next scheduled arrival, and bold the real-time predictions to both emphasize them more and make the smaller real-time values easier to read.  Smaller schedule times text size also increased to make them easier to read (but no bold).

image

PR #610, opened 2016-07-20
Fix #608: hide trip planning preferences 

### LINKED ISSUE #608: Trip planning notification preference is visible in regions without OTP URL
**Summary:** 

The preference under "Notifications" related to trip planning is visible for all regions.  It should be hidden and shown based on the presence of a regional OTP base URL, just like the main "Plan a trip" option in the navigation drawer.  There is already some code in `PreferencesActivity.onCreate()` (~line 124) for hiding preferences, although these changes aren't driven by region changes (they are based on build variants).  This visibility change would be based on region changes, so we'd likely need to implement in `onResume()`.

**Steps to reproduce:** 

1. Set the region to one without an OTP server (e.g., Puget Sound)
1. Open the navigation drawer and tap on Settings

**Expected behavior:** 

Notifications setting entry "Trip plan notifications" should be hidden.  If "Trip plan notifications" is the only preference under the "Notifications" category, then the "Notifications" category should be hidden as well.

**Observed behavior:** 

Notifications setting entry "Trip plan notifications" is visible.

Reprod on stock LG G4 w/ Android 6.0

**Screenshots:** 

image

@sdjacobs Could you please take a shot at this?

sdjacobs:
Hide trip planning preferences for regions that do not have otpBaseUrl defined. If notifications category is now empty, hide that too.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
@@ -38,6 +38,7 @@ import android.preference.Preference;
 import android.preference.Preference.OnPreferenceChangeListener;
 import android.preference.PreferenceActivity;
 import android.preference.PreferenceCategory;
+import android.preference.PreferenceGroup;
 import android.preference.PreferenceScreen;
 import android.support.v7.app.AlertDialog;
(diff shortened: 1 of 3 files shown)


barbeau:
Thanks @sdjacobs, looks good!

PR #616, opened 2016-07-20
Fix #614 - Make trip planning immediately available after app update

### LINKED ISSUE #614: Trip planning isn't immediately available after update
**Summary:** 

When installing `master` branch as an update to the old version, the "Plan a trip" functionality doesn't immediately show up in the navigation drawer, and the "What's New" message that is shown is the one without trip planning.  This is likely due to the Regions API contents not being immediately refreshed on update, and therefore either memory or the database does not contain the OTP base URL.

Workaround - close the app and re-open it, and then trip planning is available, and if you go to "Help=>What's New" you get the trip planning message.

**Steps to reproduce:** 

1. Be in Tampa, or fake the location in the emulator
1.  Install the v2.0.16 tag - https://github.com/OneBusAway/onebusaway-android/tree/v2.0.16
1.  Checkout `master` branch, bump `versionCode` in `build.gradle` and install

**Expected behavior:** 

Give me the "What's New" message that contains the message about trip planning, and show the "Plan a trip" option in the navigation drawer.

**Observed behavior:** 

There is no sign of trip planning - the "What's New" message is the one for regions without OTP base URLs, and "Plan a trip" isn't shown in the navigation drawer.

Stock LG G4 w/ Android 6.0

@sdjacobs I'm going to start working on this, but if you have a chance to look into it I'd welcome another pair of eyes.

barbeau:
* Alter ObaRegionsTask to take in multiple Callbacks
* Make HomeActivity one of these Callbacks, and move the code to refresh the navigation drawer to execute after the Regions API refresh in HomeActivity.onRegionChanged(), but only if the region changed OR the app was just updated.  This also removes the tight coupling between BaseMapFragment and HomeActivity for refreshing the navigation drawer, so the drawer will be refreshed even if the user doesn't have the "Nearby" (BaseMapFragment) screen shown
* "What's New" dialog is now triggered after the Regions API is refreshed in HomeActivity.onRegionChanged(), so it can use the new contents of the Regions API to control whether trip planning is shown as a new feature (it's currently only available in Tampa)
* Move code to show Toast on region change from BaseMapFragment to HomeActivity so this Toast will be shown even if the user doesn't have the "Nearby" (BaseMapFragment) screen visible

@sdjacobs Could you please take a look at this to make sure I'm not missing anything?

### DIFF
--- a/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -544,25 +544,6 @@ public class BaseMapFragment extends SupportMapFragment
                                 mapCenter.getLongitude() == 0.0))) {
             zoomToRegion();
         }
-
-        // If region changed and was auto-selected, show user what region we're using
-        if (currentRegionChanged
-                && Application.getPrefs()
-                .getBoolean(getString(R.string.preference_key_auto_select_region), true)
-                && Application.get().getCurrentRegion() != null
-                && mRunning
-                && UIUtils.canManageDialog(getActivity())) {
-            Toast.makeText(getActivity(),
-                    getString(R.string.region_region_found,
-                            Application.get().getCurrentRegion().getName()),
-                    Toast.LENGTH_LONG
-            ).show();
-        }
-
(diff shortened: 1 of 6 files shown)


sdjacobs:
Looks good to me.

barbeau:
Thanks!  I'll wait for this to pass on Travis and then merge and try to get this out the door again.

PR #631, opened 2016-07-25
Fix #611 - Trip planning - format "Leaving" text

### LINKED ISSUE #611: Trip planning - formatting of "Leaving" text is different
**Summary:** 

On the trip planning screen, the formatting and text size of the "Leaving" view (which is a Spinner) is different from the date and times, which are EditTexts.  For example, the text size of "Leaving" appears to be a bit bigger, and the triangle next to the text is a different size and color and has larger spacing.  We should make all 3 views look the same.

image

**Steps to reproduce:** 

1. Tap on navigation drawer and select "Plan a trip"

**Expected behavior:** 

Have consistent formatting between "Leaving", date and time text/views.

**Observed behavior:** 

"Leaving" view text is slightly larger, and has different color/shape triangle next to it, with a larger margin between the text and triangle.

Stock LG G4 w/ Android 6.0.

**Screenshots:** 

See above

cagryInside:
| Before   |      After      |
|----------|:-------------:|
|device-2016-07-25-134003 | device-2016-07-25-135012
|

### DIFF
--- a/onebusaway-android/src/main/res/layout/fragment_trip_plan.xml
+++ b/onebusaway-android/src/main/res/layout/fragment_trip_plan.xml
@@ -131,6 +131,7 @@
                     android:focusable="false"
                     android:layout_gravity="center_vertical"
                     android:textAppearance="?android:attr/textAppearanceSmall"
+                    style="?android:attr/spinnerItemStyle"
             />
 
             <TextView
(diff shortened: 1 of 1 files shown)


barbeau:
@cagryInside Thanks for taking a shot at this!  The screenshots don't look much different to me though - are you sure the "After" screenshot is correct?

cagryInside:
I opened another PR (https://github.com/OneBusAway/onebusaway-android/pull/650) that fixes this problem. Therefore I am closing this PR.

PR #633, opened 2016-07-26
Fix #618 - Add trip plan problem reporting

### LINKED ISSUE #618: Allow reporting of a trip plan that succeeds, but has problems
**Summary:** 

We should support a mechanism for reporting a trip that succeeds (i.e., an itinerary is shown to the user via list and map), but there is something wrong with the trip.  For example, if no transit leg is returned (e.g., if the GTFS expires), or it uses a pedestrian path that doesn't exist.  Currently, we only allow reporting for trip plans that fail completely and no itinerary is returned.

One possible implementation is to add a "Report trip problem" menu option in the overflow menu, accessible via the 3 dots in the upper-right corner on the Trip Planning view, under the "Reverse" option:

image

Another possible implementation is to add another entry to the general "Send feedback" menu, under "Report App Problem", labeled something like "Report a trip plan".

I think we should start with the overflow menu implementation, as it's the simplest to capture the most recently planned trip, and avoids cluttering the general "Send feedback" menu.  We can tweak this as we get a feeling for how much trip plan feedback we get.

**Steps to reproduce:** 

1. Plan a trip

**Expected behavior:** 

Provide a way to send feedback about a planned trip, including the request URL that generated the trip

**Observed behavior:** 

There is no way to report a problem with a successfully planned trip

N/A

**Screenshot:** 

See above

cagryInside:
This PR implements reporting capability when a trip plan succeeds.

Report trip plan problem button:

device-2016-07-26-110613

| Email when a trip plan succeeds  |       Email when a trip plan fails      |
|----------|:-------------:| 
|device-2016-07-26-110326 |  device-2016-07-26-110419 |

cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeServiceImpl.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeServiceImpl.java
@@ -119,7 +119,7 @@ public class RealtimeServiceImpl implements RealtimeService {
     private void checkForItineraryChange() {
         TripRequest.Callback callback = new TripRequest.Callback() {
             @Override
-            public void onTripRequestComplete(List<Itinerary> itineraries) {
+            public void onTripRequestComplete(List<Itinerary> itineraries, String url) {
                 if (itineraries == null || itineraries.isEmpty()) {
                     onTripRequestFailure(-1, null);
                     return;
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/tasks/TripRequest.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/tasks/TripRequest.java
@@ -46,7 +46,7 @@ import java.util.List;
 public class TripRequest extends AsyncTask<Request, Integer, Long> {
 
     public interface Callback {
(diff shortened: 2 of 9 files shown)


barbeau:
@cagryInside Thanks for tackling this!  I'd still like to add the request URL to the feedback email on the left, so we can easily recreate the trip the user is reporting, but just without the "Unable to plan trip" text.

Ideally we'd also change the subject line to differentiate failures from more general feedback:
* "OneBusAway Android: Trip Plan Failed Feedback"
* "OneBusAway Android: Trip Plan Feedback"

cagryInside:
> @cagryInside Thanks for tackling this! I'd still like to add the request URL to the feedback email on the left, so we can easily recreate the trip the user is reporting, but just without the "Unable to plan trip" text.
> 
> Ideally we'd also change the subject line to differentiate failures from more general feedback:
> 
> "OneBusAway Android: Trip Plan Failed Feedback"
> "OneBusAway Android: Trip Plan Feedback"

@barbeau Just pushed another commit that adds those features.

| Email when a trip plan succeeds  |       Email when a trip plan fails      |
|----------|:-------------:| 
|device-2016-07-26-161239 |  device-2016-07-26-161616 |

barbeau:

PR #643, opened 2016-07-27
Fix #579 again - Auto-center on location using default zoom

### LINKED ISSUE #579: Nearby view does not auto center on location
**Summary:** 

When I select the nearby view I expect it to show stops near my current location but instead it just shows the last location I viewed

**Steps to reproduce:** 

1. Open app
2. Select nearby
3. Pan to random location
4. Close app by removing from recent apps view
5. Open app

**Expected behavior:** 

When nearby view is opened, show nearby stops to current user location

**Observed behavior:** 

Observe map in random location when nearby view opened

Nexus 5X, OBA 2.0.14 (64)

**Screenshots:** 

N/A

barbeau:
@cagryInside Could you please give this a look and test under the various conditions where we zoom/pan to locations?

* Location is available (should snap to location at default zoom level, not previously used zoom level)
* Location isn't available (should snap to previous zoom level and lat/long)
* StopId is included in bundle - for example, tapping on map button in Action barwhen viewing a starred stop in `ArrivalsListActivity` (should snap to stop location at default zoom level)

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/map/StopMapController.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/map/StopMapController.java
@@ -183,12 +183,11 @@ public class StopMapController implements MapModeController,
     public void setState(Bundle args) {
         if (args != null) {
             Location center = UIUtils.getMapCenter(args);
-            float mapZoom = args.getFloat(MapParams.ZOOM, MapParams.DEFAULT_ZOOM);
-            mCallback.getMapView().setZoom(mapZoom);
 
             // If the STOP_ID was set in the bundle, then we should focus on that stop
             String stopId = args.getString(MapParams.STOP_ID);
             if (stopId != null && center != null) {
+                mCallback.getMapView().setZoom(MapParams.DEFAULT_ZOOM);
                 setMapCenter(center);
                 return;
             }
@@ -197,7 +196,7 @@ public class StopMapController implements MapModeController,
 
             // Try to set map based on real-time location, unless state says no
             if (!dontCenterOnLocation) {
-                boolean setLocation = mCallback.setMyLocation(false, false);
+                boolean setLocation = mCallback.setMyLocation(true, false);
                 if (setLocation) {
                     return;
                 }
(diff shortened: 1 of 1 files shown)


cagryInside:
@barbeau I found an inconsistent behavior. 

**Steps to reproduce:** 

1. Open the app when location is available.
1. Close the location and go to random location.
1. If you close the app by removing  it from recent apps view
   1. When reopen, app will focus the  previous zoom level and lat/long.
1.  If you close the app by clicking back buttom
   1. When reopen, app will focus the current location.

Tested w/ Nexus 6 v6.0.1

barbeau:
@cagryInside Interesting, nice catch.  I can reproduce this on LG G4 w/ Android 6.0 as well.

For 3), are you swiping the app away from the recents apps view, or tapping the "X" in upper right corner?

For me, swiping and "X" have same result.

And I believe in 2) you mean "Turn off system location setting..."?

barbeau:
@cagryInside and I met and talked more about this.  Repro steps (to clarify) are:

1. Open the app when location system setting (e.g., via pull down window shade) is turned on.
1. Turn off location system setting and go to random location.
1. If you close the app by removing  it from recent apps view (swiping or tapping "X" in upper-right corner)
   1. When reopen, app will focus the  previous zoom level and lat/long.
1.  If you close the app by clicking back buttom
   1. When reopen, app will focus the current location.

After some debugging, it looks like in 4) when tapping on back button, the `HomeActivity` and `BaseMapFragment` are destroyed, but the `Application` context still exists.  Therefore, when you tap on the OBA icon to re-launch the app, it creates new instances of the `HomeActivity` and `BaseMapFragment`, but is still able to pull a previously cached location reference from `Application.getLastKnownLocation()`.  Therefore, the map is still able to center on a location in `BaseMapFragment.setMyLocation()`.

In 3), the `HomeActivity` and `BaseMapFragment` AND the `Application` context are destroyed.  Therefore, this emulates starting the application from scratch without any previous state - and if location is turned off in the system settings, this means we revert to the most recent map view.

So, in summary, I think this is all correct behavior.  We want to use a recent location when it's available, which 4) does, and when location isn't available, we want to center on the most recent map view, which 3) does.  The difference in behavior is just a side-effect of how Android manages the app lifecycle.

PR #644, opened 2016-07-28
Fix #622 - Add a custom trip planning API URL setting

### LINKED ISSUE #622: Add a custom trip planning API URL setting
**Summary:** 

For other regions that have OneBusAway deployed, it would be nice if there was a way they could test an OTP server with their region, without us updating the Regions API in production.  For example, Puget Sound has an OTP server, but currently there isn't a good way of anyone in Puget Sound testing the system without going straight to their OTP URL in production or forking the app.  This would also allow regions to test new OTP server configurations (e.g., addition of real-time data or upgrades) before going to production.  

This setting is similar to the current "Custom OBA Server API" setting that we have, but for OTP.  I imagine this would also go under the "Advanced" preference category.

cagryInside:
Added a new field to set OTP API Url.

**Screenshots:**
device-2016-07-28-120207 device-2016-07-28-120253

**Behavior:**

1. Redraw navigation drawer after setting the custom url
1. When there is a custom otp url, if a user selects  a new region 
   1. If the region has otp url then custom url will be canceled.
   1. If the region hasn't otp url then custom url will remain.
 
cc'd @barbeau

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -281,6 +281,10 @@ public class Application extends android.app.Application {
     }
 
     public synchronized void setCurrentRegion(ObaRegion region) {
+        setCurrentRegion(region, true);
+    }
+
+    public synchronized void setCurrentRegion(ObaRegion region, boolean regionChanged) {
         if (region != null) {
             // First set it in preferences, then set it in OBA.
             ObaApi.getDefaultContext().setRegion(region);
@@ -288,6 +292,10 @@ public class Application extends android.app.Application {
                     .saveLong(mPrefs, getString(R.string.preference_key_region), region.getId());
             //We're using a region, so clear the custom API URL preference
             setCustomApiUrl(null);
+            if (regionChanged && region.getOtpBaseUrl() != null) {
+                setCustomOtpApiUrl(null);
+                setUseOldOtpApiUrlVersion(false);
+            }
         } else {
             //User must have just entered a custom API URL via Preferences, so clear the region info
             ObaApi.getDefaultContext().setRegion(null);
@@ -345,6 +353,48 @@ public class Application extends android.app.Application {
      */
     public void setCustomApiUrl(String url) {
         PreferenceUtils.saveString(getString(R.string.preference_key_oba_api_url), url);
+    }
+
+    /**
+     * Returns the custom OTP URL if the user has set a custom API URL manually via Preferences, or
+     * null
+     * if it has not been set
+     *
+     * @return the custom URL if the user has set a custom API URL manually via Preferences, or null
+     * if it has not been set
+     */
+    public String getCustomOtpApiUrl() {
+        SharedPreferences preferences = getPrefs();
+        return preferences.getString(getString(R.string.preference_key_otp_api_url), null);
+    }
+
+    /**
+     * Sets the custom OTP URL used to reach a OBA REST API server that is not available via the
+     * Regions
+     * REST API
+     *
+     * @param url the custom URL
+     */
+    public void setCustomOtpApiUrl(String url) {
+        PreferenceUtils.saveString(getString(R.string.preference_key_otp_api_url), url);
+    }
+
+    /**
+     * @return true if the OTP url version is old, or false  if it has not been set
+     */
+    public boolean getUseOldOtpApiUrlVersion() {
+        SharedPreferences preferences = getPrefs();
+        return preferences.getBoolean(getString(R.string.preference_key_otp_api_url_version), false);
+    }
+
+    /**
+     * Sets the OTP Api url version
+     *
+     * @param useOldOtpApiUrlVersion indicates that if otp url structure belongs to older version
+     */
+    public void setUseOldOtpApiUrlVersion(boolean useOldOtpApiUrlVersion) {
+        PreferenceUtils.saveBoolean(getString(R.string.preference_key_otp_api_url_version),
+                useOldOtpApiUrlVersion);
     }
 
     private static final String HEXES = "0123456789abcdef";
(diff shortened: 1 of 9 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
```diff
@@ -281,13 +281,20 @@ public synchronized ObaRegion getCurrentRegion() {
     }
 
     public synchronized void setCurrentRegion(ObaRegion region) {
+        setCurrentRegion(region, true);
+    }
+
+    public synchronized void setCurrentRegion(ObaRegion region, boolean didRegionChanged) {
```
barbeau: Could you change the variable name to just `regionChanged`?

onebusaway-android/src/main/java/org/onebusaway/android/directions/util/TripRequestBuilder.java
```diff
@@ -236,7 +239,23 @@ public TripRequest execute(Activity activity) {
         request.setShowIntermediateStops(true);
 
         // TripRequest will accept a null value and give a user-friendly error
-        String otpBaseUrl = Application.get().getCurrentRegion().getOtpBaseUrl();
+        String otpBaseUrl;
+        Application application = Application.get();
```
barbeau: Small tweak - could you shorten variable name to just `app`?

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -389,6 +391,15 @@ protected void onSaveInstanceState(Bundle outState) {
         }
     }
 
+//    @Override
```
barbeau: Looks like this is commented out - should it just be removed?

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -564,11 +564,16 @@
     <string name="preferences_arrival_info_style_options_a">OneBusAway classic</string>
     <string name="preferences_arrival_info_style_options_b">Cards</string>
 
+    <string name="preferences_otp_api_servername_title">Trip Planner API Server</string>
```
barbeau: Could you please change "Trip Planner" to "OpenTripPlanner" in both title and summary?

barbeau:
Thanks @cagryInside!  Looks pretty good - I haven't had a chance to test yet, but a few comments inline from looking at code.

cagryInside:
> Thanks @cagryInside! Looks pretty good - I haven't had a chance to test yet, but a few comments inline from looking at code.

@barbeau just pushed another commit that fixed inline comments.

cagryInside:
@barbeau Also added the old url version feature to this PR.

onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
```diff
@@ -259,44 +266,62 @@ private void changePreferenceSummary(String preferenceKey) {
         if (preferenceKey.equalsIgnoreCase(getString(R.string.preference_key_region))
                 || preferenceKey.equalsIgnoreCase(getString(R.string.preference_key_oba_api_url))) {
             if (Application.get().getCurrentRegion() != null) {
-                regionPref.setSummary(Application.get().getCurrentRegion().getName());
-                customApiUrlPref
+                mPreference.setSummary(Application.get().getCurrentRegion().getName());
+                mCustomApiUrlPref
                         .setSummary(getString(R.string.preferences_oba_api_servername_summary));
+                String customOtpApiUrl = Application.get().getCustomOtpApiUrl();
+                if (customOtpApiUrl == null) {
```
barbeau: Looking at the code below line 289, I think this should be `if (!TextUtils.isEmpty(customOtpApiUrl))`?

barbeau:
Thanks @cagryInside!  One comment inline.

Also, I can't seem to clear the custom OTP API after it's set - not sure if it's related to the same comment.  So if I do the following:

1. Install app, region is auto-set to Tampa
1. Set custom OTP URL to https://maps.usf.edu/otp
1. Change region to Puget Sound
1. Set custom OTP URL to an empty string (put cursor in window and backspace to remove all characters)

...I still see the "Plan a trip (beta)" option in the navigation drawer, and if I try to plan a trip, I get "Planning not possible - no region selected" error message:

image

I would expect it to clear the custom OTP API, and hide the nav. drawer option to "Plan a trip".

onebusaway-android/src/main/java/org/onebusaway/android/ui/PreferencesActivity.java
cagryInside: @barbeau do you mean line 273 instead of 289

barbeau: Yes, I'm asking about line 273, based on how line 289 looks:

Line 289:
~~~
 +            if (!TextUtils.isEmpty(otpApiUrl)) {
 +                mCustomOtpApiUrlPref.setSummary(otpApiUrl);
 +            } else {
 +                mCustomOtpApiUrlPref.setSummary(
 +                        getString(R.string.preferences_otp_api_servername_summary));
 +            }
~~~

Line 273:
~~~
+                if (customOtpApiUrl == null) {
+                    mCustomOtpApiUrlPref
 +                            .setSummary(getString(R.string.preferences_otp_api_servername_summary));
 +                } else {
 +                    mCustomOtpApiUrlPref.setSummary(customOtpApiUrl);
 +                }
~~~

Seems like these are attempting doing the same thing, but one is using a null check, and the other is using `TextUtils.isEmpty()`.  So I believe these should both be the same, unless I'm missing something.

cagryInside:
> Yes, I'm asking about line 273, based on how line 289 looks:

Yes thats correct, I pushed another commit that fixes this issue and the empty string issue.

barbeau:
Thanks @cagryInside, this looks good!  Great to both support older OTP server deployments, and to add the custom API setting for OTP servers. :+1: :+1:

PR #647, opened 2016-07-31
Fix #411 - Add markers for boat, train, subway, and tram

### LINKED ISSUE #411: Display different vehicle icons per mode
Sound Transit is now providing real-time ferry data in their OBA instance, which is awesome.

However, we currently are using the bus icon to represent all vehicles, which results in floating buses:

image

We should change the vehicle icon based on the route mode type.

Here's the API request that returns routes with ferry route types:
http://api.pugetsound.onebusaway.org/api/where/arrivals-and-departures-for-stop/95_7.json?key=TEST

~~~
agencyId: "95",
color: "",
description: "",
id: "95_73",
longName: "Seattle - Bainbridge Island",
shortName: "",
textColor: "",
type: 4,  // <-- This is the "ferry" route type
url: ""
~~~

The GTFS spec defines the following [route types](https://developers.google.com/transit/gtfs/reference?hl=en#routestxt):

>0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area.
1 - Subway, Metro. Any underground rail system within a metropolitan area.
2 - Rail. Used for intercity or long-distance travel.
3 - Bus. Used for short- and long-distance bus routes.
4 - Ferry. Used for short- and long-distance boat service.
5 - Cable car. Used for street-level cable cars where the cable runs beneath the car.
6 - Gondola, Suspended cable car. Typically used for aerial cable cars where the car is suspended from the cable.
7 - Funicular. Any rail system designed for steep inclines.

Here's where the current SVG bus vehicle icon files are:
https://github.com/OneBusAway/onebusaway-android/tree/master/icons/vehicle_markers

We can use those as a start for the marker/arrows, and then pull the other material design vehicle markers from:
https://github.com/google/material-design-icons/tree/master/maps/svg/production

...where the naming convention is `ic_directions_*` - for example, here's the boat icon for the ferry:
https://github.com/google/material-design-icons/blob/master/maps/svg/production/ic_directions_boat_48px.svg

...which looks like:

image

We can drop these icons on top of the SVG marker/arrow icons and create new mode icons.

bendu:
Icons taken from https://github.com/google/material-design-icons/tree/master/maps/svg/production

This is to start addressing #411 I was going to make the subway icon more like the tram and the train icon, but I could not figure out how to make it happen in inkscape and my knowledge of SVG is limited. If these icons are acceptable, I will make the various arrow direction icons. I already have some code ready to support displaying different vehicle icons, which I will add in a future PR.

Also, how do I generate the png files from svg?


barbeau:
Thanks @bendu!  I don't currently have access to my machine so I can't look at the SVGs now on mobile (AFAIK), but I'll take a look early next week.

IIRC I used the Android Icon Generator and just dragged the SVGs onto the page to generate the PNGs:
https://romannurik.github.io/AndroidAssetStudio/icons-generic.html#source.space.trim=1&source.space.pad=0&size=24&padding=0&color=000%2C100&name=ic_example

barbeau:
Also, I'd prefer a single PR with images and code together for final merge, although this is fine for a WIP.

barbeau:
@bendu Finally looking at this on my machine.

Ferry/boat icon looks fine:

image

Subway needs some work - here's what's currently in the PR:

image

The white/black portion should be inverted so the vehicle is white.  Also, I'd prefer using the material design subway icon (white/black needs to be inverted here too):
https://github.com/google/material-design-icons/blob/master/maps/svg/production/ic_directions_subway_48px.svg

image

It's been a while since I did the originals in Inkscape, and I don't use it regularly - I'll see if I can take a look to remember how I did the inversion there.

Train looks good:

image

Tram looks good:

image

bendu:
@barbeau I have updated the PR with code for showing different vehicles in VehicleOverlay. I also updated the subway icon. I added markers for the various directions as well. Once the icons look good. I will generate the png files. There are 36 svg files to generate, and it looks like I will have to do them 1 by 1, so I am waiting until these icons are OK'd before proceeding.

subway

bendu:
@barbeau I updated the commit with png files of the 4 markers. Here are some screenshots of a few non-bus vehicles I could find.

ferry

link

barbeau:
@bendu This looks good!  All icons look great so far.

The GTFS spec defines the following [route types](https://developers.google.com/transit/gtfs/reference?hl=en#routestxt) - here's what I believe we have covered so far (including master branch and this PR):

- [x] 0 - Tram, Streetcar, Light rail. Any light rail or street level system within a metropolitan area - *marker_with_tram_smaller_all_inside.svg*
- [x] 1 - Subway, Metro. Any underground rail system within a metropolitan area - *marker_with_subway_smaller_all_inside.svg*
- [ ] 2 - Rail. Used for intercity or long-distance travel - 
- [x] 3 - Bus. Used for short- and long-distance bus routes - *marker_with_bus_smaller_all_inside.svg*
- [x] 4 - Ferry. Used for short- and long-distance boat service - *marker_with_boat_smaller_all_inside.svg*
- [x] 5 - Cable car. Used for street-level cable cars where the cable runs beneath the car - *marker_with_tram_smaller_all_inside.svg*
- [ ] 6 - Gondola, Suspended cable car. Typically used for aerial cable cars where the car is suspended from the cable
- [ ] 7 - Funicular. Any rail system designed for steep inclines

I believe "2) Rail" could use:
https://github.com/google/material-design-icons/blob/master/maps/svg/production/ic_directions_railway_48px.svg

I believe "5) Cable car" can just use the "Tram" icon, as indicated above.

I'm fine omitting "Gondola" and "Funicular" for now :).

bendu:
@barbeau I am using marker_with_train_smaller_all_inside.svg icon for rail (2). 
I will update "cable car" to use the "tram" icon.

barbeau:
> I am using marker_with_train_smaller_all_inside.svg icon for rail (2).

Whoops!  Sorry, somehow I missed that.  Here are the two side by side for reference - train on left, railway on right:

image

I'm fine with train icon for 2) Rail - that icon looks more like what I would think of for rail.  railway icon is easier to confuse with subway as well.

bendu:
Here is a short script I used for generating the various icons with the direction arrows (written in powershell) https://gist.github.com/bendu/d87a66d4c014e0fca9fbad4a9fe94c18

barbeau:
@bendu That's awesome, thanks for sharing!   Definitely much faster than editing by hand.  Would you mind adding this to the same folder as the SVGs in the repo and add a README in markdown that describes the script and SVGs?

bendu:
Updated the code to use tram icon for cable car, added the svg script.
On Aug 20, 2016 10:31 AM, Sean Barbeau <notifications@github.com> wrote:@bendu That's awesome, thanks for sharing!   Definitely much faster than editing by hand.  Would you mind adding this to the same folder as the SVGs in the repo and add a README in markdown that describes the script and SVGs?

barbeau:
Thanks!  Is this ready for final review, or are you still working on it?  Either is fine, just don't want to drop the ball.

bendu:
Ready for final review
On Aug 21, 2016 2:33 PM, Sean Barbeau <notifications@github.com> wrote:Thanks!  Is this ready for final review, or are you still working on it?  Either is fine, just don't want to drop the ball.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -91,25 +92,30 @@
 
     private static final int NORTH = 0;
 
-    private static final int NORTH_WEST = 1;
```
barbeau: Have you tested to make sure that these changes in numeric indexes don't result in flipped icons?  For example, `2` used to be `WEST` and now it's `EAST`.  IIRC the halfWinds generated by `MathUtils` go in counter-clockwise order, which is why this was reversed.  If that's correct, then from a quick scan of the code I believe the "WEST" direction would now result in an "EAST" icon?

If this is correct, then we definitely need better comments (i.e., I should have added better comments), as this wasn't clear in the code itself.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -91,25 +92,30 @@
...
 
     private static final int NUM_DIRECTIONS = 9; // 8 directions + undirected mVehicles
 
-    private static Bitmap[] vehicle_icons = new Bitmap[NUM_DIRECTIONS];  // Black vehicle icons
+    private static final int DEFAULT_VEHICLE_TYPE = ObaRoute.TYPE_BUS; // fall back on bus
+
+    // Vehicle type (if available) -> icon set
+    private static Map<Integer, Bitmap[]> vehicle_type_icons = new HashMap<>(ObaRoute.NUM_TYPES);
+
+//    private static Bitmap[] vehicle_icons = new Bitmap[NUM_DIRECTIONS];  // Black vehicle icons
```
barbeau: Looks like this is left over from the previous implementation, you can just remove it.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
```diff
@@ -182,21 +188,11 @@ public synchronized void clear() {
...
-            vehicle_icons[1] = createVehicleIcon(NORTH_EAST);
-            vehicle_icons[2] = createVehicleIcon(EAST);
-            vehicle_icons[3] = createVehicleIcon(SOUTH_EAST);
-            vehicle_icons[4] = createVehicleIcon(SOUTH);
-            vehicle_icons[5] = createVehicleIcon(SOUTH_WEST);
-            vehicle_icons[6] = createVehicleIcon(WEST);
-            vehicle_icons[7] = createVehicleIcon(NORTH_WEST);
-            vehicle_icons[8] = createVehicleIcon(NO_DIRECTION);
-        }
+        vehicle_type_icons.put(ObaRoute.TYPE_BUS, createBusIcons());
```
barbeau: IIRC because `vehicle_type_icons` is static, it can end up being initialized multiple times, each time `VehicleOverlay` is instantiated.  I had the check for `vehicle_icons[0] == null` in the previous code to make sure that we didn't pay this initialization penalty more than once.  This is even more important now that we're loading more vehicle icon types.  You can add a similar IF block around the initialization of `vehicle_type_icons`, something like `if (!vehicle_type_icons.contains(ObaRoute.TYPE_BUS)) {`, to do this same type of check.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
bendu: IIRC the const direction vars were only used to index into the vehicle_icons array. I redefined the numeric indexes to be consistent with the halfWinds generated. I will double check when I get the chance.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
bendu: will do

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
bendu: good point.

onebusaway-android/src/amazon/java/org/onebusaway/android/map/googlemapsv2/VehicleOverlay.java
barbeau: I'll test more myself too.  I loaded it up and at a glance icons looked correct based on route directions.

bendu: I tested the app and the icons are pointing the right direction. I looked over the diff and I found that the halfwinds do go clockwise. The constants were used in the former method createvehicleicon to retrieve the icon for a particular direction. I now use the constants to determine which index the icon should go, which was not the case before.

bendu:
@barbeau I've address your code review comments and updated the diff.

barbeau:
Thanks @bendu, this looks great!  No more floating buses :+1:  :)

barbeau:
We may want to further optimize icon loading (it's doubtful that in any one execution all 5 vehicle types will be displayed), but that can happen in another issue.

barbeau:
Opened a new issue for optimizing black template icon loading at https://github.com/OneBusAway/onebusaway-android/issues/687.

PR #654, opened 2016-08-08
Fix #653 - Pass in Class instead of Activity reference for trip plan notification target

### LINKED ISSUE #653: Trip plan activity is potentially referenced outside its lifecycle
**Summary:** 

I found this when fixing https://github.com/OneBusAway/onebusaway-android/issues/652.

When a user plans a trip, we poll the server for a period to check if there are any real-time changes that might impact the trip (see https://github.com/OneBusAway/onebusaway-android/issues/617#issuecomment-234952817 for details).  If there are changes, we notify the user.

In `RealtimeServiceImpl.showNotification()`, `mActivity` is referenced to get the Class to register with the notification, so when the user taps on the notification they are brought to the trip planning activity.

The issue here is that `showNotification()` is called from the `BroadcastReceiver.onReceive()` within `RealtimeServiceImpl`, which continues to poll the server after a trip is planned.  So, the activity may have been killed by the time the `BroadcastReceiver` attempts to notify the user.  This would then result in an NPE when trying to access `mActivity`.

There is no other use of `mActivity` in `RealtimeServiceImpl`, so I believe an easy fix for this is to pass in the `Class` instead of the `Activity` in the `RealtimeServiceImpl` constructor.  I'll put together a PR for this.

cc @sdjacobs 

**Steps to reproduce:** 

I haven't actually produced an NPE, but looking at the code this should do it:

1.  Plan a trip
1.  Kill the TripPlanActivity
1.  Have the BroadcastReceiver generate a notification for a trip that's no longer recommended

**Expected behavior:** 

Not crash

**Observed behavior:** 

It would crash.

N/A.

barbeau:
@sdjacobs Could you please take a look at this?

Goal is to avoid referencing the live Activity outside of it's lifecycle, after it's been killed.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeServiceImpl.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeServiceImpl.java
@@ -53,7 +53,7 @@ public class RealtimeServiceImpl implements RealtimeService {
 
     private AlarmManager mAlarmMgr;
 
-    private Activity mActivity;
+    private Class<? extends Activity> mNotificationTarget;
 
     private IntentFilter mIntentFilter;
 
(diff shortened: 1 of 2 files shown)


sdjacobs:
Looks good. See issue [comment](https://github.com/OneBusAway/onebusaway-android/issues/653#issuecomment-238552482)

barbeau:
Thanks @sdjacobs!  I'll comment on the other issue.

PR #657, opened 2016-08-10
Fix #656 - Add GA event for trip plan requests sent to a server

### LINKED ISSUE #656: Add GA event for trip plan requests sent to a server
**Summary:** 

We should add a Google Analytics event that fires when a trip plan request is sent to an OTP server.  The "Clicked Trip Plan link" event currently fires when a user taps on the "Plan a trip" icon in the main navigation drawer, but we'd also like to capture how many times a trip is planned once they are looking at the `TripPlanActivity`.

**Steps to reproduce:** 

1. Plan a trip

**Expected behavior:** 

For each trip request sent to the server, a GA event should be recorded

**Observed behavior:** 

No GA event is recorded for trip requests sent to the server.

N/A

cagryInside:
Added a GA event for trip plan requests.

| submit  | trip_plan  | Planned a Trip |

cc'd @barbeau


barbeau:

PR #660, opened 2016-08-10
Fix #627 - Add stop direction with stop name

### LINKED ISSUE #627: Re-add stop direction with stop name
**Summary:** 

I've gotten a few requests from users to add the stop direction back with the stop name.  There are apparently some areas in Puget Sound where stops on either side of the street have the exact same names, and stop direction is important to tell them apart.  Arrows on the map apparently aren't sufficient for some users.

v1.x included the direction for the stop in the header:

image

Here's how it currently looks:

image

cagryInside:
Added `stop_direction`.

Screenshot from Nexus 6:

device-2016-08-10-110819

Screenshots from HTC EVO 3d:

device-2016-08-10-111523 device-2016-08-10-111502

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListHeader.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListHeader.java
@@ -146,6 +146,8 @@ class ArrivalsListHeader {
 
     private TextView mNameView;
 
+    private TextView mDirectionView;
+
     private EditText mEditNameView;
 
     private ImageButton mStopFavorite;
@@ -315,6 +317,7 @@ class ArrivalsListHeader {
         mNameContainerView = mView.findViewById(R.id.stop_name_and_info_container);
         mEditNameContainerView = mView.findViewById(R.id.edit_name_container);
         mNameView = (TextView) mView.findViewById(R.id.stop_name);
+        mDirectionView = (TextView) mView.findViewById(R.id.stop_direction);
         mEditNameView = (EditText) mView.findViewById(R.id.edit_name);
         mStopFavorite = (ImageButton) mView.findViewById(R.id.stop_favorite);
         mStopFavorite.setColorFilter(mView.getResources().getColor(R.color.header_text_color));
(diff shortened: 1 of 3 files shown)


barbeau:
Thanks @cagryInside!  This looks good, with the exception that `mController.getStopDirection()` can actually return an empty string for stops that don't have directional information.  These stops without directions are usually transit centers that have routes arriving from many different directions - an example in Tampa is Stop # 6497, UATC:

image

So, the new TextView should be cleared or visibility set to GONE if the stop doesn't have direction text.  In the current implementation, you can see above that for stops without direction like UATC it just keeps the direction from the stop that was tapped on prior to this stop (in this case, `(S)`).  Correct behavior would be for the direction information to be cleared, so only the stop name `University Area Transit Center` would be visible.

cagryInside:
> Thanks @cagryInside! This looks good, with the exception that mController.getStopDirection() can actually return an empty string for stops that don't have directional information. These stops without directions are usually transit centers that have routes arriving from many different directions - an example in Tampa is Stop # 6497, UATC:

Good catch! Just fixed.

barbeau:

barbeau:
After staring at this a bit longer, I changed the style a bit in https://github.com/OneBusAway/onebusaway-android/commit/c6d101eacbcff40bec89789dec8397e6c6afb540: 

image

I changed to 14sp text size and transparent white, which is the same size and color as the trip direction and "min" in header, and also similar to 14dp size used for favorite star to left of stop name.  Direction is still centered vertically so middle is center vertically aligned with the favorite star and stop name.

The previous style that was identical to the stop name seemed to draw attention away from the stop name itself, and was also confusing when mixed with stop names like "4th Ave N + 15th St. N" - it ended up looking like:

>"4th Ave N + 15th St. N (E)"

...and you couldn't easily tell the direction from the stop name.  I think the updated style offsets it a bit and makes it easier to visually parse.

Here's the before (left) and after (right) for this example:

image

@cagryInside Any thoughts?

cagryInside:
@barbeau I agree, I think it looks much better with smaller size. Also, like you said it is now easier to distinguish between stop name and the direction.

PR #678, opened 2016-08-22
Fix #655 - change RealtimeService to use WakefulBroadcastReceiver and IntentService

### LINKED ISSUE #655: User isn't notified of trip plan changes when app is backgrounded
**Summary:** 

Originally discussed in https://github.com/OneBusAway/onebusaway-android/issues/653#issuecomment-238552482.

When a user plans a trip, we poll the server for a period to check if there are any real-time changes that might impact the trip (see https://github.com/OneBusAway/onebusaway-android/issues/617#issuecomment-234952817 for details).  If there are changes, we notify the user.

However, currently these notifications only occur if `TripResultsFragment` is in the foreground.  From https://github.com/OneBusAway/onebusaway-android/issues/653#issuecomment-238606127:

> It looks like all background updates are currently killed when the user leaves the `TripResultsFragment`, in `TripResultsFragment.onDestroy()` via `mRealtimeService.disableListenForTripUpdates();`.

HART wanted this reminder to be available after the user planned the trip and changed to a different screen, so they could be notified if a delay or other issue caused problems at any point after planning a trip (e.g., en-route).

We will need to be sure that the new implementation properly stops notifications to the user after a certain threshold of time or when the user dismisses the notification.

@sdjacobs Could you please take a look at this?  

Most of this code may already be in place - I haven't looked closely.  The currently implementation may work but leak the `TripResultsFragment` after its lifecycle ends, similar to the issue I discussed with `TripPlanActivity` in https://github.com/OneBusAway/onebusaway-android/issues/653#issuecomment-238606127.  I think a good approach that probably wouldn't require much modification is to register an `IntentService` to receive the alarm from the AlarmManager, instead of using the locally declared BroadcastReceiver.  This would allow the code that executes the request to the OTP server and creates the notification to run in it's own service just for the duration that's required, and then it shuts down.  Here's an example of registering an `IntentService` with the `AlarmManager`:
http://stackoverflow.com/questions/22394615/intentservice-wont-start-using-alarmmanager

**Steps to reproduce:** 
1. Change the code so that a notification is produced every time OTP is requeried
2. Put debugging breakpoints in `TripResultsFragment.onDestroy()` and `TripPlanActivity.onDestroy()`, and `RealtimeServiceImpl.mBroadcastReceiver.onReceive()`.
3. Planning a trip 
4. Kill TripPlanActivity by pressing back

**Expected behavior:** 

Notifications should continue to occur (e.g., `RealtimeServiceImpl.mBroadcastReceiver.onReceive()` should still be called) following `TripResultsFragment.onDestroy()` and `TripPlanActivity.onDestroy()`.

**Observed behavior:** 

No notifications occur after `TripResultsFragment.onDestroy()` and `TripPlanActivity.onDestroy()` (e.g., `RealtimeServiceImpl.mBroadcastReceiver.onReceive()` is never called).  This is because all background updates are currently killed when the user leaves the `TripResultsFragment`, in `TripResultsFragment.onDestroy()` via `mRealtimeService.disableListenForTripUpdates();`

Stock LG G4 w/ Android 6.0

sdjacobs:
Updated RealtimeService (renamed from RealtimeServiceImpl) as per comments in #655.


onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeService.java
```diff
@@ -13,13 +14,258 @@
...
+
+        RealtimeWakefulReceiver.completeWakefulIntent(intent);
+    }
+
+    // Depending on preferences / whether there is realtime info, start updates.
+    private void startRealtimeUpdates(Bundle params, Itinerary itinerary) {
+
+        Log.d(TAG, "Checking whether to start realtime updates.");
+
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
```
barbeau: I'd prefer to use our singleton method to get the preferences, as used elsewhere in the app:
`SharedPreferences prefs = Application.getPrefs();`

sdjacobs:
I made another change - moved up the Preferences check so nothing is ever scheduled if notifications are disabled. Unless you have more comments I can go ahead and squash.

barbeau:
Sorry, I started reviewing this yesterday and got pulled into something else - you can leave as is for now and I'll try to finish reviewing today.

barbeau:
Thanks @sdjacobs!  Sorry this took so long to get merged.  I just manually merged to resolve conflicts via https://github.com/OneBusAway/onebusaway-android/commit/7dc279332e0dfc7638233acb1fc50b6f4f3016f9, so I'm going to close this out.

PR #716, opened 2016-11-10
Fix #713 - Allow user to hide alerts

### LINKED ISSUE #703: NPE when refreshing arrival data during fragment transition
**Summary:** 

Using current master branch (https://github.com/OneBusAway/onebusaway-android/commit/27979debe534a77731a551cb2bd79326362a5841), I had the app open on my phone on my desk to the "Nearby" screen with a stop selected (Tampa, stop_id 6497).  I starred the stop, and then tapped on the menu and tapped on "Starred Stops", and the app crashed with this stack trace:

```
E/AndroidRuntime: FATAL EXCEPTION: main
   Process: com.joulespersecond.seattlebusbot, PID: 22701
   java.lang.NullPointerException: Attempt to invoke virtual method 'int org.onebusaway.android.ui.ArrivalsListLoader.getMinutesAfter()' on a null object reference
       at org.onebusaway.android.ui.ArrivalsListFragment.setResponseData(ArrivalsListFragment.java:545)
       at org.onebusaway.android.ui.ArrivalsListFragment.onLoadFinished(ArrivalsListFragment.java:459)
       at org.onebusaway.android.ui.ArrivalsListFragment.onLoadFinished(ArrivalsListFragment.java:89)
       at android.support.v4.app.LoaderManagerImpl$LoaderInfo.callOnLoadFinished(LoaderManager.java:476)
       at android.support.v4.app.LoaderManagerImpl$LoaderInfo.reportStart(LoaderManager.java:315)
       at android.support.v4.app.LoaderManagerImpl.doReportStart(LoaderManager.java:819)
       at android.support.v4.app.Fragment.performStart(Fragment.java:2009)
       at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1102)
       at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1252)
       at android.support.v4.app.BackStackRecord.run(BackStackRecord.java:742)
       at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1617)
       at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:517)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5551)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:730)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:620)
```

**Steps to reproduce:** 

Appears to be related to switching fragments in HomeActivity right when the arrival time is refreshed in ArrivalsListFragment (steps under which it occurred are above)

**Expected behavior:** 

Not crash

**Observed behavior:** 

Crash

LG G5 w/ Android 6.0.1 w/ Google Play Services 9.6.83

### LINKED ISSUE #713: Allow users to dismiss service alerts
**Summary:** 

We're beginning to see regions take more advantage of the ability to distribute service alerts via OneBusAway, which is great.  In the past, alerts were usually temporary/transient events such as a real-time service outage or maintenance alert.  

Now, agencies are starting to publish more long running alerts about stops or routes with detours that are weeks long, or to advertise new services.  When these alerts aren't carefully targeted to specific routes or stops, they can quickly add up and take up a large portion of the screen.

For example, here's what the alerts look like for Tampa for every stop as of today:

image

Users are starting to complain:

> I do not want alerts signified by the yellow triangle with an exclamation mark. These alerts are not pertinent for what I need.

WMATA accidentally published around 6 empty service alerts (no text) last week and I also got several complaints about that.  Same with San Diego, which has 2-3 alerts for a decent number of stops:

image

We need a mechanic that allows users to selectively acknowledge that they've read an alert, and dismiss it so it doesn't appear again.  I think a simple "DISMISS" button in the bottom right would do it, similar to how the calendar app looks:

image

We should probably add a button somewhere (in Settings?) that would allow a user to restore (i.e., undismiss) all active alerts, so they would appear again.

We already have a service alerts table in the local database, although it appears to currently be unused.  There is a `markAsRead` boolean variable when inserting/updating records, which records the read time (current time) to the database.  We should be able to use this table for this feature.

Related to https://github.com/OneBusAway/onebusaway-android/issues/447, although https://github.com/OneBusAway/onebusaway-android/issues/447 is technically considered a bug in how we display a large number of alerts in that they can starve arrivals for space - this issue is for a new mechanic to allow selectively dismissing alerts so they don't take up space.

**Steps to reproduce:** 
1. Go to Tampa region and tap on stop, and slide up panel, or go to starred stop

**Expected behavior:** 

Allow me to get rid of service alerts blocking my view of arrivals

**Observed behavior:** 

I'm stuck looking at alerts from the agency for however long the agency decides to put them there.

N/A

barbeau:
- [x] Service alert header disappeared - re-add it
- [x] Add option to re-enable all active alerts - let's do this as a new gray list item where the active alerts are normally shown
- [x] Add snack bar for undo
- [x] Move snackbar out of Service Alert activity - it takes too long for that activity to finish itself when a snackbar is shown there.
- [x] Tweak formatting and layouts

* Also fixes #703, which happens more often now that we're refreshing the list when the user taps on "Show all alerts"


barbeau:
Here are screenshots for the final UI flow:

image

PR #725, opened 2016-11-29
WIP - Fix #704 - Initial work towards fixing crash with forward slash in ID

### LINKED ISSUE #704: Crash when tapping on stop that contains a forward slash "/" in stopId
**Summary:** 

If you tap on a stop that has a `/` in the `stopId`, the app crashes with:

`java.lang.IllegalArgumentException: Unknown URI: content://com.joulespersecond.oba/stops/1_Forge Park / 495`

...once the ArrivalsListFragment calls setUserInfo() and the URI is queried in OBAProvider.queryInternal().  This particular URI falls through the switch statement which filters query types and generates an IllegalArgumentException.

Seems like client-side we should probably escape `/` as `%2F` - see http://stackoverflow.com/questions/2992231/slashes-in-url-variables. 

Just opened server issue here as well - it's not clear that even if we escape the `/` client-side that servers can accept this in a REST API paramater :
https://github.com/OneBusAway/onebusaway-application-modules/issues/182

**Steps to reproduce:** 

Using the Boston OBA server, tap on a stop for the following stopId:
`1_Forge Park / 495`

It can be seen in the list of stops returned in this stops-for-location query:
http://developer.onebusaway.org/mbta-api/api/where/stops-for-location.json?lat=42.078482127189595&lon=-71.43382396548986&latSpan=0.08702002322407765&lonSpan=0.07993753999471664&version=2&key=TEST

**Expected behavior:** 

Not crash and show arrival times for that stop

**Observed behavior:** 

Crash

N/A

barbeau:
TODO:
* Find an elegant way to remove the slash when forming the stop/route Uri
* Keep the String replacement on the ObaRoute/StopElement.getId()?

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/util/IOUtils.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/util/IOUtils.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2016 University of South Florida (sjbarbeau@gmail.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.util;
+
+/**
+ * Utilities for input and output
+ */
+
+public class IOUtils {
+
+    /**
+     * Encodes the ID to ensure it's safe for local storage and submission as API request
+     * parameter.
+     * See #704.
+     *
+     * @param id the element ID
+     * @return an encoded version of that element ID
+     */
+    public static String encodeId(String id) {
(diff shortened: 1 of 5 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/util/IOUtils.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Encodes the ID to ensure it's safe for local storage and submission as API request
+     * parameter.
+     * See #704.
+     *
+     * @param id the element ID
+     * @return an encoded version of that element ID
+     */
+    public static String encodeId(String id) {
+        return id.replace("/", "%2F");
```
azizmb9494: Would / be the only char we want to URL encode?

barbeau: For this, yes.  The problem is the slash messing with the URI used internally when the ID is persisted to the provider.  Everything else is URL-encoded when the API request gets constructed.

barbeau: Also, don't bother sinking time into this if you're looking for something to tackle. In https://github.com/OneBusAway/onebusaway-application-modules/issues/182 we decided that this should really be preprocessed server side to fix this in all clients. This is such a rare issue that I'd prefer not to touch the URL code that affects every persisted element and has the potential to break a lot of stuff.

aaronbrethorst:
closing due to age and conflicts.

PR #726, opened 2016-12-01
Fix #717 - Support transit-only Open311 deployments

### LINKED ISSUE #717: Support transit-only Open311 deployments
**Summary:** 

In Tampa Bay, we originally tested the new Open311-based issue reporting integration with PSTA in Pinellas County.  In that setup, there are two transit-related categories (stops and arrival times) - all other categories aren't transit-related, because PSTA is only one of many government agencies using the [Pinellas SeeClickFix account](http://seeclickfix.com/pinellas_county/report):

image

Here's the breakdown of categories for the two transit-related categories in Pinellas County:

>PSTA - Bus Stops: 
•	Trash at Bus Stop
•	Vandalism/Graffiti
•	Benches
•	Shelters
•	Safety concern
•	Incorrect/Missing Stop ID
•	Location is wrong
•	Route/trip is missing
•	other-call 727 540-1800

>PSTA - Arrival times/schedules:
•	Bus never came
•	Bus came earlier than predicted
•	Bus came later than predicted
•	Wrong destination shown
•	Bus doesn’t stop here
•	Other- call (727) 540-1800

These categories intentionally closely resemble the original stop/trip problem reporting categories used in the original OBA issue reporting API.

HART recently purchased a SeeClickFix account for their own use.  This differs from the PSTA setup, because HART is the account owner, and therefore *all* issue categories are transit related.  Right now HART wanted to flatten out the issue categories, so root categories look like this:

>•	Trash at Bus Stop
•	Vandalism/Graffiti
•	Benches
•	Shelters
•	Safety concern
•	Incorrect/Missing Stop ID
•	Location is wrong
•	Route/trip is missing
•	Bus never came
•	Bus came earlier than predicted
•	Bus came later than predicted
•	Wrong destination shown
•	Bus doesn’t stop here

The reason behind this is that in SeeClickFix, notifications to agency staff when a user reports a problem can only be controlled at the root category level - in other words, PSTA can control who gets notified for bus stop issues and arrival time issues, but can't notify a separate group for trash at a bus stop vs. a safety concern at a bus stop.  HART wants better control over the notifications based on stop sub-categories, which was the catalyst behind flattening out the results (an aside - HART may be ok with grouping the arrival time/schedule issues under one category).

We could easily support sorting these into proper groups if SeeClickFix supported the `group` and `keyword` elements of the Open311 response - however, they do not, and it's not on their immediate roadmap.

OBA Android doesn't currently know how to handle all transit-related categories, and it tries to match the two "transit-related" categories based on the category names, which ends up looking like this:

image

So, two of the transit categories fall under the "Transit" header, and the rest of the transit categories incorrectly fall under the "Other" category.

I think the best solution to fix this is the following:

1. If there are more than 2 matches to transit-related text in the root categories, then we should remove the "Transit" and "Others" headers and just show all categories with no header.  And, we would assume/require that all categories have the "Bus stop id" as a field to automaticaly fill.
1. HART should move all the "arrival time/schedule" categories under a single root category, the same way PSTA does - This allows us to detect this category as "arrival time" related using text matching and make the user pick an arrival time in the issue reporting flow the same way we do with PSTA.

One question is which stop-related category do we default to if the user picks "Report Stop problem" in the UI flow - we no longer have a single "Bus Stop" category to direct the user too.  I suppose we could just default to no category being selected.

@cagryInside Do you have any thoughts on this?  Any pointers as for where I should get started working on this (as you can tell, I haven't started digging into code yet ;) )

**Steps to reproduce:** 

1. Change string `regions_api_url` in the file `donottranslate.xml to `https://raw.githubusercontent.com/OneBusAway/onebusaway-multiregion-support/testSeeClickFix/staging/regions-v3.json` to make OBA Android use the staging Regions API file that includes the demo SeeClickFix server that has the HART root categories
1. Tap on "Send feedback->Report Stop Problem"
1. Choose a stop in Hillsborough County (e.g., near USF)
1. Tap on the category list to see list of categories

**Expected behavior:** 

See all transit-related categories in a single list, without any category headers.  And, "Bus stop id" should be automatically populated.

**Observed behavior:** 

"Transit" category includings "Trash at Bus Stop" and "Route/trip is missing", and all other transit-related categories incorrectly appear under "Others" category.  The "Bus stop id" field is not automatically populated:

image

LG G5 w/ Android 6.0.1

**Screenshots:** 

See above

barbeau:
@cagryInside Would you have a moment to review this PR before Monday?  If not, no problem, but I wanted to get your eyes on this if possible.  This seems hacky, and there might be a better way to do this that I'm not seeing yet.  However, we are doing text-based heuristic matching, which is hacky anyway.

I moved `createTransitServices()` to the utility class to reduce the amount of code in InfrastructureIssueActivity.  I also added a test to make sure our text heuristic matching is working as intended.

I think this works as-is (i.e., no more code changes are needed), although I need to do more testing to make sure.

### DIFF
--- a/onebusaway-android/src/main/res/values/donottranslate.xml
+++ b/onebusaway-android/src/main/res/values/donottranslate.xml
@@ -45,7 +45,9 @@
     <string name="preference_key_trip_plan_notifications">preference_key_trip_plan_notifications</string>
 
     <!-- Regions API URL -->
-    <string name="regions_api_url">http://regions.onebusaway.org/regions-v3.json</string>
+    <string name="regions_api_url">
+        https://github.com/OneBusAway/onebusaway-multiregion-support/raw/testSeeClickFix/staging/regions-v3.json
+    </string>
 
     <!-- Donate URL -->
     <string name="donate_url">http://onebusaway.org/donate/</string>
--- a/onebusaway-android/src/main/res/values/donottranslate.xml
+++ b/onebusaway-android/src/main/res/values/donottranslate.xml
@@ -45,9 +45,7 @@
     <string name="preference_key_trip_plan_notifications">preference_key_trip_plan_notifications</string>
 
     <!-- Regions API URL -->
-    <string name="regions_api_url">
-        https://github.com/OneBusAway/onebusaway-multiregion-support/raw/testSeeClickFix/staging/regions-v3.json
-    </string>
+    <string name="regions_api_url">http://regions.onebusaway.org/regions-v3.json</string>
 
     <!-- Donate URL -->
     <string name="donate_url">http://onebusaway.org/donate/</string>
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemOpen311Test.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemOpen311Test.java
@@ -38,7 +38,13 @@ import edu.usf.cutr.open311client.models.ServiceListRequest;
 import edu.usf.cutr.open311client.models.ServiceListResponse;
 
 /**
- * Tests to evaluate interactions with Open311 system for regions that use Open311
+ * Tests to evaluate interactions with Open311 system for regions that use Open311.
+ *
+ * NOTE: These tests make actual HTTP requests to live Open311 servers (SeeClickFix), because we
+ * make some assumptions required for text heuristic matching to identify transit-related services.
+ * If there are changes in the live server (e.g., new services added than aren't transit-related
+ * but include some of the text we're using to identify transit services), these could break our
+ * assumptions, and we want to know about that.
  */
 public class ReportProblemOpen311Test extends AndroidTestCase {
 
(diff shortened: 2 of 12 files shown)


cagryInside:
Sure, @barbeau I will try to look at it before Monday.

onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -45,7 +45,9 @@
     <string name="preference_key_trip_plan_notifications">preference_key_trip_plan_notifications</string>
 
     <!-- Regions API URL -->
-    <string name="regions_api_url">http://regions.onebusaway.org/regions-v3.json</string>
+    <string name="regions_api_url">
+        https://github.com/OneBusAway/onebusaway-multiregion-support/raw/testSeeClickFix/staging/regions-v3.json
+    </string>
```
cagryInside: @barbeau The test api url is pushed. I am not sure if this is done by intentionally or accidentally.

cagryInside:
@barbeau I think this will work as-is. Unfortunately, the methodology that we have been applying to find out transit services is hacky. IMO, If SCF implements the `group` section into the services response, the solution will be much cleaner. 
Overall, it looks good (I also tested the screen rotation). I just have one simple in-line comment.

barbeau:
Thanks for the review @cagryInside!

PR #751, opened 2017-02-06
Fix #749 - Use primitives instead of Serializables in Bundles for AlarmManager

### LINKED ISSUE #749: NPE when trying to get updated trip plan in the background
**Summary:** 

For current `master` branch and v2.1.6 (73) - crash after planning a trip, when the app attempts to check for changes in the itinerary:

~~~
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object java.util.ArrayList.get(int)' on a null object reference
	at org.onebusaway.android.directions.realtime.RealtimeService.getItinerary(RealtimeService.java:276)
	at org.onebusaway.android.directions.realtime.RealtimeService.onHandleIntent(RealtimeService.java:76)
	at android.app.IntentService$ServiceHandler.handleMessage(IntentService.java:67)
	at android.os.Looper.loop(Looper.java:154)
	at android.os.HandlerThread.run(HandlerThread.java:61)
~~~

Very likely caused by https://github.com/OneBusAway/onebusaway-android/pull/678, which was manually merged due to conflicts in https://github.com/OneBusAway/onebusaway-android/commit/7dc279332e0dfc7638233acb1fc50b6f4f3016f9.

I'm not sure this is consistently reproducible for all trips.  I produced it back to-back on the same trip plan, but later couldn't repro it, although I couldn't find a trip with real-time info.

@sdjacobs Would you be able to take a look at this?  I'm halting the public rollout of v2.1.6 until we can figure this out so everyone doesn't encounter this during trip planning in Tampa.  

**Steps to reproduce:** 

1. Plan a trip with real-time info, and wait for the update on the same screen

**Expected behavior:** 

Not crash

**Observed behavior:** 

NPE and crash

LG G5 with Android 7.0

sdjacobs:
It appears that Android 7.0 does not allow passing Serializables in Bundles that use AlarmManager. Couldn't find this in the documentation, but see Stack Overflow questions [here](http://stackoverflow.com/questions/38775285/android-7-broadcastreceiver-onreceive-intent-getextras-missing-data) and [here](http://stackoverflow.com/questions/38466080/dp5-7-0-does-adding-extras-to-a-pending-intent-fail).

I made some changes so that data serialization to a bundle in RealtimeService will only use primitives - in particular the restriction is to types allowed by [PersistableBundle](https://developer.android.com/reference/android/os/PersistableBundle.html). (This is the bundle type used by the [JobScheduler API](https://developer.android.com/reference/android/app/job/JobScheduler.html)).

Fix #749

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/model/ItineraryDescription.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/model/ItineraryDescription.java
@@ -49,6 +49,11 @@ public class ItineraryDescription {
         mEndDate = ConversionUtils.parseOtpDate(last.endTime);
     }
 
+    public ItineraryDescription(List<String> tripIds, Date endDate) {
+        mTripIds = tripIds;
+        mEndDate = endDate;
+    }
+
     /**
      * Check if this itinerary matches the itinerary of another ItineraryDescription
      *
@@ -104,5 +109,12 @@ public class ItineraryDescription {
      */
     public boolean isExpired() {
         return getEndDate().before(new Date());
+    }
+
+    /**
+     * return list of trip IDs
+     */
+    public List<String> getTripIds() {
+        return mTripIds;
     }
 }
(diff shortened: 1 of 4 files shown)


barbeau:
Thanks @sdjacobs!  Interesting, I didn't know that.  For the record here's Google's official response stating this via the AOSP issue tracker:
https://code.google.com/p/android/issues/detail?id=216581#c6

PR #753, opened 2017-02-07
Fix #752 - Ensure correct itinerary is selected when app is backgrounded and resumed

### LINKED ISSUE #752: Selected itinerary does not persist on app resume
**Summary:** 

When the app is backgrounded and resumed, we lose the itinerary that was selected.

**Steps to reproduce:** 

1) Plan a trip.
2) Select itinerary 2 or 3.
3) Press the home button. 
4) Press the overview button (rightmost button) and select OBA.

Note the switch back to the first itinerary.

**Expected behavior:** 

Go back to the itinerary that was previously selected.

**Observed behavior:** 

The itinerary shown is the first one.

Android 7.0 on the emulator

sdjacobs:
PR for #752


barbeau:
Awesome, thanks @sdjacobs!

PR #799, opened 2017-09-05
Fix #794 - Add "transit only" mode for trip planning

### LINKED ISSUE #794: Trip planning (bikeshare beta) - add "transit only" mode
**Summary:** 

Reported by user for current 2.2.4 bikeshare beta:

>Currently in OBA android, when you want to plan a trip (using beta OTP client), you can notably choose between bikeshare or transit+bikeshare but can't choose transit only (transit+walk), whereas most of the time you need a separate subscription or ticket to be able to use bikeshare, even when you have unlimited access to transit system.
>
>Would it be possible to implement transit only option?
>
>Also, having a way to display current disruptions as in default OTP client would also be nice, at it can have a main impact on transit.

I agree, we should have a transit-only option (which includes bus and rail).

**Steps to reproduce:** 

Plan a trip in an area with bikeshare

**Expected behavior:** 

Allow me to get trips only for transit, without any bikeshare legs

**Observed behavior:** 

There is no way to eliminate bikeshare from a planned trip

N/A

@carvalhorr Would you be able to take a look at this?  I'd like to get this into the production release of the app.

carvalhorr:

- [ ] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/util/TripRequestBuilder.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/util/TripRequestBuilder.java
@@ -60,10 +60,11 @@ public class TripRequestBuilder {
     private static final String MODE_SET = ".MODE_SET";
     private static final String DATE_TIME = ".DATE_TIME";
 
-    private static final int TRANSIT_MODE = 0;
+    private static final int TRANSIT_AND_BIKE_MODE = 0;
     private static final int BUS_ONLY_MODE = 1;
     private static final int RAIL_ONLY_MODE = 2;
     private static final int BIKESHARE_MODE = 3;
+    private static final int TRANSIT_ONLY_MODE = 4;
 
 
     private TripRequest.Callback mListener;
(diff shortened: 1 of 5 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/directions/util/TripRequestBuilder.java
```diff
@@ -196,7 +199,7 @@ public TripRequestBuilder setModeSetById(int id) {
     public int getModeSetId() {
         // IF bike mode is selected in the trip plan additional preferences but bikeshare is not enabled use the default mode (TRANSTI)
         if (BIKESHARE_MODE == mModeId && !Application.isBikeshareEnabled()) {
-            return TRANSIT_MODE;
+            return TRANSIT_AND_BIKE_MODE;
```
barbeau: @carvalhorr It seems here either the comment is wrong or this should be `TRANSIT_MODE`.  Could you please resolve the two so they match?

onebusaway-android/src/main/java/org/onebusaway/android/ui/TripModes.java
```diff
@@ -0,0 +1,61 @@
+package org.onebusaway.android.ui;
+
+import org.onebusaway.android.R;
+import org.onebusaway.android.app.Application;
+
+/**
```
barbeau: Could you please add a short description to this class and remove the "Created by...".  Also please add Apache license header.

barbeau:
Thanks @carvalhorr!  Looks good!

PR #801, opened 2017-09-18
Fix #609 - Add custom reminder sound and vibration

### LINKED ISSUE #609: Allow changes to "My reminders" notification sound, volume, and vibration
**Summary:** 

Feedback from a user:

> Ideally, I would like to set a custom alert sound and volume. I would expect these options to be either in Settings (best) or at top right menu on set notification page.

I believe we currently just use the default system setting for generic notifications.

We should also allow activation/disabling of vibration - IIRC we have a permission for vibration in the manifest, but the code is currently commented out (not sure why?).

**Steps to reproduce:** 
1. Tap on "My reminders" option in navigation panel
2. Tap on a reminder

**Expected behavior:** 

Allow me to change the reminder notification settings

**Observed behavior:** 

You can't change the reminder notification settings

N/A

bridgette:
Per user request, add settings to toggle vibration and customize notification sounds. This PR also removes old commented-out code that demo'd the (at the time, new) android notification builder. 

User has additional settings under Notifications:
notifications

Tapping on reminder sound brings up the standard Android sound picker:

reminder_sound

Notifications sent from OneBusAway adhere to these user preferences.

_________________________________________________

- [X] Run the unit tests with `gradlew connectedObaGoogleDebugAndroidTest connectedObaAmazonDebugAndroidTest` to make sure you didn't break anything

- [X] If you have multiple commits please combine them into one commit by squashing them.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/NotifierTask.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/NotifierTask.java
@@ -66,8 +66,6 @@ public final class NotifierTask implements Runnable {
 
     private String mNotifyTitle;
 
-    private SharedPreferences mSettings;
-
     public NotifierTask(Context context,
                         TaskContext taskContext,
                         Uri uri,
@@ -79,8 +77,6 @@ public final class NotifierTask implements Runnable {
         mUri = uri;
         mNotifyTitle = notifyTitle;
         mNotifyText = notifyText;
-
-        mSettings = Application.getPrefs();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/tripservice/NotifierTask.java
```diff
@@ -74,6 +79,8 @@ public NotifierTask(Context context,
         mUri = uri;
         mNotifyTitle = notifyTitle;
         mNotifyText = notifyText;
+
+        mSettings = Application.getPrefs();
```
waltarmour: Does this need to be a member variable?   It is only used once and comes from a static method so it could just be a local variable down in createNotification().

bridgette: I can make it local

barbeau:
@bridgette This is great, thanks for working on it!  Looks good at a glance (thanks for moving `mSettings` to local variable), let me test it out and then I'll merge and get it out in the next beta.

bridgette:
@barbeau cool, thanks for giving it a try, and let me know if anything needs improvement. I tested it out on a Pixel and it worked like a charm, but I don't have access to any other Android devices.

barbeau:
Thanks @bridgette, this is great!!

bridgette:
While testing out the notification code, I noticed that the "create reminder" user flow is a little non-intuitive. 
@barbeau is there an active Slack channel or email alias &c. where I could propose some design tweaks?

barbeau:
@bridgette Sure, proposals welcome!  Join OBA Slack at https://onebusaway.herokuapp.com/, and there is a #android and #ux channel - sounds like this should be better off in the #ux channel, as other #ux folks hang around there too.  Once we settle on an overall concept I'd prefer that we open a Github issue to go from there - we're over the free tier message limit on Slack and messages start disappearing, so it's not a good home for things we should have a permanent record of.

@cagryInside was working on a more material design UI for reminders at one point and I saw some early prototypes.  

@cagryInside is this code on Github?  I looked at your repo but didn't see an obvious branch.

cagryInside:
> @cagryInside is this code on Github? I looked at your repo but didn't see an obvious branch.

@barbeau I made a detailed search and  looked at my local branches, but couldn't find it. Very sorry about that, my mistake! I should have committed that before.

barbeau:
@cagryInside no worries, thanks for looking!

Issue #792, opened 2017-08-30
Bike share FAB shows layer off on initial install


barbeau:
**Summary:** 

~~If you switch to the Agency X or Agency Y build variants~~ If you are located in Tampa, on initial install of these variants the bike share markers show up on the map, but if you tap on the bikeshare Layer floating action button it shows the layer as turned off (i.e., gray).

![image](https://user-images.githubusercontent.com/928045/29885945-80ad2fc8-8d86-11e7-8ee1-90d423c00b9b.png)

The bike share layer FAB icon should be blue to indicate the layer is turned on

**Steps to reproduce:** 
1. Set up an emulator or device with Tampa GPS location
~~1. Switch to agencyXGoogleDebug or agencyYGoogleDebug in Android Studio and do an initial install (see below for where you can see the "Build Variants" window in lower-left of Android Studio and where you can change build variants)~~
1. Uninstall and re-install the app
1. Tap on layers floating action button

**Expected behavior:** 

The bike share markers should appear on the map and the bike share layer control should be blue to indicate it is on

**Observed behavior:** 

The bike share markers appear on the map but the bike share layer control is gray, which inaccurately indicates the layer is turned off

**Device and Android version:** 

LG G5 with Android 7.0

@carvalhorr Would you be able to take a look at this?  Not an emergency, but I'd like to get it fixed in the next week or two.

barbeau:
@carvalhorr Actually, looks like this affects the main OBA variant as well, so that bumps the priority a bit.  We will need to fix this before exiting beta.

![image](https://user-images.githubusercontent.com/928045/29886283-89e07b4e-8d87-11e7-99d5-3fdc61950d54.png)

barbeau:
https://github.com/carvalhorr/onebusaway-android/commit/a209c1da297c99e4f6accc847dc282bf6408f6e8

PR #804, opened 2017-09-22
Fix #803 - Initialize map layer FAB on startup, then manage visibility

### LINKED ISSUE #803: Blue map layer FAB is visible when starting app with Starred stops
**Summary:** 

If HomeActivity starts up and shows the "Starred stops" or "My reminders" fragments (instead of the BaseMapFragment), the partially initialized map layer FAB shows up in the bottom right corner as a blue circle:

image

**Steps to reproduce:** 

1. Start the app and browse to "Starred stops" or "My reminders" sections via the nav drawer
1. Close the app by hitting the back button or multitasking button and clicking the "X" in the upper right corner of the OneBusAway app
1. Tap on the OBA icon to start the app again

**Expected behavior:** 

The map layer FAB should be hidden when starting the app and showing the "Starred stops" or "My reminders" fragment

**Observed behavior:** 

The map layer FAB is visible as a blue circle

LG G5 w/ Android 7.0

**Screenshots:** 

See above

cc @carvalhorr

barbeau:
* This moves the initialization of the map layer FAB into the setup method called from HomeActivity.onCreate(), so it's initialized no matter what the initial visiblity state should be.  Then, based on the region availability of bikeshare and which fragment is currently visible (e.g., Starred Stops or My reminders), the visiblity is toggled.  This treats the map layer FAB similarly to how the My Location FAB is treated.

@carvalhorr Could you please review to make sure I didn't miss anything?  My primary concern is to make sure we're not triggering bikeshare API calls even when the other non-map fragments are visible and the map isn't.  From my testing it seems like this solution wouldn't have that side-effect, but I'd like you to have a look just to make sure.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
@@ -1538,7 +1538,7 @@ public class HomeActivity extends AppCompatActivity
         } else {
             mLayersFab.setVisibility(View.GONE);
         }
-        mLayersFab.requestLayout();
+        mLayersFab.rebuildSpeedDialMenu();
     }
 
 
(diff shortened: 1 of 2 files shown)


barbeau:
After testing more I believe this PR has a regression for https://github.com/OneBusAway/onebusaway-android/issues/792.  I need to test a bit more.

PR #847, opened 2018-01-31
Fix #833 -- Remove the expand/collapse view from the navigation drawer

### LINKED ISSUE #833: ES - Remove expand/collapse in navigation drawer
**Summary:** 

Early in the Embedded Social design, we had a number of entries (I believe around 5-6) in the navigation drawer), which was a bit overwhelming.  To reduce the impact of this on users, especially those not interested in the social features, we added an expand/collapse feature for the social feature that allowed hiding some of them (screenshot from Galaxy S8+):

image

However, the most I use this feature the more I dislike it.  It breaks with the normal navigation drawer design patterns, and most importantly I don't think it's needed anymore.  We've hidden many of the ES features until the user logs in, which reduces the number of ES entries when the user is logged out to 2 (screenshot from Nexus 5):

image

And, if a user logs in, they are probably interested in the social features, in which case we probably shouldn't hide some of them.  And, when it's expanded the expand/collapse chevron icon takes up almost an entire navigation drawer entry, which is just wasted space.  Here are screenshots from a Nexus 5 after the user logs in, with ES entries collapsed on left and expanded on right:

image

So, I'm proposing that we remove the expand/collapse feature and always show all ES navigation drawer entries when the user is logged in.  Testing with a Nexus 5, a smaller/mid size screen by today's standard, it doesn't seem like getting rid of the expand/collapse feature would cause a problem.  Also, testing on a large screen Samsung Galaxy S8+, removing the expand/collapse feature definitely makes sense - it just looks so strange to include it given the amount of screen real estate.

@acrown-msft What are your thoughts on this?

**Steps to reproduce:** 

1. Open the navigation drawer
1. Log into ES if you're not logged in
1. Open the navigation drawer again

**Expected behavior:** 

Show me the entire navigation drawer without the expand/collapse feature

**Observed behavior:** 

The expand/collapse feature takes up additional screen room and isn't needed.

Nexus 6 w/ Android 6.0 and Samsung Galaxy S8+ with Android 8.0 beta

**Screenshots:** 

See above

acrown-msft:

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
@@ -36,7 +36,6 @@ import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.support.v4.content.res.ResourcesCompat;
 import android.support.v4.view.GravityCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBar;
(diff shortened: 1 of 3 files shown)


barbeau:
Looks much cleaner, thanks!

PR #859, opened 2018-02-12
Make ES "My Profile" screen a fragment instead of activity

acrown-msft:
Load Embedded Social fragments, rather than activities, from the OBA navigation drawer.

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -419,7 +419,7 @@ dependencies {
     // Amazon Maps (only for Amazon flavor)
     amazonImplementation 'com.amazon.android:amazon-maps-api:2.0'
     // Embedded Social SDK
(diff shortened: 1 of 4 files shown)


barbeau:
@acrown-msft It looks like this PR is for changing the "M y Profile" screen from an Activity to `MyProfileFragment`, and then removing the logic required for handling social Activities via the navigation drawer, right?  I just want to make sure I get the merged commit title right.

barbeau:
Ok, thanks.  Does this bump in library version fix the bug you mentioned in the ES SDK fragment handlers, or is that still outstanding?  If outstanding, could you please open an new issue so we can track that?

acrown-msft:
The bump in library version fixes the issue of having multiple listeners responding to the same event.

barbeau:
Great, thanks!

barbeau:
Thanks @acrown-msft!

PR #862, opened 2018-03-08
Fix #834 -- Check to see if the account used to authenticate with Emb…

### LINKED ISSUE #834: ES - When Google account is removed, user should be logged out of OBA ES
**Summary:** 

I just added a 2nd Google account to a device, logged into ES in OBA using that 2nd Google account, and then removed the 2nd Google account from the device.  When I re-opened OBA, it still showed me as logged in using that 2nd Google account.

@acrown-msft Do you think that OBA should log the user out of ES in this case?  Or is the current behavior correct?

**Steps to reproduce:** 

1. Add a 2nd Google account to a device (on a Galaxy S8+ you can do this via System "Settings -> Cloud and Accounts->Accounts->Add account->Google"- note that this is NOT the same thing as adding another user to the device, which is only supported on certain Nexus devices)
1. Log into ES in OBA using that 2nd Google account
1. Removed the 2nd Google account from the device
1. Open OBA and open navigation drawer

**Expected behavior:** 

I should be logged out of OBA ES (?)

**Observed behavior:** 

OBA ES still shows me as logged in

Nexus 5 w/ Android 6.0

acrown-msft:
…edded Social is attached to the device. If so, and that account is later removed from the device, the user is automatically signed out on app start.

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -419,7 +419,7 @@ dependencies {
     // Amazon Maps (only for Amazon flavor)
     amazonImplementation 'com.amazon.android:amazon-maps-api:2.0'
     // Embedded Social SDK
-    implementation('com.acrowntest.test:sdk:0.0.43:release@aar') {
+    implementation('com.acrowntest.test:sdk:0.0.45:release@aar') {
         transitive = true;
     }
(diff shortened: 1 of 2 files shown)


barbeau:
Thanks @acrown-msft!

Issue #769, opened 2017-04-12
Add custom reminder name to notification


barbeau:
**Summary:** 

Currently, you can set a custom reminder name so it shows up in the reminders list, but that name is not shown in the actual reminder notification - we currently just show "Route 5 is arriving in 7 min!".

It would be helpful to add the reminder name to the notification itself so riders can more easily differentiate them.

The reminder notifications could use a UI update in general, too.

**Steps to reproduce:** 

1. Tap on an arrival time
1. Choose "Set a reminder"
1. Set a reminder (if you want to be immediately notified, set it for more than the current ETA)
1. Look at the reminder notification

**Expected behavior:** 

The notification should include the reminder name

**Observed behavior:** 

The notification does not include the reminder name - it just says "Route 5 is arriving in 7 min!".

**Device and Android version:** 

N/A

bridgette:
Hey 👋  
I want to pick this up as my first PR, but I can't assign the issue to myself (or publish a feature branch to repo). @barbeau do you mind making me a collaborator?
thanks!

themonki:
@bridgette you can. Fork the repo, create a branch and pull request!

 See more here https://github.com/OneBusAway/onebusaway-android/blob/master/.github/CONTRIBUTING.md

barbeau:
@bridgette That's great!  Commenting here is enough to call "dibs" on an issue.  Feel free to fork the repo, create a new branch, and create a pull request!

PR #864, opened 2018-03-19
Use app name as default reminder name if none set

mttmllns:
This fixes Pebble watch notifications which apparently need a non-empty title to provide a minute-by-minute bus reminder notification.

Pebble notifications broke as a side effect of PR #769.  I'm just bringing back using app name as a default title instead of using an empty string.


barbeau:
Whoops, sorry we broke Pebble notifications!  Thanks for the fix @mttmllns, looks good!

PR #867, opened 2018-04-04
Fix #630 - Add map zoom control with checkbox preference in settings

### LINKED ISSUE #630: Add display option for map zoom buttons
**Summary:** 

User feedback:

> Just updated from an older version, and found the zoom buttons have been removed, and I was unable to find my stop with one hand, and had to free up my other hand to use pinch to zoom

As a workaround, you can double-tap on the map, but on the 2nd tap hold on the screen, and then drag your thumb on the screen.  It should zoom in and out.

For users that prefer buttons, it should be fairly simple to just add a display option to add them.

m-yang:
Work in progress PR for zoom control (https://github.com/OneBusAway/onebusaway-android/issues/630).

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
@@ -84,6 +84,8 @@ import android.view.Window;
 import android.view.accessibility.AccessibilityManager;
 import android.view.animation.Animation;
 import android.view.animation.Transformation;
+import android.widget.Button;
+import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ListView;
@@ -170,6 +172,11 @@ public class HomeActivity extends AppCompatActivity
     View mArrivalsListHeaderView;
 
     View mArrivalsListHeaderSubView;
+
+    private ImageButton mZoomInBtn;
+
+    private ImageButton mZoomOutBtn;
+
 
     private FloatingActionButton mFabMyLocation;
 
@@ -359,6 +366,8 @@ public class HomeActivity extends AppCompatActivity
         setupLayersSpeedDial();
 
         setupMyLocationButton();
+
+        setupZoomButtons();
 
         setupGooglePlayServices();
 
@@ -1549,6 +1558,27 @@ public class HomeActivity extends AppCompatActivity
             ).show();
         }
         updateLayersFab();
+    }
+
+    private void setupZoomButtons() {
+
+        mZoomInBtn = (ImageButton) findViewById(R.id.btnZoomIn);
+
+        mZoomOutBtn = (ImageButton) findViewById(R.id.btnZoomOut);
+
+        mZoomInBtn.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                mMapFragment.zoomIn();
+            }
+        });
+
+        mZoomOutBtn.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                mMapFragment.zoomOut();
+            }
+        });
     }
 
     private void setupMyLocationButton() {
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
@@ -85,6 +85,7 @@ import android.view.accessibility.AccessibilityManager;
 import android.view.animation.Animation;
 import android.view.animation.Transformation;
 import android.widget.ImageView;
+import android.widget.LinearLayout;
 import android.widget.ListView;
 import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
@@ -406,8 +407,10 @@ public class HomeActivity extends AppCompatActivity
             mArrivalsListHeader.setSlidingPanelCollapsed(isSlidingPanelCollapsed());
         }
 
-        checkDisplayZoomControls();
-        checkLeftHandMode();
+
+        boolean leftHandMode = checkLeftHandMode();
+
+        checkDisplayZoomControls(leftHandMode);
         updateLayersFab();
         mFabMyLocation.requestLayout();
     }
@@ -1579,14 +1582,54 @@ public class HomeActivity extends AppCompatActivity
         }
     }
 
-    private void checkDisplayZoomControls() {
+    private void checkDisplayZoomControls(boolean leftHandMode) {
         boolean displayZoom = Application.getPrefs().getBoolean(
                 getString(R.string.preference_key_show_zoom_controls), false);
 
-        mMapFragment.displayZoomControl(displayZoom);
+        LinearLayout zoomButtonsLayout = (LinearLayout) findViewById(R.id.zoom_buttons_layout);
+
+        if(displayZoom) {
+            zoomButtonsLayout.setVisibility(LinearLayout.VISIBLE);
+        } else {
+            zoomButtonsLayout.setVisibility(LinearLayout.GONE);
+        }
+
+        RelativeLayout.LayoutParams layoutParams = (RelativeLayout.LayoutParams) zoomButtonsLayout
+                .getLayoutParams();
+
+
+        int marginDp = (int) getResources().getDimension(R.dimen.zoom_btn_layout_margin);
+
+        if(leftHandMode) {
+
+            layoutParams.addRule(RelativeLayout.ALIGN_PARENT_LEFT, RelativeLayout.TRUE);
+            layoutParams.addRule(RelativeLayout.ALIGN_PARENT_START, RelativeLayout.TRUE);
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_RIGHT);
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_END);
+            }
+
+            layoutParams.setMarginStart(marginDp);
+            layoutParams.setMargins(marginDp, 0, 0, 0);
+
+        } else /*not left hand mode*/ {
+
+            layoutParams.addRule(RelativeLayout.ALIGN_PARENT_RIGHT, RelativeLayout.TRUE);
+            layoutParams.addRule(RelativeLayout.ALIGN_PARENT_END, RelativeLayout.TRUE);
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_LEFT);
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_START);
+            }
+
+            layoutParams.setMarginEnd(marginDp);
+            layoutParams.setMargins(0, 0, marginDp, 0);
+
+        }
     }
 
-    private void checkLeftHandMode() {
+    private boolean checkLeftHandMode() {
         boolean leftHandMode = Application.getPrefs().getBoolean(
                 getString(R.string.preference_key_left_hand_mode), false);
         if (mFabMyLocation != null) {
@@ -1611,6 +1654,8 @@ public class HomeActivity extends AppCompatActivity
                 mLayersFab.setButtonPosition(POSITION_BOTTOM | POSITION_END);
             }
         }
+
+        return leftHandMode;
     }
 
     /**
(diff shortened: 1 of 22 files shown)


matty-23:
Getting a 503 error code during the CI build. Since I didn't change the gradle files, the error is probably external. 

Since I'm new to this development environment, how important is it for the CI build to succeed for each commit?

barbeau:
@matty-23 Looks like there was an error on the CI build downloading an artifact from Google's server, so definitely something outside of our control.  I just restarted the build so hopefully that fixes it.

It's not critical that the CI build succeeds on each commit, but if it doesn't build I'd definitely take a look and see what the error is, and if it seems to be related to your changes (e.g., you forgot to commit a new file) then fix it.

barbeau:
@matty-23 Thanks for your work on this, it's looking good!

A few comments - here is the default zoom control on left and the custom one in this PR for reference (unpressed state):

image

* I'd increase the width of the plus and minus by 1dp or so to more closely match the default controls
* I'd change the color of plus and minus to be material design gray instead of theme green
* The custom control seems more transparent, or less "white", than the default.  I'd try slightly decreasing the transparency or increasing the "whiteness" to see if you can match the default control more closely
* The default control has rounded corners - you can add `<corners android:radius="6dp"/>` in the `<shape>` to implement this
* The default control has a slight drop shadow.  You can try adding elevation and padding with an `android:background` of a solid color to see if that looks close.

Here's the default zoom control on left and custom on right in the pressed state:

image

Comments:

* The pressed state looks a little too transparent, or not dark enough, compared to the default.

These are nitty-gritty details - you've mostly nailed the overall look.

matty-23:
Thanks for the feedback!

As of now,
- Buttons have been restyled.
- Buttons respect left hand mode . 
- Zoom functionality works. 

Just wanted to mention a couple points about what I've worked on. 

- It turns out that the material gray color is a bit lighter than what the default zoom control has.

- I had some difficulties in perfecting the drop shadows. There's a small difference between what I have and what the default zoom control's drop shadow looks like and I'm not sure what steps to take to make it look more similar. 

Let me know what you think of the changes! :smiley:

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1579,14 +1582,54 @@ public void onClick(View arg0) {
...
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_RIGHT);
+                layoutParams.removeRule(RelativeLayout.ALIGN_PARENT_END);
+            }
+
+            layoutParams.setMarginStart(marginDp);
+            layoutParams.setMargins(marginDp, 0, 0, 0);
+
+        } else /*not left hand mode*/ {
```
barbeau: You can eliminate the comment here, I think it's clear enough given the boolean name

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1579,14 +1582,54 @@ public void onClick(View arg0) {
...
+            zoomButtonsLayout.setVisibility(LinearLayout.GONE);
+        }
+
+        RelativeLayout.LayoutParams layoutParams = (RelativeLayout.LayoutParams) zoomButtonsLayout
+                .getLayoutParams();
+
+
+        int marginDp = (int) getResources().getDimension(R.dimen.zoom_btn_layout_margin);
+
+        if(leftHandMode) {
```
barbeau: If there a reason we shouldn't move this IF statement to the `checkLeftHandMode()` method so it's consistent with the FABs?

barbeau: Also, here and elsewhere please use a space between the `if` and `(` - should be `if (`

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1579,14 +1582,54 @@ public void onClick(View arg0) {
...
 
-    private void checkDisplayZoomControls() {
+    private void checkDisplayZoomControls(boolean leftHandMode) {
         boolean displayZoom = Application.getPrefs().getBoolean(
                 getString(R.string.preference_key_show_zoom_controls), false);
 
-        mMapFragment.displayZoomControl(displayZoom);
+        LinearLayout zoomButtonsLayout = (LinearLayout) findViewById(R.id.zoom_buttons_layout);
+
+        if(displayZoom) {
```
barbeau: should be `if (`

barbeau:
@matty-23 Looks very nice!  You're drop shadow is close enough, I wouldn't worry about any further changes for that.   A few comments inline specific to the code.

One last design comment - the placement of the zoom control looks odd to me up against the side of the My Location (and Layers in the Tampa region) FAB(s).  I understand that for usability you've grouped it next to the FAB (and it can't be above the FAB due to the Layers FAB expanding upward), but I'm leaning towards putting it on the side of the screen opposite of the My Location FAB.  So, for right-hand mode it would be on the left side, and for left hand mode it would be on the right side.  This would be closer to the default controls that appear on their own in a corner of the screen. Any thoughts on this?

matty-23:
@barbeau 

I went ahead and reformatted the code with the AndroidStyle template and factored code for placing the the location FAB and zoom controls in the `checkLeftHandMode()` method. 

> So, for right-hand mode it would be on the left side, and for left hand mode it would be on the right side.

To make sure I'm understanding this correctly, the zoom controls would be on the opposite corner of the My Location FAB?

I do agree that with the zoom controls next to the Location and Layers FAB, the UI does seem cluttered. However, depending on the size of the device's screen, it may be difficult to physically reach zoom controls if they were on opposite corners. In my opinion, the zoom controls and FAB should be grouped together so they can be easily reached.

barbeau:
>the zoom controls would be on the opposite corner of the My Location FAB?

That's what I was proposing, but I agree on some wide devices that could cause usability issues. 

What if we rotate the control 90 degrees clockwise and center it horizontally at the bottom of the screen? This is actually where the Maps API v2 default zoom controls were - plus sign on right and minus on left. This way we wouldn't need to shift control for lefthand mode either.

barbeau:
Example screenshot from Maps API v1 - https://forums.crackberry.com/attachments/bb10-android-app-sideloading-f279/211882d1382231529t-how-fix-any-android-app-dependent-google-maps-api-make-maps-working-2eobi3l.png

m-yang:
Zoom buttons moved to the bottom of the screen. I think this layout looks better than previous layout and is more functional since there less chance of accidentally pressing the FAB when zooming, or vice versa. 

What are your thoughts?

On a side note: I recently changed my github username and the there's some issues with the CLA since the previous commits were tied to my old username. How can I address this?

barbeau:
@m-yang I like it!  I would swap the dimensions of the sides and top/bottom of the buttons, though.

So instead of this:

image


...dimensions would look like:

image

In other words, the height of each button should be less than the width.  Right now it looks like two vertical buttons stacked side-by-side, instead of two horizontal buttons side-by-side.  So, with the changes it would be a true 90 degree clockwise turn of the original vertical control.

Don't worry about the CLA, I'm not sure if there is a way to resolve it with your old name.  As long as the change of names is documented in this PR you should be fine.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1578,9 +1613,32 @@ public void onClick(View arg0) {
...
+
+    }
+
     private void checkLeftHandMode() {
         boolean leftHandMode = Application.getPrefs().getBoolean(
                 getString(R.string.preference_key_left_hand_mode), false);
+
+        setFABLocation(leftHandMode);
+
+        //setZoomControlLocation(leftHandMode);
```
barbeau: This can be removed now.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1605,6 +1663,45 @@ private void checkLeftHandMode() {
         }
     }
 
+    private void setZoomControlLocation(boolean leftHandMode) {
```
barbeau: This can be removed now

onebusaway-android/src/main/res/drawable/zoom_btn_drop_shadow_left.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
```
barbeau: Could you please remove one of the new lines here (and elsewhere) so there is only one line in between elements?

onebusaway-android/src/main/res/layout/main.xml
```diff
@@ -65,15 +65,41 @@
...
+
+                <LinearLayout
+                        android:id="@+id/zoom_buttons_layout"
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:orientation="horizontal" android:layout_centerHorizontal="true"
+                        android:layout_alignBottom="@+id/btnMyLocation">
+
+                    <ImageButton
+
```
barbeau: Could you please remove the new line here for consistency?

onebusaway-android/src/main/res/layout/main.xml
```diff
@@ -65,15 +65,41 @@
...
+                            android:id="@+id/btnZoomOut"
+                            android:layout_width="wrap_content"
+                            android:layout_height="wrap_content"
+                            android:background="@drawable/zoom_btn_drop_shadow_left"
+                            android:src="@drawable/zoom_out_button"/>
+
+                    <ImageButton
+
```
barbeau: Could you please remove the new line here for consistency?

onebusaway-android/src/main/res/values/dimens.xml
```diff
@@ -112,4 +112,19 @@
...
+    <dimen name="zoom_btn_shadow_padding">0.75dp</dimen>
+    <dimen name="zoom_btn_layout_margin">83dp</dimen>
+    <dimen name="zoom_btn_border">1.5dp</dimen>
+    <dimen name="zoom_btn_bottom_offset">1dp</dimen>
+    <dimen name="zoom_out_btn_left_offset">4dp</dimen>
+    <dimen name="zoom_btn_top_offset">@dimen/zoom_btn_bottom_offset</dimen>
+    <dimen name="zoom_out_btn_right_offset">@dimen/zoom_out_btn_left_offset</dimen>
+    <dimen name="zoom_in_btn_left_offset">4.5dp</dimen>
+    <dimen name="zoom_in_btn_right_offset">@dimen/zoom_in_btn_left_offset</dimen>
+
```
barbeau: Could you please remove the extra new line here for consistency?

onebusaway-android/src/main/res/values/strings.xml
```diff
@@ -568,6 +568,12 @@
         negative arrival times
     </string>
 
+    <string name="preferences_show_zoom_controls_title">Show zoom controls</string>
+    <string name="preferences_show_zoom_controls_summary">Display zoom controls on map
+    </string>
+
```
barbeau: Could you please remove the extra new lines here for consistency (just one line break)?

barbeau:
@m-yang This looks good to me!  Thanks for all your work on this.  There is just some cleanup remaining that I commented on inline (remove unused left-hand mode for zoom buttons, remove extra newlines), and then we can get this merged.

m-yang:
@barbeau 

Thanks for the feedback! 

I want to point out an issue. With left hand mode enabled, the My Location FAB partly covers the Google logo in the bottom corner of the screen. We would want to move the logo so that's it's entirely visible to comply with the API's terms of service. 

One option is to move the logo to the bottom right corner of the screen in left hand mode and have it remain in its bottom left corner in right hand mode. Another option we have is to move the FAB up a bit. 

What are your thoughts?

barbeau:
>I want to point out an issue. With left hand mode enabled, the My Location FAB partly covers the Google logo in the bottom corner of the screen. We would want to move the logo so that it's entirely visible to comply with the API's terms of service.

That's a good point - I didn't catch that in the original implementation.

>One option is to move the logo to the bottom right corner of the screen in left hand mode and have it remain in its bottom left corner in right hand mode.

Yes, I think this is the best solution.  Could you please open a new issue for this?  We can get this PR merged and tackle that in a separate workstream.

m-yang:
@barbeau 

Styled the code! Let me know your feedback.

barbeau:
Looks great!  Thanks for your work on this @m-yang!  It will go out in the next release (not v2.3.5, which is already rolling out now).

PR #869, opened 2018-04-10
Fix #868 - Use application context to check for user level restrictions

### LINKED ISSUE #868: NPE in EmbeddedSocialUtils.isSocialRestricted() after AsyncTask callback
**Summary:** 

From the Android developer console (1 crash):

~~~
java.lang.NullPointerException: 
  at org.onebusaway.android.util.EmbeddedSocialUtils.isSocialRestricted (EmbeddedSocialUtils.java:112)
  at org.onebusaway.android.util.EmbeddedSocialUtils.isSocialEnabled (EmbeddedSocialUtils.java:155)
  at org.onebusaway.android.ui.NavigationDrawerFragment.populateNavDrawer (NavigationDrawerFragment.java:429)
  at org.onebusaway.android.ui.HomeActivity.redrawNavigationDrawerFragment (HomeActivity.java:1383)
  at org.onebusaway.android.ui.HomeActivity.onRegionTaskFinished (HomeActivity.java:1532)
  at org.onebusaway.android.region.ObaRegionsTask$2.run (ObaRegionsTask.java:268)
  at android.os.Handler.handleCallback (Handler.java:751)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:163)
  at android.app.ActivityThread.main (ActivityThread.java:6228)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:886)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:776)
~~~

I'm guessing that the user opened the app and then quickly exited it.  Then, the callback coming from the `ObaRegionsTask` triggered the NavigationDrawerFragment to be redrawn, which in turn triggered the check to see if the embedded social features for that region were enabled or restricted by the user.  And, by the time `EmbeddedSocialUtils.isSocialRestricted()` was called the `HomeActivity` context passed into `EmbeddedSocialUtils.isSocialEnabled()` no longer existed, generating the NPE when it tried to look up `((UserManager)context.getSystemService(Context.USER_SERVICE))` in `EmbeddedSocialUtils.isSocialRestricted()`.

@acrown-msft One solution for this is to pass in the Application context into `EmbeddedSocialUtils.isSocialEnabled()`, like:

~~~
            if (EmbeddedSocialUtils.isSocialEnabled(Application.get().getApplicationContext())) {
                // Social items
~~~

That way even if the Activity is destroyed the Application context should still exist.  An alternate approach would be to just return true from `isSocialRestricted()` if the `context` passed into the method is null.  I'm guessing this NPE will only happen on lower-end devices with less memory that are more aggressive with killing activities - I'm not sure if they would destroy the application context as aggressively.  If so, then the safer solution is to return true and not rely on the context existing.

Please take a look and let me know what you think.

**Steps to reproduce:** 

Unsure - probably quickly launch and stop the app, but due to the AsyncTask it will likely be hard to reproduce.

**Expected behavior:** 

Not crash

**Observed behavior:** 

Crash

ZTE ZTE AVID 4 (calbee), 2048MB RAM, Android 7.1

acrown-msft:

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListAdapterStyleB.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListAdapterStyleB.java
@@ -208,7 +208,7 @@ public class ArrivalsListAdapterStyleB extends ArrivalsListAdapterBase<CombinedA
         });
 
         ObaRegion currentRegion = Application.get().getCurrentRegion();
-        if (currentRegion != null && !EmbeddedSocialUtils.isSocialEnabled(context)) {
+        if (currentRegion != null && !EmbeddedSocialUtils.isSocialEnabled()) {
             discussBtn.setVisibility(View.GONE);
         }
 
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
@@ -733,10 +733,10 @@ public class ArrivalsListFragment extends ListFragment
 
         List<String> items = UIUtils
                 .buildTripOptions(getActivity(), isRouteFavorite, hasUrl, isReminderVisible);
-        List<Integer> icons = UIUtils.buildTripOptionsIcons(getContext(), isRouteFavorite, hasUrl);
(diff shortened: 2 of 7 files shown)


acrown-msft:
Fixes #868

barbeau:
Thanks @acrown-msft!

PR #890, opened 2018-06-07
Fix #889 Remove "Popular" from the navigation drawer.

### LINKED ISSUE #889: Social "Popular" feed is empty
**Summary:** 

The "Popular" topics section of the navigation drawer is always empty.

**Steps to reproduce:** 

Open the popular topics feed from the navigation drawer.

**Expected behavior:** 

Display popular stops and routes.

**Observed behavior:** 

UI displays an empty feed.

Current solution will be to remove this item from the Navigation Drawer until an Embedded Social server-side fix is ready.

acrown-msft:

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
@@ -81,6 +81,9 @@ public class NavigationDrawerFragment extends Fragment {
     protected static final int NAVDRAWER_ITEM_SEND_FEEDBACK = 5;
 
     protected static final int NAVDRAWER_ITEM_PLAN_TRIP = 6;
+    
+    // 7 is reserved for re-adding the Embedded Social "Popular" item - See #889
+    //protected static final int NAVDRAWER_ITEM_POPULAR = 7;
 
     protected static final int NAVDRAWER_ITEM_PINS = 8;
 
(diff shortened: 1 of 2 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/NavigationDrawerFragment.java
```diff
@@ -82,8 +82,6 @@
 
     protected static final int NAVDRAWER_ITEM_PLAN_TRIP = 6;
 
-    protected static final int NAVDRAWER_ITEM_POPULAR = 7;
-
```
acrown-msft: Note that popular still appears in the NAVDRAWER_TITLE_RES_ID, NAVDRAWER_ICON_RES_ID, and NAVDRAWER_ICON_SECONDARY_RES_ID arrays.  This is to avoid a bounds issue in makeNavDrawerItem.  We can remove this entry in these arrays by decrementing the other constants by 1.

barbeau: I think it's fine to leave the resource references as-is.  I'm also going to keep the constant here and add a comment explaining why it's still there with a reference to the issue number.

barbeau:
Thanks @acrown-msft!

barbeau:
BTW, I double-checked the behavior for if you have "Popular" selected, quit the app, apply the update, and re-start.  You just get a blank screen, and then can re-select one of the other fragments from the nav drawer.  IMHO this is acceptable, as there are very few, if any, users that will experience this.

PR #900, opened 2018-07-09
WIP - Fix #896 - Use new OTP API v1.x format exclusively

### LINKED ISSUE #896: Legacy OTP code can exacerbate connectivity issues
**Summary:** 

We have some backwards compatibility code in OBA Android to handle older versions of OpenTripPlanner. This is only used under failure conditions where the main URL isn't responding. Something rare event must have happened yesterday (device, network or server) that tripped this for a few Tampa users at the same time. I received multiple error report emails from users with the 404 errors for the URL being used, which made me think it was a server-side issue.  However, while the initial connectivity problem wasn't within the app, the app failing over to the OTP legacy code can exacerbate the problem and cause the user to have connectivity problems planning trips until they restart the app (i.e., the legacy URL keeps getting used and fails with newer OTP versions).

I need to revisit how we're using backwards compatibility for OBA Android and OTP servers, and at this point if Puget Sound is using the newer OTP URL format I should yank out the compatibility code so this doesn't happen again.

**Steps to reproduce:** 

1. Plan a trip and have a series of failed requests

**Expected behavior:** 

1. Recover and successfully plan a trip as soon as network/server connectivity is restored

**Observed behavior:** 

1. If a legacy URL works one, then the app falls back to permanently using that legacy OTP URL.  If the legacy URL is removed, the app will never use the new URL unless it's reinstalled.

N/A

barbeau:
* This removes the format for the OTP API pre v1.0

TODO:
* This fails with a HTTP 404 for Sound Transit's OTP server.  Need to follow up with Sound Transit and see if it's a server whitelist issue or something else.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -304,7 +304,6 @@ public class Application extends MultiDexApplication {
             setCustomApiUrl(null);
             if (regionChanged && region.getOtpBaseUrl() != null) {
                 setCustomOtpApiUrl(null);
-                setUseOldOtpApiUrlVersion(false);
             }
         } else {
             //User must have just entered a custom API URL via Preferences, so clear the region info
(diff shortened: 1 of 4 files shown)


barbeau:
So Sound Transit is using the old URL format with their v1.x OTP deployment, so this won't work as proposed.  I'm closing this for now until we can find a better solution.

PR #906, opened 2018-08-12
Fix #895 - Insert route into recents when viewing a route's map/status from a stop

### LINKED ISSUE #895: Recent routes isn't populated as often as it should be
**Summary:** 

Right now, I believe the only action that will trigger a route to be added to the "popular routes" list is searching for that route_id in the main search method.  We should add a route to favorites for other actions viewing routes, including:
* Show route on map
* Show trip status

**Steps to reproduce:** 

1. Tap on arrival time
1. Select "Show route on map" or "Show trip status"
1. Go back to main screen (hit back several times)
1. Tap on 3 dots overflow button and select "Recent stops/routes"
1. Tap on "Routes" tab

**Expected behavior:** 

The route I just looked at should be shown in the recent routes list.

**Observed behavior:** 

The route that I just looked at isn't added to the recent routes list.

N/A

chasedog:
* Fix order of assertEquals
* Delete row after ProviderTest.testLimit

`ProviderTest.testStops` was failing because state from `testLimit` was left over. Is `setUp()` not called per test method? Is there a better fix?

Do I need to add any tests for my code?

Not able to run tests through gradlew but I was able to run with Android Studio.
Output from gradlew:
```
* What went wrong:
A problem occurred configuring project ':onebusaway-android'.
> Could not generate a proxy class for class com.android.build.gradle.tasks.BuildArtifactReportTask.
```

Tested with Pixel 2XL with Pie in Puget Sound, San Diego, and Tampa regions.

In the Puget Sound region, the description for recent routes isn't correct when it is first added (e.g. **240** doesn't have its description in the screenshot below), but it gets populated when after you click it. It does seem to work for other regions though. Is this important?
screenshot_20180812-102822

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
@@ -60,9 +60,9 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
                 new String[]{ObaContract.Stops._ID, ObaContract.Stops.DIRECTION},
                 null, null, null);
         assertNotNull(c);
-        assertTrue(c.getCount() == 1);
+        assertEquals(1, c.getCount());
         c.moveToNext();
-        assertEquals(c.getString(0), stopId);
+        assertEquals(stopId, c.getString(0));
         c.close();
 
         // Test counting
@@ -70,16 +70,16 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
                 new String[]{ObaContract.Stops._COUNT},
                 null, null, null);
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
+        assertEquals(1, c.getCount());
         c.moveToNext();
-        assertTrue(c.getInt(0) == 1);
+        assertEquals(1, c.getInt(0));
         c.close();
         // Get the one that we care about
         c = cr.query(uri, new String[]{ObaContract.Stops.CODE}, null, null, null);
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
+        assertEquals(1, c.getCount());
         c.moveToNext();
-        assertEquals(c.getString(0), "11060");
+        assertEquals("11060", c.getString(0));
         c.close();
 
         //
@@ -88,15 +88,15 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
         values = new ContentValues();
         values.put(ObaContract.Stops.USE_COUNT, 1);
         int result = cr.update(uri, values, null, null);
-        assertEquals(result, 1);
+        assertEquals(1, result);
 
         //
         // Delete
         //
         result = cr.delete(uri, null, null);
-        assertEquals(result, 1);
+        assertEquals(1, result);
         result = cr.delete(uri, null, null);
-        assertEquals(result, 0);
+        assertEquals(0, result);
     }
 
     @Test
@@ -118,7 +118,7 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
         Uri uri = cr.insert(ObaContract.Stops.CONTENT_URI, values);
         assertNotNull(uri);
 
-        final String stopId2 = "1_1010101";
+        final String stopId2 = "1_1010101-TEST";
         values.put(ObaContract.Stops._ID, stopId2);
         uri = cr.insert(ObaContract.Stops.CONTENT_URI, values);
         assertNotNull(uri);
@@ -127,7 +127,7 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
                 new String[]{ObaContract.Stops._COUNT},
                 null, null, null);
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
+        assertEquals(1, c.getCount());
         c.moveToNext();
         assertTrue(c.getInt(0) == 2);
         c.close();
@@ -140,7 +140,7 @@ public class ProviderTest extends ProviderTestCase2<ObaProvider> {
                 null, null, null
         );
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
+        assertEquals(1, c.getCount());
         c.close();
     }
 }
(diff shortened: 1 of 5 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1616,6 +1621,23 @@ private void addToDB(ObaStop stop) {
         ObaContract.Stops.insertOrUpdate(stop.getId(), values, true);
     }
 
+    private void addRouteToDB(ObaArrivalInfo info) {
```
barbeau: Could you please refactor this method to `QueryUtils`, and see if you can make RouteInfoListFragment and TripInfoActivity use the same method?  We have very similar code floating around in these methods (basically, most places where `ObaContract.Routes
                        .insertOrUpdate()` is used) and it would be nice to get them all using the same method.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1616,6 +1621,23 @@ private void addToDB(ObaStop stop) {
...
+        String routeId = info.getRouteId();
+        String shortName = info.getShortName();
+        String longName = info.getRouteLongName();
+        longName = ConversionUtils.getRouteLongNameSafe(longName, shortName, false);
```
barbeau: `ConversionUtils.getRouteLongNameSafe()` in the `directions` package really needs to be eliminated (it was the side-effect of a quick merge to get something out the door fast for planning trips), as it's basically redundant.  Let's tackle that in a separate issue (feel free to open).  For here, please replace with `UIUtils.getRouteDisplayName(String routeShortName, String routeLongName)`.

barbeau: Also, this might be why you're sometimes getting a blank name in the list.

onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
```diff
@@ -140,7 +140,15 @@ public void testLimit() {
...
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
+        assertEquals(1, c.getCount());
         c.close();
+
+        //
+        // Delete
+        //
+        int result = cr.delete(uri, null, null);
+        assertEquals(1, result);
```
barbeau: The `stop_Id2` in line 121 should probably be changed to `1_1010101-TEST` to avoid a possible collision with a real stop (I know you didn't change this, but we should probably fix it while we're here).

onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
```diff
@@ -60,26 +60,26 @@ public void testStops() {
...
-        assertEquals(c.getString(0), stopId);
+        assertEquals(stopId, c.getString(0));
         c.close();
 
         // Test counting
         c = cr.query(ObaContract.Stops.CONTENT_URI,
                 new String[]{ObaContract.Stops._COUNT},
                 null, null, null);
         assertNotNull(c);
-        assertEquals(c.getCount(), 1);
```
ckragenb: @barbeau this was the assertion that was failing. It was 2 instead of one.

onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
ckragenb: :+1:

onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
chasedog: Whoops. Accidentally commented as my work account ;)

onebusaway-android/src/main/java/org/onebusaway/android/ui/RouteInfoListFragment.java
```diff
@@ -467,15 +464,7 @@ private void setHeader(ObaRouteResponse routeInfo, boolean addToDb) {
             agencyText.setText(mRouteInfo.getAgency().getName());
 
             if (addToDb) {
-                ContentValues values = new ContentValues();
-                values.put(ObaContract.Routes.SHORTNAME, shortName);
-                values.put(ObaContract.Routes.LONGNAME, longName);
-                values.put(ObaContract.Routes.URL, url);
```
chasedog: My new method does not have `URL` in the ContentValues. Is it needed?

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
chasedog: we actually need the description for it to match what's in the recents list. We can get it from `UIUtils.getRouteDescription(ObaRoute route)` but we don't have an instance of `ObaRoute`, so the "long name" is still going to be short name most of the time (until you click). But for now, I have changed it to use `getRouteDisplayName`

chasedog:
@barbeau done implementing code review feedback. Let me know if I need to do anything else.

onebusaway-android/src/main/java/org/onebusaway/android/ui/RouteInfoListFragment.java
barbeau: Yes, let's keep route `URL`.  Even if it's not used today it will likely be used in a future favorite routes list.

onebusaway-android/src/androidTest/java/org/onebusaway/android/provider/test/ProviderTest.java
barbeau: Huh.  Not sure why this is working on my device but not yours.  I'm fine with changing it as you have.

barbeau:
@chasedog Thanks!  A few more in-line comments.  Also, please push individual commits for changes as it makes it easier to review (I know the PR guidelines are outdated, I just fixed them in the master branch).

>In the Puget Sound region, the description for recent routes isn't correct when it is first added (e.g. 240 doesn't have its description in the screenshot below), but it gets populated when after you click it. It does seem to work for other regions though. Is this important?

I still need to poke around this to see why this is happening.

barbeau:
>In the Puget Sound region, the description for recent routes isn't correct when it is first added (e.g. 240 doesn't have its description in the screenshot below), but it gets populated when after you click it. It does seem to work for other regions though. Is this important?

@chasedog Also - was this an issue in the master branch too?

aaronbrethorst:
Thank you for contributing! I am closing due to merge conflicts. I'd love to land this if you have the bandwidth to finish it!

PR #908, opened 2018-08-17
Fix #907 - Fail oba release build if Embedded Social API key is missing

### LINKED ISSUE #907: Invalid/missing Embedded Social API keys are allowed in oba release builds
**Summary:** 

It turns out that when I did the v2.3.9 release, I forgot to update one of the local config files on my machine (not my usual release machine) and as a result the Embedded Social API key went missing from the build, which resulted in the social features disappearing.  I didn't catch this during testing and apparently no one did in beta testing either.

Going forward, to keep this from happening again we should add an explicit check for the `oba` flavor release builds to ensure there is an ES API key, and if not the release build should fail.

**Steps to reproduce:** 

Use a bad API key name in `gradle.properties`, like:
`EmbeddedSocialApiKey_bad=xxxxxxxxxxxxxxxxxx`

...instead of the right key name:
`EmbeddedSocialApiKey_oba=xxxxxxxxxxxxxxxxxx`

Do:

`gradlew assembleRelease -D org.gradle.daemon=false`

or (faster build option):

`gradlew assembleObaGoogleRelease -D org.gradle.daemon=false`

**Expected behavior:** 

Release build should fail due to invalid API key name for `oba` build variant during a release build, but pass in development builds

**Observed behavior:** 

Build is successful, and social features end up missing from a release build.

N/A

barbeau:
#### This PR does the following

When ES key name in `gradle.properties` is **invalid** (for example, `EmbeddedSocialApiKey_bad=xxxxxxxxxxxxxxxxxx`):
* all debug builds (including running from Android Studio) succeed (Example commands - `gradlew assembleObaGoogleDebug`, `gradlew assembleAgencyXGoogleDebug`)
* all non-oba flavor release builds, such as `gradlew assembleAgencyXGoogleRelease  -D org.gradle.daemon=false` and `gradlew assembleAgencyYGoogleRelease  -D org.gradle.daemon=false`, succeed
* `gradlew assembleObaGoogleRelease  -D org.gradle.daemon=false` fails  (this is the important part that fixes #907)

When ES key name in `gradle.properties` is **valid** (for example, `EmbeddedSocialApiKey_oba=xxxxxxxxxxxxxxxxxx`):
* All clean and builds succeed (debug and release) (Example commands - `gradlew assembleObaGoogleDebug`, `gradlew assembleObaGoogleRelease  -D org.gradle.daemon=false`, `gradlew assembleAgencyXGoogleDebug`, `gradlew assembleAgencyXGoogleRelease  -D org.gradle.daemon=false`

####  TODO:
- [x] Currently `gradlew clean` fails when the oba flavor key name is invalid (i.e., , `EmbeddedSocialApiKey_bad=xxxxxxxxxxxxxxxxxx` or `EmbeddedSocialApiKey_xxx` is missing) presumably because it's cleaning release builds too.  We need to fix so we can clean the project for debug builds without the ES key.

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -238,6 +238,14 @@ android {
             }
     }
 
+    variantFilter { variant ->
+        def names = variant.flavors*.name
+        // Make sure that OBA release builds always have an Embedded Social API key (#907)
+        if (names.contains("oba") && variant.buildType.name == "release" && getEmbeddedSocialApiKey(variant.flavors.get(0).name) == "") {
+            throw new InvalidUserDataException("gradle.properties must contain an EmbeddedSocialApiKey_oba API key for oba release builds")
+        }
+    }
+
     // http://stackoverflow.com/questions/20673625/gradle-0-7-0-duplicate-files-during-packaging-of-apk
     packagingOptions {
         exclude 'META-INF/LICENSE'
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -238,14 +238,6 @@ android {
             }
     }
 
-    variantFilter { variant ->
-        def names = variant.flavors*.name
-        // Make sure that OBA release builds always have an Embedded Social API key (#907)
-        if (names.contains("oba") && variant.buildType.name == "release" && getEmbeddedSocialApiKey(variant.flavors.get(0).name) == "") {
-            throw new InvalidUserDataException("gradle.properties must contain an EmbeddedSocialApiKey_oba API key for oba release builds")
-        }
-    }
-
     // http://stackoverflow.com/questions/20673625/gradle-0-7-0-duplicate-files-during-packaging-of-apk
     packagingOptions {
         exclude 'META-INF/LICENSE'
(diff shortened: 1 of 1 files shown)


barbeau:
@paulnabanita1 Could you please see if you can finish implementing this feature?  We can talk more in our meeting today.

paulnabanita1:
@barbeau Sure, I will have a look into this.

barbeau:
Another possible strategy - we currently only ask for passwords on the command line when building the project for release builds - you can see this code at line 390:

~~~
tasks.whenTaskAdded { theTask ->
    if (theTask.name.matches("\\bpackage\\w+Release\\b")) {
        theTask.dependsOn "askForPasswords"
    }
}
~~~

This matches on both the "package" and "Release" names - maybe we can add another IF clause here, something like:

~~~
tasks.whenTaskAdded { theTask ->
    if (theTask.name.matches("\\bpackage\\w+Release\\b")) {
        theTask.dependsOn "askForPasswords"
    }
    if (theTask.name.matches("\\bpackage\\w+oba\\w+Release\\b") && !project.hasProperty('EmbeddedSocialApiKey_oba')) {
        throw new InvalidUserDataException("gradle.properties must contain an EmbeddedSocialApiKey_oba API key for oba release builds")
    }
}
~~~

PR #918, opened 2018-10-02
Added LocationUtils.isDuplicate() method to NavigationTest to check for duplicates

paulnabanita1:
Also fix indexes for trips 2 and 23c, which were incorrect.


onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -792,7 +844,7 @@ public int getAutomatedGetReadyIndex() {
         void runSimulation(int expectedGetReadyIndex, int expectedPullCordIndex) {
             NavigationServiceProvider provider = new NavigationServiceProvider(mTripId,
                     mDestinationId);
-
+            Location mPrevLocation = null; // variable to store the location to check for duplicates
```
barbeau: Could you please change this to `prevLocation`?  The `m` prefix is only for class member variables.

barbeau: Also, you can remove the comment here.  Typically I don't like comments on the same line as code, and I think the use of this is obvious enough we don't need a comment.

onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -831,7 +890,14 @@ void runSimulation(int expectedGetReadyIndex, int expectedPullCordIndex) {
                 } catch (Exception e) {
                     Log.e(TAG, e.getMessage());
                 }
-                provider.locationUpdated(l);
+
```
barbeau: I believe above line 886 when this 2nd `for` loop starts, you'll need to re-initialize mPrevLocation` to `mLocations[expectedGetReadyIndex - 1]`.

The first `for` loop ends with `i = expectedGetReadyIndex`, and the second for loop starts with `i = expectedGetReadyIndex`, so starting the 2nd loop `mPrevLocation` and  `mLocations[i]` would both be pointing to `mLocations[expectedGetReadyIndex]`, which will then always be skipped.

Please take a closer look at this to make sure I'm reading the code right.

paulnabanita1:
@barbeau Thank you for your review comments. I have modified the PR to incorporate the changes. Please let me know if any other changes are required.

barbeau:
@paulnabanita1 is this PR ready for review and merge now?

paulnabanita1:
@barbeau Yes, it is ready for review and merge. I have updated the pull request to fix testTrip2() and testTrip23c().

barbeau:
@paulnabanita1 Can you please add a short comment here about why the indexes for trips 2 and 23c were changed?

A description (with screenshots if necessary) like those you added in https://github.com/OneBusAway/onebusaway-android/pull/925 would be excellent.

onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -808,7 +809,14 @@ void runSimulation(int expectedGetReadyIndex, int expectedPullCordIndex) {
...
+                // Code added to check for duplicate locations in the .csv log files
+                if (prevLocation == null) {
```
barbeau: Could you please collapse this IF statement to:

`if (prevLocation == null || !LocationUtils.isDuplicate(prevLocation, l)) {`

onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -824,14 +832,25 @@ void runSimulation(int expectedGetReadyIndex, int expectedPullCordIndex) {
             Boolean check1 = provider.getGetReady() && !provider.getFinished();
             assertTrue(check1);
 
+            if(expectedGetReadyIndex!= 0){
```
barbeau: Should be `if (expectedGetReadyIndex != 0) {` (add spaces)

onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -824,14 +832,25 @@ void runSimulation(int expectedGetReadyIndex, int expectedPullCordIndex) {
...
                 Location l = mLocations[i];
                 try {
                     Thread.sleep((mTimes[i] / SPEED_UP));
                 } catch (Exception e) {
                     Log.e(TAG, e.getMessage());
                 }
-                provider.locationUpdated(l);
+
+                // Code added to check for duplicate locations in the .csv log files
+                if (prevLocation == null) {
```
barbeau: Same as above - you can collapse to:

`if (prevLocation == null || !LocationUtils.isDuplicate(prevLocation, l)) {`

onebusaway-android/src/androidTest/java/org/onebusaway/android/nav/test/NavigationTest.java
```diff
@@ -29,6 +29,7 @@
 import org.onebusaway.android.nav.NavigationServiceProvider;
 import org.onebusaway.android.nav.model.Path;
 import org.onebusaway.android.nav.model.PathLink;
+import org.onebusaway.android.util.LocationUtils; // package added to utilize LocationUtils.isDuplicate() method
```
barbeau: You can remove the comment here, it's not needed.

paulnabanita1:
For testTrip2(), the original expectedPullCordIndex was 14. Since there wasn't a code to check for duplicates in the trip log, the same location was considered as two different locations and the "Pull the cord" notification was getting triggered. However, after adding the check it was found that the correct location for the notification should 15. Hence, the expectedPullCordIndex has been updated to 15.
uptest2

paulnabanita1:
Similar to testTrip2(), the coordinateIDs 17 and 18 describing the same location were considered as two different locations and the "Pull the cord" notification was getting triggered for trip23c(). However, after adding the check it was found that the correct location for the notification should 19. Hence, the expectedPullCordIndex has been updated to 19.
uptest23c

paulnabanita1:
@barbeau Thank you for your review comments. I have committed the necessary changes to the PR.

barbeau:
Thanks @paulnabanita1!  Looks good!

Issue #884, opened 2018-06-05
Google Play warning - app doesn't target API level 26


barbeau:
**Summary:** 

When releasing a new version on Google Play, after uploading the new APK it says:

~~~
Warnings
Your app currently targets API level 21 and must target at least API level 26 to ensure it is built on the latest APIs optimized for security and performance.

From August 2018, new apps must target at least Android 8.0 (API level 26).
From November 2018, app updates must target Android 8.0 (API level 26).

Tip: Change your app's target API level. Learn how.
~~~

Here's the link with more info:
https://developer.android.com/distribute/best-practices/develop/target-sdk

Blocked by:
- [x] Runtime permissions #330 - WIP PR at https://github.com/OneBusAway/onebusaway-android/pull/930
- [x] Notification channels - PR at https://github.com/OneBusAway/onebusaway-android/pull/935
- [x] Background execution of service for notifications - Rescheduling of reminders after boot - https://github.com/OneBusAway/onebusaway-android/issues/934
- [x] Background execution of service -  Embedded Social SDK - https://github.com/OneBusAway/onebusaway-android/issues/933

**Steps to reproduce:** 

Upload new release to Google Play

**Expected behavior:** 

Upload without warnings

**Observed behavior:** 

Google Play warns re: not targeting API level 26

**Device and Android version:** 

N/A

barbeau:
This is blocked by at least runtime permissions, maybe others (we need to check):
https://github.com/OneBusAway/onebusaway-android/issues/330

vinogeo:
Hi you can check this out How to Change Target API Level 26 or Above in Android Studio 
 http://www.oniverse.in/2018/08/how-to-change-target-api-level-26-or.html this might help

barbeau:
Closing this, as we've checked all the items off the list necessary to target API level 26 in a release

ahmedaniss25:
problem fixed here : https://youtu.be/o4iDJWjfUHs

Issue #826, opened 2017-12-21
Hillsborough County Open311 test fails


barbeau:
**Summary:** 

If you run the unit tests, `ReportProblemOpen311Test.testHillsboroughCounty()` fails with:

~~~
junit.framework.AssertionFailedError
at org.onebusaway.android.io.test.ReportProblemOpen311Test.testHillsboroughCounty(ReportProblemOpen311Test.java:103)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:198)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:183)
at android.test.InstrumentationTestRunner.onStart(InstrumentationTestRunner.java:560)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2102)
~~~

It looks like HART changed their Open311 service types (i.e., issue categories) so now only 3 are identified as "transit-related" based on heuristic keyword matching (a word-around because SeeClickFix doesn't support explicit groups or keywords in their Open311 API).  Our threshold for identifying deployments like HART where all services are transit-related is 4 (defined in `ReportConstants.NUM_TRANSIT_SERVICES_THRESHOLD`).

**Steps to reproduce:** 

Run the unit tests

**Expected behavior:** 

All tests should pass

**Observed behavior:** 

`ReportProblemOpen311Test.testHillsboroughCounty()` fails

**Device and Android version:** 

Reprod on Samsung Galaxy S8+ with Android 8.0 beta

barbeau:
Best solution at this point seems to be to adapt our transit heuristics matching to include the word "passenger".  That brings our count of transit-related issues back up to be equal to the threshold at 4.  See below for the current list of HART services (i.e., issue categories).

![image](https://user-images.githubusercontent.com/928045/34275583-fbbd7a90-e66b-11e7-85a5-1ce40a78340d.png)

barbeau:
For the record - this bug manifested in the app for HART service area with some issue categories being split into a "Transit" category, and other issues being split into an "Other" category in the issue category drop-down spinner list.  Because HART has their own SeeClickFix account and controls all the issue categories, all categories should show up under the "Transit" category.

barbeau:
I'm seeing this again - looks like number of heuristic-matched transit-services for HART dipped to 3 again:

![image](https://user-images.githubusercontent.com/928045/35648573-2a9b4a92-06a4-11e8-91ab-8f280ab42f95.png)

[...]

PR #927, opened 2018-10-22
Fix #851 - Allow filtering by arrivals or departures at a stop

### LINKED ISSUE #851: Allow filtering by arrivals or departures at a stop
**Summary:** 

Currently, if there are incoming arrivals and departures at a stop that have the same route and headsign, you can't filter the view to show only the arrivals or only the departures.

From a Puget Sound user:

>Is there a way to filter arrivals/departures at a stop? My bus stop is the northern terminus of the D-Line so there are actually independent arrival and departure times (the busses park there for a while).
>
>Since there are always a couple of D-Line busses parked at the stop, I don't need to see when more busses arrive, only when the ones already there are going to leave.

It would be nice if we could provide a way to filter these to show only arrivals or (probably more useful) only departures.  I've asked the user to provide a screenshot of this case and I'll add here once I get it.

**Steps to reproduce:** 

View arrivals for a stop that has an incoming and outgoing route with the same route and headsign

**Expected behavior:** 

Give me some way to show only the arrivals for this route, or only the departures, or both

**Observed behavior:** 

There are no way to filter only arrivals or departures for this type of route

SAMSUNG-SM-G930A w/ Android 7.0
Google Play Services App: 11.9.75 (448-182402865)
Google Play Services Library: 9452000

sjgllghr:
- Adds option to filter arrivals/departures from stop arrivals. Fixes #851. 
- Saves user preferences for stops (I made a new table for this, but wasn't sure of the best option between that and editing the stops table...)
- Changes 'load more arrivals' and 'no arrivals' messages to say departures when a stop is set to show only departures. This could help indicate to a user that they have 'only departures' on rather than there being no arrivals. 

I think I'm mostly done with this, but want to make sure I'm on the right track.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
@@ -873,61 +873,91 @@ public class UIUtilTest extends ObaTestCase {
     public void testGetNoArrivalsMessage() {
         String result;
 
-        // Arrivals only or both arrivals and departures
-        boolean onlyDepartures = false;
+        // Test for both arrivals and departures
+        ArrivalFilter arrivalFilter = ArrivalFilter.BOTH;
 
         // Less than an hour
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, false, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, false, arrivalFilter);
         assertEquals("No arrivals in the next 35 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, false, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, false, arrivalFilter);
         assertEquals("No additional arrivals in the next 35 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, true, arrivalFilter);
         assertEquals("35+ min", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, true, arrivalFilter);
         assertEquals("35+ min", result);
 
         // More than an hour
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, false, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, false, arrivalFilter);
         assertEquals("No arrivals in the next 1 hour and 15 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, false, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, false, arrivalFilter);
         assertEquals("No additional arrivals in the next 1 hr and 15 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, true, arrivalFilter);
         assertEquals("1h 15+ min", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, true, arrivalFilter);
         assertEquals("1h 15+ min", result);
 
-        onlyDepartures = true;
+        // Test for only arrivals
+        arrivalFilter = ArrivalFilter.ONLY_ARRIVALS;
 
         // Less than an hour
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, false, onlyDepartures);
-        assertEquals("No departures in the next 35 min.", result);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, false, arrivalFilter);
+        assertEquals("No arrivals in the next 35 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, false, onlyDepartures);
-        assertEquals("No additional departures in the next 35 min.", result);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, false, arrivalFilter);
+        assertEquals("No additional arrivals in the next 35 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, false, true, arrivalFilter);
         assertEquals("35+ min", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 35, true, true, arrivalFilter);
         assertEquals("35+ min", result);
 
         // More than an hour
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, false, onlyDepartures);
-        assertEquals("No departures in the next 1 hour and 15 min.", result);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, false, arrivalFilter);
+        assertEquals("No arrivals in the next 1 hour and 15 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, false, onlyDepartures);
-        assertEquals("No additional departures in the next 1 hr and 15 min.", result);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, false, arrivalFilter);
+        assertEquals("No additional arrivals in the next 1 hr and 15 min.", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, true, arrivalFilter);
         assertEquals("1h 15+ min", result);
 
-        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, true, onlyDepartures);
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, true, arrivalFilter);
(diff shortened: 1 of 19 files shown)


sjgllghr:
At least when I run this locally, I consistently fail `testHillsboroughCounty` in `ReportProblemOpen311` even on the master branch. The failure happens here:
https://github.com/OneBusAway/onebusaway-android/blob/212dbac273f620681f4b5897b81145a5a5cc1637/onebusaway-android/src/androidTest/java/org/onebusaway/android/io/test/ReportProblemOpen311Test.java#L109-L112

It looks like s.getGroup() and s.getKeywords() are always returning null, so `markTransitServices` always returns false.

https://github.com/OneBusAway/onebusaway-android/blob/212dbac273f620681f4b5897b81145a5a5cc1637/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/util/ServiceUtils.java#L45

https://github.com/OneBusAway/onebusaway-android/blob/212dbac273f620681f4b5897b81145a5a5cc1637/onebusaway-android/src/main/java/org/onebusaway/android/report/ui/util/ServiceUtils.java#L59

barbeau:
Thanks! I'll try to take a look tomorrow. Yes, don't worry about test failure, that an unrelated issue - https://github.com/OneBusAway/onebusaway-android/issues/826.

barbeau:
@sjgllghr Thanks for working on this!  I skimmed through the code and overall it looks pretty good.  I agree with changing the "no arrivals.." message to "departures" when filtering only for departures.

A few comments/questions in-line.

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
```diff
@@ -782,6 +784,153 @@ public void testArrivalInfoLabels() {
...
+        assertEquals("35+ min", result);
+
+        // More than an hour
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, false, onlyDepartures);
+        assertEquals("No departures in the next 1 hour and 15 min.", result);
+
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, true, false, onlyDepartures);
+        assertEquals("No additional departures in the next 1 hr and 15 min.", result);
+
+        result = UIUtils.getNoArrivalsMessage(getTargetContext(), 75, false, true, onlyDepartures);
```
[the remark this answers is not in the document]
sjgllghr: I was just testing all of the possible paths through `getNoArrivalsMessage`, it doesn't look like there are actually any calls to it with `shortFormat` enabled. So the '1hr 15+ min' and '35+ min' style messages don't appear to be used in the UI anywhere. Would you like me to remove them and/or the tests for them?

sjgllghr:
@barbeau Thanks for the quick review! I agree with your suggested changes, passing the filter enum is a lot cleaner and altering the stops table is a lot simpler.

One question above.

onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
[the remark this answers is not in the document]
barbeau: Oh, thanks! Sorry, forgot about that. No, that's fine, go ahead and leave them and the test (and thanks for the tests). IIRC the short format is used with the narrative text format which I haven't fully ported over from Alexa, but could be really useful for accessibility on Android.

onebusaway-android/src/main/java/org/onebusaway/android/provider/ObaContract.java
```diff
@@ -646,6 +635,33 @@ public static boolean markAsUnused(Context context, Uri uri) {
             values.putNull(ObaContract.Stops.ACCESS_TIME);
             return cr.update(uri, values, null, null) > 0;
         }
+
+        public static int getArrivalFilter(Context context, Uri uri) {
```
barbeau: What do you think about changing the return type here to the enum `ArrivalFilter`?  Seems like a conversion always needs to be done anyway, so we might as well do it here.

onebusaway-android/src/main/java/org/onebusaway/android/provider/ObaContract.java
```diff
@@ -646,6 +635,33 @@ public static boolean markAsUnused(Context context, Uri uri) {
...
+            }
+
+            if (c != null) {
+                c.close();
+            }
+
+            return result;
+        }
+
+        public static void setArrivalFilter(Context context, Uri uri, int filterOption) {
```
barbeau: Same here - what you do you think about changing the int parameter to the enum `ArrivalFilter`?

barbeau:
Thanks @sjgllghr.  Updated changes look look.  Two questions in-line, and let me know if this is ready for merge.

sjgllghr:
Totally agree about the contract methods, those were bothering me too, thanks for the suggestions. I think this should be ready to merge, you picked up on most of the things I was unsure about.

barbeau:
@sjgllghr One last thing we haven't discussed that I forgot to mention earlier (and sorry I haven't gotten to actually run this code on my device yet, I've been swamped lately) - we should include a UI ribbon at the top of the arrivals list to indicate when the user has filtering turned on.  The code and logic for this already exists in context of the route filter, when it says "Showing 3 of 5 routes (show all)" (and service alerts too, which are like "1 active alert is hidden (show all)", although you don't see these in Puget Sound).  So it shouldn't take long to implement.

Could you please add this, if you haven't already?

I think it should show "Departure are hidden (show all)" or "Arrivals are hidden (show all)", but let me know if you have a better idea for text.

sjgllghr:
@barbeau thanks for the suggestion! I’ve added the UI ribbon. For cases where the user has a route and an arrival/departure filter, I took a guess and made it so they both show up, but I'm not 100% sure about the styling/format. Thoughts?

barbeau:
Thanks @sjgllghr, this looks great!  I agree that both filters should be shown because they can operate independently.  For styling, this is fine - ideally we'd have a thin light white separator between the two rows but I wouldn't hold up merging based on that.

Thanks for all your work on this feature!  I'm going to test and then merge.  This will be included in the next release, although the next release currently has a few blockers as discussed in #884 that we're working on resolving first.

barbeau:
@sjgllghr BTW - please feel free to add yourself at the bottom of the acknowledgments section in strings.xml `about_text` string under the "Code" section how you'd like to be named.  Can be in this PR or a different one.

barbeau:
@sjgllghr A few items came up during testing:

1. When one alert, but not all alerts, are hidden, the alert filter and arrivals/departures filter ribbons are separated (see below screenshot) - I believe the best approach is to but arrivals/departures filter at the bottom under the alert filter.  You may have to test in Tampa for this as IIRC Puget Sound doesn't publish many (any?) alerts.  "University Area Transit Center" is a good testing stop.

image

2. Arrivals/departures filter appears in Nearby view when the sliding panel is closed (see below screenshot) - The filter shouldn't be visible until you pull up the panel (same as alert and routes filters).

image

3. When tapping on a stop in Nearby view (with the sliding panel closed) that already has an arrival/departure filter active, there is a strange jumping action where the header tries to resize itself.  This may be due to 2) above as the header layout height is changing based on the filter (and it shouldn't be).  This is the biggest item as it results in some strange layout behavior, including sometimes when the sliding panel header collapses on itself and becomes invisible.

FYI - this PR where I originally added the ability to filter alerts may be helpful (?) - https://github.com/OneBusAway/onebusaway-android/pull/716

If you want to give me push permissions to this branch on your repo I can try to address these items as I find time, or you can take a shot if you wish.  All my spare cycles right now are focused on #884 so I won't have a lot of time to dig into this over the next week or so.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListFragment.java
```diff
@@ -1472,6 +1509,87 @@ private void setRoutesFilter(boolean[] checks) {
...
+
+        boolean switchedFromOnlyDepartures =
+                mArrivalFilter == ArrivalFilter.ONLY_DEPARTURES &&
+                        newFilter != ArrivalFilter.ONLY_DEPARTURES;
+
+        mArrivalFilter = newFilter;
+
+        ObaContract.Stops.setArrivalFilter(getActivity(), mStopUri, ArrivalFilter.fromInt(selection));
+        refreshSituations(UIUtils.getAllSituations(getArrivalsLoader().getLastGoodResponse(), mRoutesFilter));
+        refreshLocal();
```
barbeau: I believe `refreshLocal();` should be moved to the end of this method as it triggers an adapter update.  We want to make sure that any UI content (e.g., "load more" text) is updated before triggering the reload.

sjgllghr:
@barbeau Yikes! Sorry about those, I should be able to look into them over the next couple of weeks. Thanks for your patience.

barbeau:
@sjgllghr No worries!  Let me know if you have problems producing 3).  I tried to record my screen to create a GIF but I'm having device issues capturing via Android Studio right now (TODO - contact Samsung for warranty repair :disappointed:).

sjgllghr:
@barbeau I was able to recreate 3, and I agree that is definitely not ideal behavior (sorry about your phone). I’m a bit confused about 2 though. When I run the app from master or the play store, the route filter does show up at the bottom of the screen when the panel is closed..?

barbeau:
@sjgllghr Re: 2 - sorry, you're right.  That's what I get for firing off a comment without thinking it through.  I went back and tested old v2.0.x versions and they behave the same.  And, thinking back, this makes sense.  I wanted the filter to appear when the sliding panel is closed so you'd know if the 2 times shown in the header are affected.  IMHO Arrive/depart filter has the same potential impact so it should behave the same way - it should appear at the bottom of the screen when the sliding panel is collapsed (as implemented in the PR and pictured in https://github.com/OneBusAway/onebusaway-android/pull/927#issuecomment-440394726).

- [ ] On a related note - I think I found a bug with the current implementation of alert/route filters (i.e., the master branch).  If there are two alerts active for the stop (and you haven't hidden them), and you add a route filter, the route filter ribbon replaces one of the active alerts - that alert just disappears.  I'm not sure if this same behavior will be triggered by the arrive/depart filter, and I'm not sure if this will need to be fixed to get the arrive-depart filter to work right.  If it doesn't impact your work don't feel like you need to fix it in this PR, but just let me know so I can open a separate issue.  I'm also making this a check list item so we can check it off after fixing it or opening a new issue.

sjgllghr:
@barbeau after some further testing it looks like the hidden alert behavior  described in 1) also happens in master if you have a route filter, an active alert and a hidden alert.




I’ve been able to fix some of the issues, but it still behaves strangely at times (the view doesn’t stay invisible but for some reason the arrival/departure filter is the first thing to load). I’ll keep looking at it, but just in case I’ve also given you push permissions.

For the issue you just described, I’m wondering if maybe you filtered out the route the alert was for? If I recall correctly when you filter out alerts with a route filter it doesn’t notify you that they are hidden, which if thats a bug then that is actually my bad…

aaronbrethorst:
@sjgllghr I am going to close this due to age and merge conflicts. I'd love to land this if you have the time and energy to rebase. thanks for your contributions!

PR #931, opened 2018-11-11
Create notification channels

paulnabanita1:
Implementing notification channels for Android versions Oreo and above.
This code will be maintaining backward compatibility with older versions of Android.
Please make sure these boxes are checked before submitting your pull request - thanks!

- [ ] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [ ] Run the unit tests with `gradlew connectedObaGoogleDebugAndroidTest connectedObaAmazonDebugAndroidTest` to make sure you didn't break anything

- [ ] If you have multiple commits please combine them into one commit by squashing them for the initial submission of the pull request.  When addressing comments on a pull request, please push a new commit per comment when possible (reviewers will squash and merge using GitHub merge tool)

PR #935, opened 2018-11-26
Create notification channels

barbeau:
Implementing notification channels for Android versions Oreo and above.
This code will be maintaining backward compatibility with older versions of Android.

This PR is to replace https://github.com/OneBusAway/onebusaway-android/pull/931 due to Git history issues with conflicts.


barbeau:
Great, thanks @paulnabanita1!

PR #947, opened 2018-12-13
Fix #946 - Don't show multiple notifications for reminders

### LINKED ISSUE #946: Multiple reminder notifications are confusing
**Summary:** 

This issue is based on user feedback on the implementation in https://github.com/OneBusAway/onebusaway-android/pull/937.

In https://github.com/OneBusAway/onebusaway-android/pull/937, we modified TripService to always run in the foreground.  The primary reason for this is that on Android 8.0 Oreo and higher you can no longer run a service in the background after receiving a `BOOT_COMPLETED` Intent (i.e., on device reboot) - if you do, you'll get an `IllegalStateException`.  A secondary reason for running the service in the foreground is that anecdotally this results in better reliability of services executing immediately - although in our testing for TripService it seemed to execute immediately even when running in the background.

However, to run a service in the foreground you have to provide a notification to a user.  And, if we always run TripService in the foreground, this results in duplicate notifications shown to the user when they get a reminder, like this:

image

I've gotten feedback that this is confusing, and I'd like to try and improve the user experience surrounding this.

One possible implementation is that we run the TripService in the foreground for all Intents EXCEPT for our `TripService.ACTION_NOTIFY` intent.  This would remove the need for the 2nd icon when the user is actually looking at the notifications (for the other executions of TripService for scheduling and polling the server the notification would still appear, but would disappear so quickly most users would never notice it).  This would fix the UX issue with reminder notifications and hopefully still keep reliability high by running the service in the foreground for the task of scheduling alarms and polling the server.

**Steps to reproduce:** 

Set a reminder on current master branch

**Expected behavior:** 

Give me one notification for my reminder

**Observed behavior:** 

Two notifications are shown, one generic notification being related to running the service in the foreground

Samsung Galaxy S8+ with Android 8.0

barbeau:
Fixes https://github.com/OneBusAway/onebusaway-android/issues/946.

* Allows various application components to explicitly specify whether or not the TripService should run in the foreground
* Never run in the foreground for ACTION_NOTIFY to avoid the "sticky" notification that shows while the service is running
* Always run in the foreground on device reboot
* Run in foreground for most other situations, even as a result of UI interactions.  We could potentially scale this back if it's not found to be necessary.
* Remove deprecated onStart() method, which dates back to Android 2.0 (TBT)
* Some code cleanup
* Update What's New text to remove reference to "sticky" notifications

@paulnabanita1 Could you please review and test?  New expected behavior is that we shouldn't see the foreground service notification when the reminder notifications trigger.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/BootstrapService.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/BootstrapService.java
@@ -28,6 +28,6 @@ public class BootstrapService extends BroadcastReceiver {
      */
     @Override
     public void onReceive(Context context, Intent intent) {
-        TripService.scheduleAll(context);
+        TripService.scheduleAll(context, true);
     }
 }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/CancelNotifyTask.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/tripservice/CancelNotifyTask.java
@@ -53,7 +53,7 @@ public final class CancelNotifyTask implements Runnable {
             values.put(ObaContract.TripAlerts.STATE, ObaContract.TripAlerts.STATE_CANCELLED);
             cr.update(mUri, values, null, null);
 
-            TripService.scheduleAll(mContext);
+            TripService.scheduleAll(mContext, true);
         } finally {
             mTaskContext.taskComplete();
         }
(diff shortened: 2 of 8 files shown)


paulnabanita1:
@barbeau: Thank you for the review request.
I have tested the implemented changes and the app is no longer triggering multiple notifications (one indicating that a service is running in foreground) for bus arrivals. Also, after setting a reminder, I rebooted the device and the app successfully provided notifications for arrivals.

PR #948, opened 2018-12-17
Fix #215 - Add occupancy

### LINKED ISSUE #215: Add support for vehicle occupancy
`OccupancyStatus` is now an "experimental/proposed" GTFS-realtime field, [as discussed on the gtfs-realtime list](https://groups.google.com/d/msg/gtfs-realtime/_HtNTGp5LxM/0mvvEW1WuMMJ).

Here’s an explanation of the enumeration values:
https://groups.google.com/forum/#!msg/gtfs-realtime/_HtNTGp5LxM/0mvvEW1WuMMJ

...and the same text inline, for convenience:

```
 enum OccupancyStatus {
    // The vehicle is considered empty by most measures, and has few or no
    // passengers onboard, but is still accepting passengers.
    EMPTY = 0;

    // The vehicle has a relatively large percentage of seats available.
    // What percentage of free seats out of the total seats available is to be
    // considered large enough to fall into this category is determined at the
    // discretion of the producer.
    MANY_SEATS_AVAILABLE = 1;

    // The vehicle has a relatively small percentage of seats available.
    // What percentage of free seats out of the total seats available is to be
    // considered small enough to fall into this category is determined at the
    // discretion of the feed producer.
    FEW_SEATS_AVAILABLE = 2;

    // The vehicle can currently accommodate only standing passengers.
    STANDING_ROOM_ONLY = 3;

    // The vehicle can currently accommodate only standing passengers
    // and has limited space for them.
    CRUSHED_STANDING_ROOM_ONLY = 4;

    // The vehicle is considered full by most measures, but may still be
    // allowing passengers to board.
    FULL = 5;

    // The vehicle is not accepting additional passengers.
    NOT_ACCEPTING_PASSENGERS = 6;
  }
```

We should consume this data from the OBA REST API (when it is implemented) and expose it in the app.  See https://github.com/OneBusAway/onebusaway-application-modules/issues/121 for progress in providing this data via the OBA REST API.

Companion issue on iOS:
https://github.com/OneBusAway/onebusaway-iphone/issues/361

Please note that not all agencies have the hardware to support real-time occupancy.

We can also support historical occupancy using a very similar format if an agency isn't able to supply real-time occupancy data, but is able to supply historical data via GTFS (exact design TBD).

barbeau:
TODO:
- [x] Initial layout and UI for arrivals list (styles A and B), trip status, and vehicle marker
- [x] Parse server response and unit tests for arrivals-and-departures-for-stop
- [x] Parse server response and unit tests for trips-for-route
- [x] Parse server response and unit tests for trip-details
- [x] Set number of person icons for arrival list style A
- [x] Set number of person icons for map on route map marker balloon
- [x] Set number of person icons for trip details list items
- [x] Set number of person icons for trip details header
- [x] Set content description for accessibility based on occupancy
- [x] Set weight of occupancy icon to left when there is less than 3
- [x] Set number of person icons for arrival list style B
- [x] Set number of person icons for trip problem reporting arrival list
- [x] Test above values by setting when debugging
- [x] Add context menu when tapping on arrival time "About occupancy"


barbeau:
Final result looks like this (for arrivals list):

image

Issue #950, opened 2018-12-17
Switch to HTTPS for all URLs


barbeau:
**Summary:** 

HTTPS is now required by default when targeting Android P API Level 28 - from https://developer.android.com/about/versions/pie/android-9.0-changes-28#framework-security-changes:

>If your app targets Android 9 or higher, the isCleartextTrafficPermitted() method returns false by default. If your app needs to enable cleartext for specific domains, you must explicitly set cleartextTrafficPermitted to true for those domains in your app's Network Security Configuration.

We should change to HTTPS whenever possible.

**Steps to reproduce:** 

Go to URL in app

**Expected behavior:** 

Use HTTPS

**Observed behavior:** 

HTTP is still used for some URLs

**Device and Android version:** 

N/A

PR #952, opened 2018-12-18
Changed target SDK version to 26 to fix Android Pie bus stop detection.

paulnabanita1:
This is a workaround for https://github.com/OneBusAway/onebusaway-android/issues/950 until it's implemented.


barbeau:
Thanks @paulnabanita1!

PR #967, opened 2019-03-11
Nav branch - Collect and upload User feedback

paulnabanita1:
This code does not return the screen to Home activity after submitting feedback,

- [ ] If you have multiple commits please combine them into one commit by squashing them for the initial submission of the pull request.  When addressing comments on a pull request, please push a new commit per comment when possible (reviewers will squash and merge using GitHub merge tool)

Screenshots of UI :
mar21
mar2102


onebusaway-android/src/main/res/values/donottranslate.xml
```diff
@@ -61,6 +61,8 @@
     <string name="preference_key_social_category">preference_category_social</string>
     <string name="preference_key_social">preference_social</string>
     <string name="preference_key_never_show_payment_warning_dialog">never_show_payment_warning_dialog</string>
+    <string name="preferences_key_user_share_logs">preferences_user_share_logs</string>
+    <string name="preferences_key_user_logs">Share trip logs</string>
```
barbeau: You'll need to split this into two strings - the one in `donottranslate.xml` should have the underscores to match the conventions for the other strings, but the one that's used for the preference category title (that the user sees) needs to be in `strings.xml` so it can be translated into other languages.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private void deleteLog() {
+        File lFile = new File(NavigationService.LOG_FILE);
+        boolean deleted = lFile.delete();
+        Log.d(TAG,"Log deleted " + deleted);
+    }
+
+    public void setSendLogs(View view) {
+        CheckBox checkBox = (CheckBox)view;
+        if(checkBox.isChecked()){
```
barbeau: Should have a space between the `if` and `(`, and the `)` and `{`:
`if (...) {`

Same for anywhere else.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
```diff
@@ -0,0 +1,168 @@
...
+
+            try {
+                FileUtils.moveFileToDirectory(
+                        FileUtils.getFile(lFile),
+                        FileUtils.getFile(destFolder), true);
+                Log.d(TAG, "Move file successful.");
+            }
+            catch (Exception e) {
+                Log.d(TAG, "File move failed");
+            }
```
barbeau: Looks like there is an extra bracket `}` here?

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private void submitFeedback() {
+        PreferenceUtils.saveBoolean(NavigationServiceProvider.FIRST_FEEDBACK, false);
+        Log.d(TAG, "First Feedback : " + String.valueOf(Application.getPrefs()
+                .getBoolean(NavigationServiceProvider.FIRST_FEEDBACK, true)));
+        String feedback = ((EditText) this.findViewById(R.id.editFeedbackText)).getText().toString();
+        if (Application.getPrefs()
+                .getBoolean(getString(R.string.preferences_key_user_share_logs), true)) {
+            moveLog(feedback);
+        }
```
barbeau: For all of the IF statements, could you please move the `else` onto the same line as the preceding bracket?

`} else {`

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
```diff
@@ -78,7 +80,7 @@
     private boolean mGetReadyFlag = false;
     private boolean mPullTheCordFlag = false;
 
-    private NavigationServiceProvider mNavProvider;
+    public NavigationServiceProvider mNavProvider;
```
barbeau: Can `mNavProvider` remain `private`?  If so let's change back.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
```diff
@@ -64,6 +64,8 @@
 
     public static final String LOG_DIRECTORY = "ObaNavLog";
 
+    public static String LOG_FILE = null;
```
barbeau: Let's call this `mLogFile` instead, as that's the usual convention for class member variables.  Also, I don't believe this should be `static`?  If there is more than one service they should each have their own file to write too.  This doesn't happen in our current design but we should write the code to handle this.

onebusaway-android/build.gradle
```diff
@@ -450,6 +451,10 @@ dependencies {
     implementation 'androidx.multidex:multidex:2.0.0'
     // Unit tests - seems like this is still necessary w/ Android X even though useLibrary is declared earlier
     androidTestImplementation 'androidx.test:runner:1.1.0'
+    // WorkManager (Java only)
+    implementation "android.arch.work:work-runtime:1.0.0-rc01"
+    // WorkManager - Kotlin + coroutines
```
barbeau: Let's remove the Kotlin + coroutines depedency if we're not using it.

onebusaway-android/src/main/res/values/donottranslate.xml
paulnabanita1: @barbeau I have split into two strings to match the conventions. Please review and let me know if it looks good.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
paulnabanita1: Modified all files to fix the spacing for if and brackets.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
paulnabanita1: We had to include two try-catch blocks - one to append the user free text comment to the log file and the second one to move the file to 'yes'/'no' device folder. There were some misplaced alignments in the '}'. I have fixed the alignments. Please let me know if it looks ok.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
paulnabanita1: Modified the codes to move the else onto the same line.

onebusaway-android/build.gradle
paulnabanita1: Removed the unused dependencies.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
paulnabanita1: We don't need to make the access specifier of mNavProvider public, so changing it back to private.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
paulnabanita1: @barbeau  We are using the String variable "LOG_FILE" to store the device location of a trip log csv file. This variable is used in FeedbackActivity.java and FeedbackReceiver.java to locate the log and move to the 'yes' or 'no' device folders for upload to Firebase. The variable value will remain the same for the entire service. Can you please let me know if it is ok to continue with the existing variable name or should we change the name and access specifier?

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_feedback);
+        setTitle(getResources().getString(R.string.feedback_label));
+
+        Intent intent = this.getIntent();
+        CheckBox sendLogs = (CheckBox) findViewById(R.id.feedback_send_logs);
+        if (intent != null) {
+            mAction = intent.getExtras().getString("CallingAction");
```
barbeau: "CallingAction", "no" and "yes" (and any other intent extras and actions) need to be defined as public static final variables - let's put them in FeedbackActivity and reference them always from here.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
barbeau: Oh, I missed that in my first pass just looking at the code in the PR.  That needs to change - it's not safe to depend on a variable in a Service from an Activity or Receiver - the Service could easily end and be destroyed before the user interacts with the notification to provide feedback, in which case LOG_FILE would be null.

The file name needs to be stored somewhere outside of the Service.  Because we only have one notification prompt for feedback at this point, I believe you could store it as an Intent extra when creating the notification, and then read it from the intent when you receive it in the BroadcastReceiver, and pass it into the Activity via an intent extra as well.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
```diff
@@ -32,17 +25,32 @@
...
 import androidx.core.app.NotificationCompat;
+import androidx.core.app.RemoteInput;
 
 /**
  * This class provides the navigation functionality for the destination reminders
  */
 public class NavigationServiceProvider implements TextToSpeech.OnInitListener {
 
     public static final String TAG = "NavServiceProvider";
+    public static String REPLY_ACTION = "org.onebusaway.android.nav.REPLY_ACTION";
```
barbeau: This should be moved to `FeedbackReceiver` and should be declared as:

public static final String ACTION_REPLY =
            BuildConfig.APPLICATION_ID + ".action.REPLY";

onebusaway-android/src/main/AndroidManifest.xml
```diff
@@ -268,6 +276,16 @@
                 <action android:name=".nav.NavigationReceiver" />
             </intent-filter>
         </receiver>
+        <receiver
+            android:name=".nav.FeedbackReceiver"
+            android:enabled="true"
+            android:exported="true" >
+            <intent-filter>
+                <action android:name=".nav.FeedbackReceiver" />
+                <action android:name="${applicationId}.action.SCHEDULE"/>
```
barbeau: `<action android:name="${applicationId}.action.SCHEDULE"/>` and `<action android:name="${applicationId}.action.POLL"/>` should be removed here and replaced with the Intents that the FeedbackReceiver actually receives, which should be in the same format of:

`<action android:name="${applicationId}.action.REPLY"/>`

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
```diff
@@ -0,0 +1,135 @@
...
+
+    public static final String TRIP_ID = ".TRIP_ID";
+    public static final String NOTIFICATION_ID = ".NOTIFICATION_ID";
+    public static final String CALLING_ACTION = ".CALLING_ACTION";
+    public static final String ACTION_NUM = ".ACTION_NUM";
+
+    public static final int FEEDBACK_NO = 1;
+    public static final int FEEDBACK_YES = 2;
+
+    public static final int DISMISS_FEEDBACK_NOTIFICATION = 1;
```
barbeau: This should be changed to an Intent action:

`public static final String ACTION_DISMISS_FEEDBACK =
            BuildConfig.APPLICATION_ID + ".action.DISMISS_FEEDBACK";`

onebusaway-android/src/main/AndroidManifest.xml
barbeau: You'll also need your delete intent notification for when the feedback notification gets dismissed listed here:
<action android:name="${applicationId}.action.DISMISS_FEEDBACK"/>

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.d(TAG, "onReceive");
+
+        int notifyId = intent.getIntExtra(NOTIFICATION_ID,
+                NavigationServiceProvider.NOTIFICATION_ID + 1);
+        int actionNum = intent.getIntExtra(ACTION_NUM, 0);
```
barbeau: Instead of checking for an intent extra here, you should be checking for an action:

`String action = intent.getAction();`

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
```diff
@@ -706,4 +716,113 @@ private void silence(long duration, int queueFlag) {
...
+            PendingIntent fdPendingIntentNo = PendingIntent.getActivity(app.getApplicationContext()
+                    , 1, fdIntent,PendingIntent.FLAG_UPDATE_CURRENT);
+
+            fdIntent = new Intent(app.getApplicationContext(), FeedbackActivity.class);
+            fdIntent.setAction(REPLY_ACTION);
+            fdIntent.putExtra("CallingAction", "yes");
+            PendingIntent fdPendingIntentYes = PendingIntent.getActivity(app.getApplicationContext()
+                    , 2, fdIntent, PendingIntent.FLAG_UPDATE_CURRENT);
+
+            rIntent.putExtra(FeedbackReceiver.ACTION_NUM,
```
barbeau: Instead of setting this as an extra it should be the Intent action:
`intent.setAction(...)`

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
```diff
@@ -706,4 +716,113 @@ private void silence(long duration, int queueFlag) {
...
+            PendingIntent pDelIntent = PendingIntent.getBroadcast(app.getApplicationContext(),
+                    0, rIntent, 0);
+
+            mBuilder = new NotificationCompat.Builder(Application.get().getApplicationContext()
+                    ,Application.CHANNEL_DESTINATION_ALERT_ID)
+                    .setSmallIcon(R.drawable.ic_stat_notification)
+                    .setContentTitle(Application.get().getResources()
+                            .getString(R.string.feedback_notify_title))
+                    .setContentText(message)
+                    .addAction(0, "No", fdPendingIntentNo )
```
barbeau: If these "No" and "Yes" values are user-facing in the UI, they need to be in `strings.xml` so they can be translated.

barbeau:
@paulnabanita1 I added a few notes in-line about Intent Actions.  Generally speaking, you should register Intent actions for the BroadcastReceiver in the manifest for the actions you expect to receive from other app components - in our case the notification.  Then you have to set the Intent action to that same string value.  And, when you get an intent in the BroadcastReceiver you should check to make sure the action matches before doing something with that Intent.  In a few places it looks like you're passing extras that should really be the actions, like the action of the response from the user or the action of the user dismissing the notification.  This is important because we need to ensure the intents are filtered according to the applicationId of the app and the correct receiver responds to them.

Everything else that are attributes of the action, like whether it's "yes" or "no" feedback, and the location of the file, can be stored as intent extras.  Keys and values for intent extras can be declared as normal public static final variable Strings because they aren't exposed in the manifest.

Let me know if you have any questions about this.

onebusaway-android/src/main/AndroidManifest.xml
barbeau: Upon further discussion this isn't needed unless we expose the receiver to external apps, so let's just remove the `<intent-filter>` block

onebusaway-android/src/main/AndroidManifest.xml
```diff
@@ -268,6 +276,16 @@
                 <action android:name=".nav.NavigationReceiver" />
             </intent-filter>
         </receiver>
+        <receiver
+            android:name=".nav.FeedbackReceiver"
+            android:enabled="true"
+            android:exported="true" >
```
barbeau: We don't want other apps to be able to see this receiver, so let's remove `android:exported="true"`

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
barbeau: Per further discussion, let's leave these as extras and just rename them from actions to "response", or something similar.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
```diff
@@ -706,4 +716,113 @@ private void silence(long duration, int queueFlag) {
...
+
+        // Create delete intent to set flag for snackbar creation next time the app is opened.
+        Intent rIntent = new Intent(app.getApplicationContext(), FeedbackReceiver.class);
+        rIntent.putExtra(FeedbackReceiver.NOTIFICATION_ID, NOTIFICATION_ID + 1);
+
+        if ((mFirstFeedback) || (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) || (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P)) {
+
+            Intent fdIntent = new Intent(app.getApplicationContext(), FeedbackActivity.class);
+            fdIntent.setAction(REPLY_ACTION);
+            fdIntent.putExtra("CallingAction", "no");
```
barbeau: Per further discussion, let's leave these as extras and just rename them from actions to "response", or something similar.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
barbeau: This is the delete intent, so it should have the action of `ACTION_DISMISS_FEEDBACK`.

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
```diff
@@ -0,0 +1,135 @@
...
+
+        if (actionNum == DISMISS_FEEDBACK_NOTIFICATION) {
+            Log.d(TAG, "Dismiss intent");
+            //TODO : Create Snack bar if user dismissed feedback notification without providing feedback
+        } else {
```
barbeau: Let's add another IF statement here that explicitly checks for the REPLY action.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -2018,4 +2025,28 @@ private boolean isSlidingPanelCollapsed() {
     public ArrivalsListFragment getArrivalsListFragment() {
         return mArrivalsListFragment;
     }
+
+    private void setupLogUploadTask() {
```
barbeau: Let's try moving the setupLogUploadTask() method after the user responses to the feedback with a yes or no answer.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -2018,4 +2025,28 @@ private boolean isSlidingPanelCollapsed() {
...
+
+        // Create the actual work object:
+        PeriodicWorkRequest uploadCheckWork = uploadLogsBuilder.build();
+
+        // Then enqueue the recurring task:
+        WorkManager.getInstance().enqueue(uploadCheckWork);
+    }
+
+    private void setupLogCleanupTask() {
+        PeriodicWorkRequest.Builder cleanupLogsBuilder =
```
barbeau: Let's try moving the setupLogCleanupTask() method after the trip finishes and after we set up the notification that the user can interact with.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationService.java
paulnabanita1: We have moved the user feedback notification code to NavigationService.java so there is no need to access the log file name outside of the service and can be added as an intent extra.

onebusaway-android/src/main/java/org/onebusaway/android/ui/FeedbackActivity.java
paulnabanita1: We have modified the code to reference the values as public static final variables

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
paulnabanita1: Added public static final String ACTION_REPLY to FeedbackReceiver

onebusaway-android/src/main/AndroidManifest.xml
paulnabanita1: We have removed the `<intent-filter>` block from AndroidManifest as we don't need to expose the receiver to other apps.

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
paulnabanita1: Modified `public static final int DISMISS_FEEDBACK_NOTIFICATION` to `ACTION_DISMISS_FEEDBAC`

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
paulnabanita1: We modified the code to set the dismiss action as an action of the intent.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
paulnabanita1: We have created a delete action and modified the code to set it as an action of the intent.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
paulnabanita1: Added "No" and "Yes" values to `strings.xml`.

onebusaway-android/src/main/AndroidManifest.xml
paulnabanita1: Removed `android:exported="true"`.

onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java
paulnabanita1: We have renamed `"CallingAction"` to `"response"` so that we can clearly distinguish between the REPLY and DISMISS actions and the user feedback.

onebusaway-android/src/main/java/org/onebusaway/android/nav/FeedbackReceiver.java
paulnabanita1: We have modified the code to explicitly check for REPLY action.

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
paulnabanita1: We have moved the `setupLogUploadTask()` method after the user responses are processed by `FeedbackActivity` or `FeedbackReceiver`

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
paulnabanita1: We have moved the `setupLogCleanupTask()` to `NavigationService.java` to enqueue the Work Manager after the trip finishes.

barbeau:
From @paulnabanita1 as she was working on merging the current nav branch with this PR, for future reference:

>`Program type already present: com.google.common.util.concurrent.ListenableFuture` was occuring because we have removed the dependency `com.google.firebase:firebase-firestore:17.1.4` from the latest project level gradle. Adding it back fixed the issue. Also, the other issue `FirebaseApp was not initialized with a bucket name` was fixed when I reverted the gradle version to `com.android.tools.build:gradle:3.2.1`. Somehow, the latest version is not able to able firebase reference creation.

barbeau:
Thanks @paulnabanita1, this looks good!

PR #970, opened 2019-04-07
Add York Region Transit (YRT) build flavor

tmsyrt:
This pull request adds the build variant for York Region Transit (https://www.yrt.ca) used for the whitelabel app at https://play.google.com/store/apps/details?id=can.yrt.onebusaway.

All suggested changes  are now implemented.


barbeau:
@tmsyrt Thanks, this looks good!

I made a few more commits in this PR to make it compile right (there was a stray `0` in the build.gradle) and to remove the Google Analytics config files, which are no longer used (we now use Firebase instead).

One thing I think is still missing, though - it looks like you changed the "Visit us on GitHub" navigation drawer text and icon to "Live chat with YRT" (see below).

Could you include this config in the pull request too?

image

barbeau:
Per discussion with @tmsyrt offline, we'll merge this for now and as YRT decides on permanent text for the above we'll merge in future PR.

PR #984, opened 2019-05-27
Fix #982 - Add text datatype handling to fix arrival time problem

### LINKED ISSUE #982: Arrival time Open311 problem reporting is broken in Tampa
**Summary:** 

Reported by a user and I've confirmed - if you try to submit a problem report it gives a toast saying "Required questions should be answered" even if you've filled out all the fields.

**Steps to reproduce:** 

1. Tap on a stop
1. Tap on an arrival time and select "Report an arrival time problem"
1. Fill out all the fields provided

**Expected behavior:** 

Accept the problem report

**Observed behavior:** 

It gives a toast saying "Required questions should be answered" even if you've filled out all the fields

Samsung Galaxy S8+ with Android 9.0

cagryInside:
This PR handles the `text` datatype to fix arrival time problem.  Currently, OBA isn't showing an Open311 attribute if its datatype is text.

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -412,7 +412,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.0.0-alpha3'
     implementation 'commons-io:commons-io:2.4'
     // Open311 client library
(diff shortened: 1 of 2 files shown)


barbeau:
Thanks @cagryInside!!

PR #999, opened 2019-07-11
Add userId as a field of the user's firebase collection

cagryInside:
`UserId` is added as a field of firebase collection.  Now, the data looks like this:
Screen Shot 2019-07-11 at 11 08 24 AM


barbeau:
Thanks @cagryInside!

For the record, we're adding this because we can't query a Firebase document to get a list of subcollections unless the document has at least one field, per Firebase design.

PR #1002, opened 2019-07-14
Add time info to travel behavior activity

cagryInside:
Timestamps are added to transition events.  For each transition event we collect the following timestamps:
(A) -> `TransitEvent.elapsedRealtimeNanos() `,
(B) -> `SystemClock.elapsedRealtimeNanos()`,
(C) -> `System.currentTimeMillis()`, and
(D) -> (C) - (A)  = number of nanoseconds in the past when the event happened.

The data looks like the following in Firebase:

Screen Shot 2019-07-14 at 2 02 19 PM

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/travelbehavior/model/TravelBehaviorInfo.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/travelbehavior/model/TravelBehaviorInfo.java
@@ -28,25 +28,31 @@ public class TravelBehaviorInfo {
         public String detectedActivity;
         public String detectedActivityType;
         public Integer confidenceLevel;
-        public Long transitEventElapsedRealtimeNanos;
+        public Long eventElapsedRealtimeNanos;
         public Long systemClockElapsedRealtimeNanos;
         public Long systemClockCurrentTimeMillis;
         public Long numberOfNanosInThePastWhenEventHappened;
+        public Long eventTimeMillis;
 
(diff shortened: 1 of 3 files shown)


barbeau:
@cagryInside Thanks for implementing this!  So, I made a mistake in the original Google drawing describing this, which I've corrected here:
https://docs.google.com/drawings/d/1AHvO62fClfBOfXmfUGWUltg8hnjkXitohq4FYq1quho/edit?usp=sharing

Here's the correct version:
~~~
(A) -> event.elapsedRealtimeNanos() ,
(B) -> SystemClock.elapsedRealtimeNanos(),
(C) -> System.currentTimeMillis(), and
(D) -> (B) - (A) = number of nanoseconds in the past when the event happened.
(E) -> Clock time when transition happened
~~~

I've changed the implementation to match this (including the new field "(E)" in https://github.com/OneBusAway/onebusaway-android/pull/1002/commits/bc2728149fecd63098cc2be8c837ba7549a8bc3f, but could you please review to confirm I got it right?

I also changed the field name from "transitEventElapsed..." to just "eventElapsed", as it's shorter and a better description of the data (again, my fault from not correctly the text in the diagram).

barbeau:
Sorry, I just rebased it to get rid of other unrelated changes, new commit of mine is https://github.com/OneBusAway/onebusaway-android/pull/1002/commits/f2377e3f96de09c4cdfebe805eb773b51422fd2c

PR #1008, opened 2019-07-18
Fix #1007 - Call save-device-info method at application launch

### LINKED ISSUE #1007: Travel behavior tracking - Duplicate device information records
**Summary:** 
OBA pushes multiple duplicate device-information records to Firebase.

**Steps to reproduce:** 
N/A

**Expected behavior:** 

A device-information record should be unique. 

**Observed behavior:** 

For some users, we have seen duplicate device-information data.

**Screenshots:** 

Screen Shot 2019-07-18 at 2 45 17 PM

cagryInside:
Calling the `saveDeviceInfo` method in the app launch.  Previously, we were calling this method on every transition event and it was causing duplicate ~device-information` records on Firebase.  Now, we only call once when the app is started.  The new data looks like this on Firebase:

Screen Shot 2019-07-18 at 3 21 10 PM


barbeau:
Thanks @cagryInside! How about putting the timestamp as a field too so we can query on it?

cagryInside:
> Thanks @cagryInside! How about putting the timestamp as a field too so we can query on it?

@barbeau good idea.  Just added and updated the commit:

Screen Shot 2019-07-19 at 11 08 12 AM

barbeau:
Thanks @cagryInside!

PR #1013, opened 2019-09-23
Fix #1012 - Trip status text being clipped when using larger font sizes

### LINKED ISSUE #1012: Trip status times get cut off with large font size
**Summary:** 

From a user:

>Love the app overall but I'm partially sighted so I have the font settings in my phone set to a large size. This means if I try a look at the trip status ("show trip status") the timings get cut off (screen shot).
>Is this something that can be fixed?

image

**Steps to reproduce:** 

Tap on arrival time in list, select "Show trip details"

**Expected behavior:** 

For larger font sizes, expand the height of each row so the text isn't cut off

**Observed behavior:** 

The arrival times get cut off

>App Version: 2.5.3
Model: Pixel 2
OS Version: 9 / 2
Google Play Services App: 18.7.19 (100400-262610125)
Google Play Services Library: 12451000
Region/API: Puget Sound (selected automatically)

sausag3:
When using larger font sizes, the arrival time is clipped and not readable because the items are of a fixed size. This changes it so that the item size is based on the size of the text content, the line on the left is then also sized based on the height of that content. This is to fix Issue #1012 

Currently this is how it looks at the largest font size:

Screenshot_1569198114

With this PR, it should now look like this:

Screenshot_1569198041

And at the smallest font size it looks like this:

Screenshot_1569198908

### DIFF
--- a/onebusaway-android/src/main/res/layout/trip_details_listitem.xml
+++ b/onebusaway-android/src/main/res/layout/trip_details_listitem.xml
@@ -17,7 +17,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="fill_parent"
-    android:layout_height="@dimen/trip_details_transit_line_height"
+    android:layout_height="wrap_content"
     android:orientation="horizontal">
 
     <include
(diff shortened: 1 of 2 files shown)


barbeau:
Thanks @sausag3!

Issue #1033, opened 2020-04-14
Support CANCELED trips


barbeau:
**Summary:** 

Amid transit operations with COVID-19, a lot of transit trips need to be canceled due to sick or quarantined drivers. Currently no real-time information is coming from agencies for these trips, so they show up as "scheduled" in the app. Agencies are working to adapt their CAD/AVL systems to be able to share canceled trips via GTFS-realtime (e.g., via TripUpdate.schedule_relationship = CANCELED), but this won't happen quickly due to the operational and software changes needed. But, we can still work towards supporting this in the app so when agencies can provide this data it shows in the app.

Here's where the CANCELED TripUpdate gets set server-side:
https://github.com/camsys/onebusaway-application-modules/blob/unified/onebusaway-transit-data-federation/src/main/java/org/onebusaway/transit_data_federation/impl/realtime/BlockLocationServiceImpl.java#L577

This would be supported via the OBA REST API `tripStatus` element (e.g., that gets populated in the arrivalsAndDeparturesForStop API) via the “status” field:
http://developer.onebusaway.org/modules/onebusaway-application-modules/current/api/where/elements/trip-status.html
 
The OBA REST API on the camsys repo should support the follow statuses (although they aren't deployed to production yet) (see `TransitDataConstants` in v2 branch):
* `default`
* `CANCELED`

On the MTA fork, other status are supported such as detour and deadhead, but I don’t believe they are surfaced in the API.

UI design also needs to be completed for this - possibly a "CANCELED" label over the entire arrival time card, or update the "early/late/on-time" status to show "CANCELED".

**Steps to reproduce:** 

View a canceled trip in the arrivals list

**Expected behavior:** 

Give me some indication that the trip was canceled

**Observed behavior:** 

Trip shows up as scheduled in the app (due to no real-time info from transit agency saying it's canceled)

barbeau:
For the UI, I like the strike-through text proposed for the iPhone UI in https://github.com/OneBusAway/onebusaway-iphone/pull/1475:

![image](https://user-images.githubusercontent.com/928045/79357519-25e11880-7f0e-11ea-8199-1030c74d00b0.png)

barbeau:
For vehicle positions on the map, we probably shouldn't show CANCELED trips at all.

PR #1034, opened 2020-04-15
feat: Support CANCELED trips

barbeau:
TODO:
- [x] Parse and test trip status for "default" and "CANCELED"
- [x] Add CANCELED trips to arrivals list UI - Style A
- [x] Add CANCELED trips to arrivals list UI - Style B
- [x] Hide vehicle positions for CANCELED trips on the map
- [x] Add CANCELED trips to trip status UI
- [x] Add CANCELED trips to ArrivalsListHeader
- [x] Add CANCELED to Legend in Help
- [x] Test with real data - Apparently this is live in Puget Sound [trips-for-route](https://gist.github.com/barbeau/f0d2f2a94736af8156562b472eb768c4) (which we use to show vehicles on the map), but CANCELED trips are omitted entirely from the arrivals-and-departures-for-stop API, which means that the canceled trips will never be visible in the arrivals list UI.

Screenshots:

image

image

Closes https://github.com/OneBusAway/onebusaway-android/issues/1033.


barbeau:
I'm going to go ahead and merge this so we start filtering out canceled trips from the vehicles map view. We won't currently see any canceled trips in the arrivals view as they are currently being filtered out of the results entirely, but the UI will support it if we decide to allow them through in the future.

Issue #1088, opened 2022-06-14
Allow manual sync of travel behavior data


barbeau:
**Summary:** 

After a user opts into travel behavior data collection currently Firebase controls when this data is sync'd to the server. This is normally good as it saves battery life, but there are scenarios such as testing where we'd like to be able to force the sync to the server to happen. Previously it seemed like any network activity would force this sync, but updated versions of Firebase don't always seem to trigger this.

This won't be used by normal users, so we can hide it in the Settings under "Advanced".

It looks like `FirebaseDatabase.purgeOutstandingWrites()` might do the trick:
https://firebase.google.com/docs/reference/android/com/google/firebase/database/FirebaseDatabase#purgeOutstandingWrites()

We could also call `setPersistenceEnabled(true)` to try and ensure that data isn't lost if they aren't persisted to disk before the app closes:
https://firebase.google.com/docs/reference/android/com/google/firebase/database/FirebaseDatabase#public-synchronized-void-setpersistenceenabled-boolean-isenabled

**Steps to reproduce:** 

1. Opt into travel behavior data collection
2. Wait for data to sync or access internet within app

**Expected behavior:** 

Give me an option to manually force the sync if it hasn't happened yet

**Observed behavior:** 

You just have to wait and hope the sync happens.

**Device and Android version:** 

N/A

barbeau:
Ah, no, upon closer look `purgeOutstandingWrites()` definitely isn't want we want - emphasis mine:

>The Firebase Database client automatically queues writes and sends them to the server at the earliest opportunity, depending on network connectivity. In some cases (e.g. offline usage) there may be a large number of writes waiting to be sent. Calling this method will purge all outstanding writes **so they are abandoned.**

>
>All writes will be purged, including transactions and [onDisconnect()](https://firebase.google.com/docs/reference/android/com/google/firebase/database/DatabaseReference#onDisconnect()) writes. **The writes will be rolled back locally**, perhaps triggering events for affected event listeners, and the client will not (re-)send them to the Firebase backend.

So this would result in us losing data.

barbeau:
A Google Firebase engineer in this SO post:
https://stackoverflow.com/a/55873853/937715

...said:

[...]

PR #1089, opened 2022-06-14
chore: Add preference to force push data to Firebase server

barbeau:
Adds a preference in "Settings->Advanced" where the user can force a push of the Firebase data to the Firebase server, if they've enrolled in the travel behavior study. Intended only for use by OBA developers and those testing travel behavior data collection with the app.

Closes https://github.com/OneBusAway/onebusaway-android/issues/1088

Here's the main code that forces a push:

```kotlin
val db = FirebaseFirestore.getInstance()
db.enableNetwork() // This seems to be required to avoid an NPE in Firebase (forces internal check to ensureClientConfigured())
db.waitForPendingWrites().addOnCompleteListener { ... }
```

If I don't call `db.enabledNetwork()` I get an NPE from within Firebase (the internal `client` reference is null, as it doesn't seem to be initialized), but calling `db.enableNetwork()` seems to force initialization and then `db.waitForPendingWrite()` succeeds. I think this is a bug in Firebase and I've filed an issue with Google about it (Case 10184819 - apparently Firebase doesn't have a public issue tracker so I can't link to it).

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -409,7 +409,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-analytics:21.0.0'
     // Cloud Firestore (for storing destination alert test data)
     implementation 'com.google.firebase:firebase-firestore:24.1.2'
-    implementation 'com.google.firebase:firebase-auth:21.0.3'
+    implementation 'com.google.firebase:firebase-auth:21.0.5'
     implementation 'com.google.firebase:firebase-storage:20.0.1'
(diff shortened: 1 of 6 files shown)


barbeau:
This was indeed a bug and firebase and it turns out the SDK is actually open source, so here's the fix:
https://github.com/firebase/firebase-android-sdk/pull/3900/files/228932742f85cb59ac52147302dc79537b2665a2

Will be in future release, so at that point we could remove the workaround mentioned above.

Issue #354, opened 2015-11-10
Add "Starred routes" list


barbeau:
We now allow users to star a route/headsign/stop combination (see https://github.com/OneBusAway/onebusaway-android/issues/155#issuecomment-131239723).  However, currently this is only used to pin favorite routes to the sliding panel header for particular stops.

We should add a list of "starred routes", similar to how we currently have "starred stops".  This could be implemented as another navigation drawer item, or starred stops/routes could become part of the same tabbed interface accessible from the current "starred stops" navigation drawer item (which would be renamed "Starred", or "Favorites".

adingler:
Over the past couple of weeks I have been working on implementing this feature in my spare time, and it is nearly finished.  I am hoping to complete this over the long weekend, time allowing.  However, I have one quick question.  The way I am implementing this is, essentially, to reproduce what was done for starred stops.  So, there is a "Starred routes" item in the navigation drawer that launches a fragment with similar (well, nearly identical) functionality to MyStarredStopsFragment.  The other option, per the original comment above, would be to create a "starred/favorites" item in the navigation drawer.  My question is, is there a preference for one implementation over the other?  If the latter option is preferred, I can work on that before I submit a pull request.

barbeau:
@adingler Awesome, thanks for working on this!  I've been leaning towards a single "Starred/Favorites" entry in the nav drawer, and then a tabbed/swipeable layout so you can easily toggle back and forth.  I think this works better in the mental model than favorites being broken out into two at the top navigation level.  Let me know your thoughts though.

I think there is some UI design TBD around exact layout presentation of route/headsign in the list, and also UI/logic TBD surrounding removing a starred route.  If you have some existing mockups or implementation/thoughts to share on this I'd welcome them.  You might want to take a look at iOS as well just to see how routes/headsigns are shown in a list there.

adingler:
For the UI design for routes in a list, I have the new fragment extending MyRouteListFragmentBase.  This means the starred route list looks the same as the recent routes list.  Here is a screenshot of what the list looks like currently:
![screenshot_1509830844](https://user-images.githubusercontent.com/6656951/32694665-ca3c41da-c6fa-11e7-90e0-fee0fbcdc237.png)

[...]

PR #1097, opened 2023-12-03
Add Starred Routes list (#354)

millanp:
I added a new tab below Starred Stops called Starred Routes.

Screenshot from 2023-11-30 21-16-17

It takes you to a list of routes which are starred at at least one stop.

Screenshot from 2023-12-01 20-54-47

A long press on a route shows a popup menu with various options, including an option to remove all stars from that route.

Screenshot from 2023-11-30 21-18-50

There is also a button in the top right menu to remove all stars from all routes.

Screenshot from 2023-12-01 21-24-00

When the user taps one of the starred routes, it shows the map view for that route. 

Screenshot from 2023-12-01 21-17-20

This is better than the old method of showing a text-based list of stops because the map is much more readable and it provides useful information about the actual location of buses. Most users I have spoken to do not know that the route-specific map view exists, and they say that it is useful when I show it to them. I modified the "recent routes" list to have similar behavior.

I made one change to existing code, factoring out the classes RouteInfoLoader and RouteInfoLoaderCallback from RouteInfoListFragment into QueryUtils. This is because I need to load the long name of a route when it is starred by the user, to make sure that the Starred Routes page is fully populated.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/provider/ObaContract.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/provider/ObaContract.java
@@ -18,6 +18,16 @@
  */
 package org.onebusaway.android.provider;
 
+import com.google.firebase.analytics.FirebaseAnalytics;
+
+import org.onebusaway.android.BuildConfig;
+import org.onebusaway.android.R;
+import org.onebusaway.android.app.Application;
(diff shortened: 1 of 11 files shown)


aaronbrethorst (approved):
this looks great! thanks for building out this feature. I can't wait to get it into the Play Store!

PR #1103, opened 2024-01-09
Fix bug in Starred Routes list

millanp:
Right now the Starred Routes list sometimes omits the route description to the right of route numbers. This is because I previously loaded the "long name" of the buses in a Loader whose lifecycle was tied to that of a fast-closing dialog box. Sometimes the dialog box would close before the request could complete. I fixed it to ensure that the Loader is tied to the long-lasting HomeActivity rather than RouteFavoriteDialogFragment.


aaronbrethorst (approved):
looks great, thank you for taking care of this! :shipit:

Issue #1104, opened 2024-01-13
Dark Mode Override Doesn't 'Stick'


aaronbrethorst:
Repro steps:

Given a light system theme.

1. Change the app theme to dark
2. Kill the app
3. Relaunch the app

Expected: app theme is still dark

Actual: App theme is light 

Customer description:

> My system is in dark mode.  The app override to use light mode anyway isn't working.  I prefer maps in light mode because maps are impossible to read in dark mode in bright outdoor light. Thanks. 

> App Version: 2.11.0
> Model: SM-S911U
> OS Version: 13 / 33
> Google Play Services App: 23.49.14 (190408-590296185)
> Google Play Services Library: 12451000
> Region/API: Puget Sound (selected automatically)

aaronbrethorst:
Fixed in #1105 – thank you @millanp!

PR #1105, opened 2024-01-17
Make the dark mode override 'stick' (#1104)

millanp:
This fix ensures that the map remembers the user's selection of either light or dark mode, even when the app stops and starts.


aaronbrethorst (approved):
nice work! 🚀

PR #1152, opened 2024-03-09
Fixed #895 - Allow saving for routes into the DB

### LINKED ISSUE #895: Recent routes isn't populated as often as it should be

amrhossamdev:
Fixes #895 

### Details

After a couple of hours investigating the issue and the DB and trying to debug the saving into the DB, I found that class `ObaContract` is responsible for the CRUD operations and after reviewing the class code I found that an inner class called `Routes` that was having a function called `InsertOrUpdate` and sure the class responsible for managing `CRUD` operations of recent routes and I found that this method isn't used when showing a route on the map / showing trip statues / searching for a route and of course, saving will not work in this case.

### Fix
I wrote two functions in the `DBUtil` class that will work to save the route when showing it showing trip statues / or clicking on the route in search results. and called this function when all of these operations are used and now it's working perfectly.

### Video showing trying different test cases

[Screencast from 03-09-2024 01:47:30 PM.webm](https://github.com/OneBusAway/onebusaway-android/assets/29693819/e55d8d84-08c8-4dc5-b6ca-1e3c5bd78128)

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
@@ -1,10 +1,15 @@
 package org.onebusaway.android.util;
 
 import org.onebusaway.android.app.Application;
+import org.onebusaway.android.io.elements.ObaRoute;
 import org.onebusaway.android.io.elements.ObaStop;
 import org.onebusaway.android.provider.ObaContract;
+import org.onebusaway.android.ui.ArrivalInfo;
 
 import android.content.ContentValues;
+import android.content.Context;
+
+import com.squareup.okhttp.Route;
 
 /**
  * Created by azizmb9494 on 2/20/16.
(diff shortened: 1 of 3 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
```diff
@@ -25,4 +29,47 @@ public static void addToDB(ObaStop stop) {
...
+        String longName = arrivalInfo.getInfo().getRouteLongName();
+
+        if (TextUtils.isEmpty(longName)) {
+            longName = UIUtils.formatDisplayText(arrivalInfo.getInfo().getHeadsign());
+        }
+
+        routeValues.put(ObaContract.Routes.SHORTNAME, shortName);
+        routeValues.put(ObaContract.Routes.LONGNAME, longName);
+
+        if (Application.get().getCurrentRegion() != null) {
```
aaronbrethorst, line 46: It seems to me that this check should be moved up to the very top of the method as a precondition for everything else. If the current region is `null`, then just return since there's no reason to save a route to the DB; it'll be unusable.

onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
```diff
@@ -25,4 +29,47 @@ public static void addToDB(ObaStop stop) {
...
+            shortName = longName;
+        }
+        if (TextUtils.isEmpty(longName) || shortName.equals(longName)) {
+            longName = route.getDescription();
+        }
+        routeValues.put(ObaContract.Routes.SHORTNAME, shortName);
+        routeValues.put(ObaContract.Routes.LONGNAME, longName);
+        routeValues.put(ObaContract.Routes.URL, route.getUrl());
+
+        if (Application.get().getCurrentRegion() != null) {
```
aaronbrethorst, line 69: same as above.

aaronbrethorst (changes requested):
Excluding some minor refactoring in the database layer, I think this looks great and is ready to go.

onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
amrhossamdev: Thanks for correcting me this makes sense, I have submitted a new commit.

onebusaway-android/src/main/java/org/onebusaway/android/util/DBUtil.java
```diff
@@ -31,6 +31,8 @@ public static void addToDB(ObaStop stop) {
     }
 
     public static void addRouteToDB(Context ctx, ArrivalInfo arrivalInfo){
+        if (Application.get().getCurrentRegion() == null) return;
```
aaronbrethorst, line 34: 👍

aaronbrethorst (approved):
great work!

aaronbrethorst:
I'm going to merge this in PR #1175 because GitHub is complaining about merge conflicts.

image

amrhossamdev:
Thank you !

Issue #791, opened 2017-08-30
Real-time trip planning monitoring occasionally returns empty bundle


barbeau:
**Summary:** 

This is the root cause behind https://github.com/OneBusAway/onebusaway-android/issues/790 which is causing the below NPE:

~~~
java.lang.NullPointerException: 
  at org.onebusaway.android.directions.realtime.RealtimeService.possibleReschedule (RealtimeService.java:124)
  at org.onebusaway.android.directions.realtime.RealtimeService.onHandleIntent (RealtimeService.java:82)
  at android.app.IntentService$ServiceHandler.handleMessage (IntentService.java:67)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.os.HandlerThread.run (HandlerThread.java:61)
~~~

We need to figure out why sometimes the bundle from `TripResultsFragment` that is passed to `RealtimeService` in line 337 - `RealtimeService.start(getActivity(), getArguments());` is empty.

For a quick fix, I'm going to add code to fix https://github.com/OneBusAway/onebusaway-android/issues/790 just to avoid the NPE and drop scheduling the check for real-time directions changes.  But, we need to determine the root cause, which is this issue.

**Steps to reproduce:** 

Not sure - plan a trip so that `RealtimeService.start(getActivity(), getArguments());` is executed in `TripResultsFragment`.

**Expected behavior:** 

Provide the trip objects in the bundle passed to `RealtimeService`

**Observed behavior:** 

It appears the bundle is empty - see https://github.com/OneBusAway/onebusaway-android/issues/790.

**Device and Android version:** 

Wide variety - Android versions 5.1 to 7.0, LG, Samsung, and Pixel.  These are from crash reports in the Android Developer Console.

sahilempire:
Hello @barbeau,

I am interested in investigating and resolving the issue where real-time trip planning monitoring occasionally returns an empty bundle, leading to a NullPointerException in `RealtimeService.java`. I have reviewed the related issues #790 and #791. I plan to:

1. **Identify the Root Cause**: Examine the conditions under which the bundle becomes empty when passed to `RealtimeService`.

2. **Implement a Solution**: Ensure that the bundle contains the necessary trip objects before being processed by `RealtimeService`.

3. **Enhance Error Handling**: Add checks to prevent NullPointerExceptions if the bundle is unexpectedly empty.

[...]

Issue #790, opened 2017-08-28
NPE on real-time directions refresh


barbeau:
**Summary:** 

I'm seeing a large number of NPEs (278 in last week, 2618 total, affecting 1301 users) in the Android Developer Console for OBA with the below stack trace:

~~~
java.lang.NullPointerException: 
  at org.onebusaway.android.directions.realtime.RealtimeService.possibleReschedule (RealtimeService.java:124)
  at org.onebusaway.android.directions.realtime.RealtimeService.onHandleIntent (RealtimeService.java:82)
  at android.app.IntentService$ServiceHandler.handleMessage (IntentService.java:67)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.os.HandlerThread.run (HandlerThread.java:61)
~~~

**Steps to reproduce:** 

Not sure - I'm assuming planning a trip and wait for the real-time refresh to happen

**Expected behavior:** 

Not crash

**Observed behavior:** 

App crashes

**Device and Android version:** 

Wide variety - Android versions 5.1 to 7.0, LG, Samsung, and Pixel.

barbeau:
The root cause behind this is now ticketed in https://github.com/OneBusAway/onebusaway-android/issues/791.

For now, I'm going to add workaround code just to avoid the NPE and drop scheduling the check for real-time directions changes.

PR #1161, opened 2024-03-11
Fixed #1076: Refactored RealtimeService using WorkManager

### LINKED ISSUE #977: Ghost notification "Updating reminders"
**Summary:** 

Reported on Twitter at:
https://twitter.com/MoeFwacky/status/1118593631659495424

>For what feels like months now, I've been getting ghost notifications on my phone. I finally tracked down the cause. OneBusAway "updating reminders." I don't even take the bus, I just kept the app for when I might occasionally need to. I'm relieved, but also so angry about this.... I discovered that it was the cause of the ghost vibrations that had been interrupting me since the Android 8 update...they only seemed to happen at work, right next to a transit center.

Another user replied saying:

>The update about the tunnels was also causing some ghost notifications on this side of the water too. It was quite sometime after the reroutes before I actually got to see it.

I'm not sure what's causing this, but a guess is that the rescheduling Service isn't being canceled or rescheduled correctly in certain conditions, and the Service that reschedules them is repeatedly triggering, perhaps with higher priority than it should have, which could cause vibration and alarm tone.  Alternately, it could be a data problem, with trip_ids changing as a result of the massive Puget Sound data update with the tunnel closure and updates being triggered somewhat randomly because of this.

**EDIT** More info from user - https://twitter.com/gatzby_/status/1120794385300377601:

>No Reminders here. Moto x4, provider: Google Fi (also the software variant), Android 9 with March 1 security update

**Steps to reproduce:** 

Unknown

**Expected behavior:** 

Only provide notifications when the user schedule

**Observed behavior:** 

User is getting "ghost notifications"

Moto x4, provider: Google Fi (also the software variant), Android 9 with March 1 security update

### LINKED ISSUE #1076: IntentService deprecated in Android 11
**Summary:** 

https://developer.android.com/guide/components/services says:

>The Android framework also provides the IntentService subclass of Service that uses a worker thread to handle all of the start requests, one at a time. Using this class is not recommended for new apps as it will not work well starting with Android 8 Oreo, due to the introduction of Background execution limits. Moreover, it's deprecated starting with Android 11. You can use JobIntentService as a replacement for IntentService that is compatible with newer versions of Android.

Maybe this could be related to #977?

**Steps to reproduce:** 

Look at code

**Expected behavior:** 

IntentService is deprecated and we should review alternatives

**Observed behavior:** 

We use IntentService in:
https://github.com/OneBusAway/onebusaway-android/blob/6cd597042d61fe262888b0d8b8e4ed3ede6ecaf9/onebusaway-android/src/main/java/org/onebusaway/android/directions/realtime/RealtimeService.java

Adithya-hv:
## Refactor
RealtimeServices for OTP Trip Results **hasn't been functioning since Android 8.0** (API level 26) due to the background execution limits it introduces. This was fixed by refactoring Intent Service using Work Manager, swapping out WakefulBroadcastReceiver & Alarm Manager with **[WorkManager](https://developer.android.com/jetpack/androidx/releases/work?hl=en) & [Workers](https://developer.android.com/reference/kotlin/androidx/work/Worker?hl=en)** for all persistent work.

## RealTimeService
RealTimeService was refactored to handle all the checks the old service was doing, it then converts and simplifies the bundle into a Data object which is passed into [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) which runs the RealTimeWorker class every 15 minutes. PeriodicWorkRequest is also started with an initial delay, it is only started an hour before the scheduled trip time which is the same as the old RealtimeService. 

## RealTimeWorker
RealTimeWorker is the new [Worker](https://developer.android.com/reference/kotlin/androidx/work/Worker?hl=en) class that only checks for any changes in the itineraries and responds to it accordingly, this logic was not changed and functions the same as the old RealtimeService. It returns the result as success if it was not canceled and failure if it was.

## Empty Bundle
The old RealtimeService checked if the function receives an empty Bundle, and disables RealtimeServices if it's empty. The refactor includes this but it should be removed when issues #790 and #791 are resolved. 

## Core Functionality Change
The only major difference change in the refactor was the number of times the service polls, the old RealtimeService polls the OTP server every minute while the new one polls it every 15 minutes. This is due to a limitation on [PeriodicWorkRequest](https://developer.android.com/reference/androidx/work/PeriodicWorkRequest) whose minimum repeat interval is 15 minutes. [WorkManager](https://developer.android.com/jetpack/androidx/releases/work?hl=en) was used as it was recommended by the issue(#1076) and Android to schedule deferrable, asynchronous tasks that must be run reliably.

## Issues Affected
Closes #1076 
Could affect / close #977 

## Pull Request Requirements


amrhossamdev:
Hi @Adithya-hv, good work!

I have questions polling will check if we have new data every 15 minutes, does this affect real-time changes when the trip status changes like the trip was delayed?

amrhossamdev:
I think also it's not related to updating reminder notifications because it's still shown, we need to investigate that

image

aaronbrethorst:
This feature is going away with #1260

aaronbrethorst:
I have been informed that I am wrong 🙃

amrhossamdev:
Closing due to age.

PR #1168, opened 2024-03-12
Fix : Trip Planner Layout clipped on smaller devices

### LINKED ISSUE #603: Trip planning layout clips top section on Android 2.3 device
**Summary:** 

First mentioned in https://github.com/OneBusAway/onebusaway-android/pull/541#issuecomment-233441124, and partially fixed in https://github.com/OneBusAway/onebusaway-android/pull/541/commits/f9cddd570ba27b65bce52b64466bb6019f2faa2a.

Here's the current layout:

image

It is usable now, although you can't set the arrive/depart by or depart time or date.  Here is what the trip plan results screen looks like now:

image

We should see if we can adjust this further to avoid clipping the top section.

cc @sdjacobs @cagryInside 

**Steps to reproduce:** 

On Android 2.3 device:

1. Set region to Tampa
1. In main navigation menu, tap "Plan a trip"

**Expected behavior:** 

Complete top section, including origin, destination, and trip arrive/depart and time and date should all be visible.

**Observed behavior:** 

Top section is clipped, so arrive/depart and time and date isn't visible.

Samsung Epic 4G from Sprint (SPH-D700 - basically a Galaxy S) w/ Android 2.3.6:

**Screenshots:** 

See above

neeldoshii:
## Fixes
* Fixes #603 
## Testing Instructions
1. Make Sure you have smaller device or emulator of resolution less than 200x440
2. Open the navigation drawer
3. Click on Plan a trip(beta)

## Problems/ Approach
On smaller screen devices, time spinner & image button were getting clipped. I solved it by simplifing the nested relative layouts, removed the redundant VIEW which was used for empty space, this got easily handled by the constraint layout.
Simplified the layout by adding Schedule & Arrival/Departure to know exactly about the spinner. Increased the touch target size.

### After (Longer Device)

- [ ] Run the unit tests with `gradlew connectedObaGoogleDebugAndroidTest` to make sure you didn't break anything

- [ ] If you have multiple commits please combine them into one commit by squashing them for the initial submission of the pull request.  When addressing comments on a pull request, please push a new commit per comment when possible (reviewers will squash and merge using GitHub merge tool)

### DIFF
--- a/onebusaway-android/src/main/res/layout/fragment_trip_plan.xml
+++ b/onebusaway-android/src/main/res/layout/fragment_trip_plan.xml
@@ -12,183 +12,163 @@
      See the License for the specific language governing permissions and
      limitations under the License.
 -->
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical"
-    android:layout_margin="16dp">
+    android:paddingStart="16dp"
+    android:paddingTop="16dp"
+    android:paddingEnd="16dp"
+    android:paddingBottom="16dp">
 
-    <RelativeLayout
-        android:id="@+id/to_and_from_layout"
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
I think this compromises the trip planner experience on the vast majority of devices in order to gain a small benefit on a device class that I am not convinced really gets used any longer. I'm open to the possibility of accepting this change if it doesn't alter the experience on larger phones, but it really needs to be rock solid in that case.

Issue #882, opened 2018-05-29
Add weather overlay to maps view


barbeau:
**Summary:** 

OBA iOS is in the process of a UI update (see https://github.com/OneBusAway/onebusaway-iphone/pull/1303), and in the process @aaronbrethorst set up a nice weather API based on Dark Sky.

Example URL is:
http://onebusaway.co/regions/1/weather.json?lat=47.63671875&lon=-122.6953125

...where `1` is the region ID from the [Regions API](http://regions.onebusaway.org/regions-v3.json).

Details:
* The weather API is using write-through caching with a 30 minute TTL and constraining results to grid squares of approximately 20km on a side
* lat and lon are optional, but they’ll give you better, more accurate data, especially in larger regions
* Planned addition of etag support, so we could set up conditional GET support so that the client only continues the request if appropriate
* I'd also like to cache results locally on Android device with ~10 min TTL to avoid hitting battery too hard.

For UI, I'd like to have a transparent overlay on the maps view, like the below iOS view, but ours should  be at the top of the screen or bottom left. We have My Location and Layers FABs in bottom right corner, and optional zoom control at bottom middle of screen.  IIRC the Maps heading view appears in the upper left when you rotate or tilt map, so that area is off limits too.

![image](https://user-images.githubusercontent.com/928045/40663871-ab2835f0-6327-11e8-913b-91a15332da57.png)

**Steps to reproduce:** 

Go to "Nearby" map view

**Expected behavior:** 

Show me current weather info

**Observed behavior:** 

No weather info is provided

**Device and Android version:** 

N/A

barbeau:
@aaronbrethorst Did the oba.co API for weather change?

http://onebusaway.co/regions/1/weather.json?lat=47.63671875&lon=-122.6953125 is giving me a 404.

Seeing your demo made me want to add it to Android again ;)

aaronbrethorst:
Yes sorry I was hoping to have this documented. It’s moved to its final home, though, which is available through this versioned API: https://onebusaway.co/api/v1/regions.json

The path templates hew to the IETF’s URI template spec, or you can just directly hit the API like so:

https://onebusaway.co/api/v1/regions/0/weather.json

And include lat and lon as query parameters if desired. I promise it won’t move again :)

tuesd4y:
Hi, all!

I was thinking about doing a PR on this issue in the near future, is there anyone working on that right now?

And last but not least: 
[...]

PR #1174, opened 2024-03-14
Added #882 - Weather feature overlay

amrhossamdev:
Added #882.

| Light Theme                                 | Dark Theme                                 | Options                                   |
|:-------------------------------------------:|:------------------------------------------:|:-----------------------------------------:|
| Light | Dark | Option |

# Design

I talked to Mr. Aaron to pick out the design and icons we'd use. We focused on different weather states and I picked up the icons from https://www.amcharts.com/free-animated-svg-weather-icons/. I got 'fog' and 'wind' icons from Google Material Design. To make sure we covered everything, I checked the weather states listed [here](https://pirateweather.net/en/latest/API/#icon) and the credit of all of the icons used added to the about.

## Edge cases

- Out of range -> hide the weather view.
- In range then out of range -> hide the weather view.
- In range -> show the weather view.
- Out of range and go to the region manually -> show the weather view.
- Click on show route on the map will make the arrival header appear -> overlay the weather view with the arrival header.
- Expanding the slide panel -> overlay the weather view

Also, I have added an `option` to change the temperature unit, automatically, and by default the unit will be picked up using the user device's local country setting then he can change the unit from the setting to `Celsius` or `Fahrenheit`

Also, i have made sure that all icons are supported & suitable for the `Dark mode`

# Logic

I have used `OBA Weather API` to get the weather when we are in a valid region.

### But how we will know that we are in a valid region?

There was an interface in `BaseMapFragment` that was responsible for notifying the user about when he will be out of range, I implemented an interface `onValidRegion` in `HomeActivity` that will callback to us from `BaseMapFragment` and tell us if are in a valid region or not, to get the weather and manage the weather view as well.

### But what if we are already in a valid region ?

There is a method called `showStops` in `BaseMapFragment`, that will work if we are in a valid region so I will also callback to `HomeActivity` to notify we are in a valid region to `Show the weather`

I observed that `showStops` is a real-time function that will be invoked a lot of times so I have made sure to call the `OBA Weather API` only when we `Change the region` or `going from invalid region to valid region` to avoid multiple calls.

### Tests

I have written a `IO Test` test for checking that we have for each `Region id` a `Weather Response`

# Video showing trying different test cases.

Video : [Click here](https://drive.google.com/file/d/1MM1C7_y571-3Ni9rWCzkaTPsYo0EScod/view?usp=sharing)

I'm looking for your feedback and I will be happy to listen to it


onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
```diff
@@ -1968,4 +1981,57 @@ private void showIgnoreBatteryOptimizationDialog() {
...
                 .create().show();
     }
+
+    // Getting a callback from BaseMapFragment to check if we are in a valid region or not
+    @Override
+    public void onValidRegion(boolean isValid) {
+        if(isValid){
+            makeWeatherRequest();
+        }else{
+            //
```
aaronbrethorst, line 1991: please delete this empty comment.

aaronbrethorst (approved):
Terrific work, Amr! I can't wait to get this into users' hands!

onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
amrhossamdev: done

PR #1176, opened 2024-03-15
feat: highlight starred stops on map

### LINKED ISSUE #680: Highlight starred stops on map
**Summary:** 

From user:

>It would be neat to highlight starred stops on the map. When I'm scrolling around a crowded section of the map, that would help me orient / locate more quickly.

OBA iOS now has a feature that shows stop icons a different color if they are starred - see https://github.com/OneBusAway/onebusaway-iphone/pull/569 for details - screenshot is below showing yellow-highlighted stops for favorites.

image

We may need to be careful with performance here, as currently all stops in the `StopOverlay` are held in memory as elements returned directly from the OBA REST API.  To check if a particular stop is a starred stop, we need to pull that info from the local database.  If we do this synchronously while the stops are being added to the map, it could cause slow-down, and the map can already be pretty sluggish at high zoom levels due to Google Maps v2 behavior.  We should benchmark to make sure we're not adding to the problem.  An option is to do the marking of favorites asynchronously after the makers are added to the map, and just update the image via `Marker.setIcon()`.

**Steps to reproduce:** 

1. Go to "Nearby" view to show stops on the map, and hover over a stop you have starred

**Expected behavior:** 

Have a different icon than an unstarred stop

**Observed behavior:** 

Starred and unstarred stops look the same on the map

N/A

**Screenshots:** 

See above for implementation on iOS.

sebesti0n:
**fix #680** 

Preview:

https://github.com/OneBusAway/onebusaway-android/assets/114914556/4a243787-62b6-44ea-8ae9-ddc5835d4ec8


amrhossamdev:
I think you can use a golden color with a starred stop, and use the same shape as a green one what do you think?

amrhossamdev:
Closing to outdating.

sebesti0n:
Yes it sounds good
Sorry for late response this month I'm little bit busy.
> I think you can use a golden color with a starred stop, and use the same shape as a green one what do you think?

PR #1298, opened 2025-02-06
Fixed the temprature unit

MRELEC1R0N:

- [ ✅] Apply the `AndroidStyle.xml` style template to your code in Android Studio.

- [✅ ] Run the unit tests with `gradlew connectedObaGoogleDebugAndroidTest` to make sure you didn't break anything

- [✅ ] If you have multiple commits please combine them into one commit by squashing them for the initial submission of the pull request.  When addressing comments on a pull request, please push a new commit per comment when possible (reviewers will squash and merge using GitHub merge tool)

The issue was in getDefaultUserTemp where it was getting users local country code and the temperature fetched by weather API was of stations user was seeing , so to fixed it I passed the country code from the response of weather API

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/weather/WeatherUtils.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/weather/WeatherUtils.java
@@ -31,24 +31,19 @@ public class WeatherUtils {
         String preferredTempUnit = sharedPreferences.getString(app.getString(R.string.preference_key_preferred_temperature_units), automatic);
 
         String temperatureText;
+        String defaultTempUnit = getDefaultUserTemp();
+        boolean isCelsius = defaultTempUnit.equals("C");
 
-        if (preferredTempUnit.equals(automatic)) {
-            String defaultTempUnit = getDefaultUserTemp();
-            if (defaultTempUnit.equals("C")) {
-                double tempInCelsius = convertToCelsius(temp);
-                temperatureText = (int) tempInCelsius + "° C";
-            } else {
-                temperatureText = (int) temp + "° F";
-            }
-        } else {
-            if (preferredTempUnit.equals(app.getString(R.string.celsius))) {
-                double tempInCelsius = convertToCelsius(temp);
-                temperatureText = (int) tempInCelsius + "° C";
-            } else {
-                temperatureText = (int) temp + "° F";
-            }
-        }
+        int convertedTemp = (preferredTempUnit.equals(automatic))
+                ? (int) (isCelsius ? convertToCelsius(temp) : temp)
+                : (preferredTempUnit.equals(app.getString(R.string.celsius)) ? (int) convertToCelsius(temp) : (int) temp);
+
+        String unit = (preferredTempUnit.equals(automatic)) ? defaultTempUnit : (preferredTempUnit.equals(app.getString(R.string.celsius)) ? "C" : "F");
+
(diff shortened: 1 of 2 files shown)


amrhossamdev:
Hi @MRELEC1R0N  thanks for your contribution!

onebusaway-android/src/main/java/org/onebusaway/android/ui/weather/WeatherUtils.java
```diff
@@ -33,7 +33,8 @@ public static void setWeatherTemp(TextView weatherTempTxtView, double temp) {
         String temperatureText;
 
         if (preferredTempUnit.equals(automatic)) {
-            temperatureText = (int) temp + "° " + getDefaultUserTemp();
+            temperatureText = (int) temp + "° " + getDefaultUserTemp(units);
```
amrhossamdev, line 36: The issue arises when the temperature unit is set to be determined automatically based on the country of the local device. To resolve this, we need to ensure that the temperature is correctly converted according to the detected unit.

so you need only to check if we should convert the value to `Celsius` when because by default it's `Fahrenheit`

amrhossamdev (changes requested):
The issue arises when the temperature unit is set to be determined automatically based on the country of the local device. To resolve this, we need to ensure that the temperature is correctly converted according to the detected unit. so you need only to check if we should convert the value to Celsius when because by default it's Fahrenheit

MRELEC1R0N:
Ok, misinterpreted  few things while fixing the issue , i will update this pr will proper fix asap

MRELEC1R0N:
@amrhossamdev i have fixed it as per your instructions , please review it .

onebusaway-android/src/main/java/org/onebusaway/android/ui/weather/WeatherUtils.java
```diff
@@ -33,9 +33,20 @@ public static void setWeatherTemp(TextView weatherTempTxtView, double temp) {
         String temperatureText;
 
         if (preferredTempUnit.equals(automatic)) {
-            temperatureText = (int) temp + "° " + getDefaultUserTemp();
+            String defaultTempUnit = getDefaultUserTemp();
```
amrhossamdev, line 36: For `setWeatherTemp` you can do this:

```Java
        String defaultTempUnit = getDefaultUserTemp();
        boolean isCelsius = defaultTempUnit.equals("C");

        int convertedTemp = (preferredTempUnit.equals(automatic))
                ? (int) (isCelsius ? convertToCelsius(temp) : temp)
                : (preferredTempUnit.equals(app.getString(R.string.celsius)) ? (int) convertToCelsius(temp) : (int) temp);

        String unit = (preferredTempUnit.equals(automatic)) ? defaultTempUnit : (preferredTempUnit.equals(app.getString(R.string.celsius)) ? "C" : "F");

        String temperatureText = convertedTemp + "° " + unit;

        weatherTempTxtView.setText(temperatureText);
```

amrhossamdev (changes requested):
Well done @MRELEC1R0N , please look at my comments thanks !

MRELEC1R0N:
@amrhossamdev Here are the changes asked for in the review.

amrhossamdev:
LGTM @MRELEC1R0N :fireworks:

amrhossamdev (approved):
LGTM @MRELEC1R0N.  I made small changes for spacing but otherwise, well done! :100:

Issue #510, opened 2016-05-24
Support multiple map tile providers


barbeau:
From a user:

>I really like this [app] and use it all the time. One thing that would make it easier, though, would be to make the street names bigger when I make the map bigger. The names are a little small and stay small and sometimes I have to use my magnifying glass to read them. If I could make the names bigger that would be great! 

We have no control over the default Maps API v2 Google map tiles, but we could support multiple map tile providers, including custom Mapbox tiles where we could edit the size of street names, etc, and potentially even add a custom tile for large street names for users with visual impairments.  We're looking at using Mapbox custom tiles for our USF Maps app - more info [here](https://github.com/CUTR-at-USF/usf-mobullity/issues/266).

We also support multiple map tile providers in OTP Android.  PR that added support for another tile provider (Lyrk) is [here](https://github.com/CUTR-at-USF/OpenTripPlanner-for-Android/pull/218/files).  [Here's the code](https://github.com/CUTR-at-USF/OpenTripPlanner-for-Android/blob/master/opentripplanner-android/src/main/java/edu/usf/cutr/opentripplanner/android/fragments/MainFragment.java#L3570) that switches tile overlays.

barbeau:
I asked the user the following:

>It would help us to know how big the street names would need to be for them to be readable to you.  We’d need to balance the text size while still keeping the other map features visible.  Could you share a screenshot of another app we can use as an example of how big the text would need to be for it to be readable to you?  It doesn’t need to be another mapping app, even an email app like Gmail that includes text would be useful.

They responded:

>I usually use "huge" for the text size in Gmail but "normal" is OK too. The street names on the maps are smaller than gmail's "normal".

barbeau:
One of the map tile types can be Google satellite, so this issue is an umbrella issue that includes the functionality discussed in #31.  I'm going to close #31 as a duplicate as we now have more info here than in that issue.  Mapbox and MapQuest OSM tiles are also included in OTP Android, and could be included here as well, which addresses https://github.com/OneBusAway/onebusaway-android/issues/31#issuecomment-8183067.

PR #1337, opened 2025-03-25
Added satellite functionality, updated settings menu, and updated localization terminology

theKimSingh:
- Changed the CheckBoxPreference for 2d -> 3d map into a ListPreference
 - Added satellite view functionality to the map (it is a part of #510)
 - Added localizations for the multiple languages for the ListPreference
 - Ran tests and ran AndroidStyle.xml
image
image

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
@@ -278,6 +278,7 @@ public class BaseMapFragment extends SupportMapFragment
 
         /**
          * Called when a result has been obtained after requesting user location permission.
+         *
          * @param grantResult The grant results for the location permission which is either PackageManager.PERMISSION_GRANTED or PackageManager.PERMISSION_DENIED. Never null.
          */
         void onLocationPermissionResult(int grantResult);
@@ -371,8 +372,8 @@ public class BaseMapFragment extends SupportMapFragment
         uiSettings.setMyLocationButtonEnabled(false);
         // Hide Toolbar
         uiSettings.setMapToolbarEnabled(false);
-        // Check for 3D map mode settings
-        updateMap3DModeSettings();
+        // Check for map mode settings
+        updateMapModeSettings();
         // Instantiate class that holds generic markers to be added by outside classes
         mSimpleMarkerOverlay = new SimpleMarkerOverlay(mMap);
 
@@ -515,7 +516,7 @@ public class BaseMapFragment extends SupportMapFragment
                 controller.notifyMapChanged();
             }
         }
-        updateMap3DModeSettings();
+        updateMapModeSettings();
         super.onResume();
     }
 
@@ -921,6 +922,7 @@ public class BaseMapFragment extends SupportMapFragment
             mMap.animateCamera((CameraUpdateFactory.newLatLngBounds(b, width, height, padding)));
         }
     }
+
     private RegionCallback regionCallback;
 
     public void setRegionCallback(RegionCallback callback) {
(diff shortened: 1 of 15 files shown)


amrhossamdev:
Amazing, well done @theKimSingh! That needs more testing on different devices. I have a question: Does the satellite map view not support 3D buildings? I suggest keeping 3D mode isolated, regardless of the map type.

theKimSingh:
From my understanding and extensive research, 3D view is not available via phone,

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
```diff
@@ -1496,15 +1498,31 @@ public boolean onMarkerClick(Marker marker) {
...
+     * Updates the map settings based on the current state of map mode preference.
      */
-    private void updateMap3DModeSettings() {
-        if(mMap == null) return;
-
-        boolean isEnabled = Application.getPrefs().getBoolean(getString(R.string.preference_key_enable_map_3d_mode), true);
-
-        mMap.getUiSettings().setTiltGesturesEnabled(isEnabled);
-        mMap.setBuildingsEnabled(isEnabled);
+    private void updateMapModeSettings() {
```
amrhossamdev, line 1503: We can refactor that :thinking: 

```Java   
 private void updateMapModeSettings() {
        if (mMap == null) return;

        String normal2D = getString(R.string.preferences_preferred_map_option_normal2d);
        String normal3D = getString(R.string.preferences_preferred_map_option_normal3d);
        String satellite = getString(R.string.preferences_preferred_map_option_satellite);
        
        String mapType = Application.getPrefs().getString(getString(R.string.preference_key_map_mode), normal2D);

        if (mapType.equals(normal2D)) {
            setMapType(GoogleMap.MAP_TYPE_NORMAL, false, false);
        } else if (mapType.equals(normal3D)) {
            if (mMap.getMapType() == GoogleMap.MAP_TYPE_HYBRID) {
                mMap.setMapType(GoogleMap.MAP_TYPE_NORMAL);
            }
            setMapType(GoogleMap.MAP_TYPE_NORMAL, true, true);
        } else if (mapType.equals(satellite)) {
            setMapType(GoogleMap.MAP_TYPE_HYBRID, false, false);
        } else {
            return; // Should never happen
        }

        resetCameraTilt();
    }

    private void setMapType(int type, boolean tiltEnabled, boolean buildingsEnabled) {
        mMap.setMapType(type);
        mMap.getUiSettings().setTiltGesturesEnabled(tiltEnabled);
        mMap.setBuildingsEnabled(buildingsEnabled);
    }

    private void resetCameraTilt() {
        mMap.moveCamera(CameraUpdateFactory.newCameraPosition(
                new CameraPosition.Builder()
                        .target(mMap.getCameraPosition().target)
                        .zoom(mMap.getCameraPosition().zoom)
                        .tilt(0)
                        .build()
        ));
    }
`

File: onebusaway-android/src/main/java/org/onebusaway/android/ui/SituationDialogFragment.java [hunk 2]
```diff
@@ -166,12 +166,8 @@ public void onClick(DialogInterface dialog, int which) {
 
         String desc = args.getString(DESCRIPTION);
 
-        if (descTxtView != null) {
-            if (!TextUtils.isEmpty(desc)) {
-                descTxtView.setText(Html.fromHtml(desc));
-            } else {
-                descTxtView.setText(R.string.no_description_available);
-            }
+        if (descTxtView != null &&!TextUtils.isEmpty(desc)) {
```
amrhossamdev, line 169: Could you explain why this change was made? Was it a mistake?

amrhossamdev (changes requested):
Great job, @theKimSingh! 👋 Please take a look at my comments.

onebusaway-android/src/main/java/org/onebusaway/android/ui/SituationDialogFragment.java
theKimSingh: This one was weird. I’m pretty sure the change occurred when I tried running the AndroidStyle.xml over the file to make sure the code followed AOSP Code Style Guidelines, and since it was the same functionally, I decided to leave it. I can remove this if needed, though.

onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/BaseMapFragment.java
theKimSingh: Ah, good point! I’ll push a commit fixing this.

onebusaway-android/src/main/java/org/onebusaway/android/ui/SituationDialogFragment.java
amrhossamdev: Yes, please ensure it is restored to match the main repository, as it also changed some logic.

theKimSingh: Oh, I just noticed what you meant. I'll push another commit to fix this!

amrhossamdev (approved):
Well done :fire: @theKimSingh

PR #1346, opened 2025-03-28
Add information about the number of cars on Link Light Rail trains

millanp:
Adds a badge next to Link Light Rail arrival times indicating the number of cars on the arriving train. Looks like this:

lightrailbadgecars


amrhossamdev:
Good work @millanp... I was testing it with a large font. Can you figure out how to make that responsive :thinking: 
image

amrhossamdev (changes requested):
Well done, @millanp... I was testing it with a large font. Can you figure out how to make that responsive 🤔

image

millanp:
Thanks for getting back so quickly! I just made the "4 car" badge much smaller, and this seems to get rid of the problem with large font size (?) I thought a bit about making it responsive, but I'm not sure if it's worth the effort, given that most of the existing UI elements are not responsive and given that statically shrinking it seems to solve the problem. Let me know your thoughts.

barbeau:
This is cool! Note that this probably won't work outside of Seattle (and maybe just Link Light Rail) because the vehicle ID format isn't standardized.

millanp:
Interesting, I didn't know that this wasn't standardized! This vehicle ID format has been stable for quite some time for the link light rail, and I haven't seen it in any other transit systems or routes. Do you think this is a blocker for displaying this info on the app?

barbeau:
>Do you think this is a blocker for displaying this info on the app?

That's up to @amrhossamdev 🙂

millanp:
My case for including this is that it works right now across all transit systems and routes: it either displays an accurate car count, or it doesn't display anything. The format of vehicle IDs has been stable for a long time, certainly for over a year. 

If something breaks it is easy to turn this feature off, and the worst that can happen is briefly displaying an inaccurate car count. 

Thanks again for looking at this!

amrhossamdev:
Do you guarantee that I won’t show up if we don’t have a stable vehicle ID, or will it cause a problem? :thinking: @millanp

amrhossamdev:
For responsive, I think it's looking great now well done @millanp :clap: 

image

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaArrivalInfo.java
```diff
@@ -347,4 +347,19 @@ public Occupancy getHistoricalOccupancy() {
     public Occupancy getOccupancyStatus() {
         return Occupancy.fromString(occupancyStatus);
     }
+
+    /**
+     * @return a String indicating the number of cars, if this is a light rail arrival.
+     *         Otherwise returns null.
+     */
+    public String getNumCars() {
+        assert this.getVehicleId() != null;
```
amrhossamdev, line 356: if(this.getVehicleId() == null) return null;

onebusaway-android/src/main/java/org/onebusaway/android/io/elements/ObaArrivalInfo.java
```diff
@@ -347,4 +347,19 @@ public Occupancy getHistoricalOccupancy() {
...
+        String[] splitByBracket = this.getVehicleId().split("\\[");
+        if (splitByBracket.length == 2) {
+            int numCars = splitByBracket[1].split("-").length;
+            return numCars + " " + "car";
```
amrhossamdev, line 360: Please ensure the string `car` is translated, and use `cars` when numCars > 1.

onebusaway-android/src/main/java/org/onebusaway/android/ui/ArrivalsListAdapterStyleA.java
```diff
@@ -103,6 +104,17 @@ protected void initView(View view, ArrivalInfo stopInfo) {
...
         status.setText(stopInfo.getStatusText());
 
+
+        String numCars = stopInfo.getInfo().getNumCars();
+        if (numCars != null) {
+            carCount.setBackgroundResource(R.drawable.round_corners_style_b_status);
+            carCount.setText(stopInfo.getInfo().getNumCars());
+            carCount.setVisibility(View.VISIBLE);
+        } else {
+            carCount.setVisibility(View.GONE);
```
amrhossamdev, line 114: I think you can make it by default `gone` and when we have numCars just populate it :thinking:

amrhossamdev (changes requested):
Well done @millanp :clap:  please check my comments.

amrhossamdev:
Thanks for your participation, @barbeau! Could you elaborate more on the different formats of vehicle IDs? Are you suggesting that we need to find a way to parse each format differently based on the region? :thinking: Thanks!

barbeau:
>Could you elaborate more on the different formats of vehicle IDs?

AFAIK there is typically no defined format, assuming these come from GTFS Realtime data sources. IDs could be `1234`, a UUID, or passed through from some internal data source from the agency. So you'd need to investigate each agency/AVL vendor to see if they use some kind of internal proprietary format that happens to have the number of cars embedded similar to Link Light Rail.

millanp:
Another option is I could explicitly turn this off for everything but the link light rail

amrhossamdev:
@millanp Looks good to me! Would love to hear your thoughts as well, @aaronbrethorst.

millanp:
Forgot to say this earlier, but one thing I find cool is that no other app gives this information! This would be useful data available only on OneBusAway.

aaronbrethorst:
I agree with @amrhossamdev's feedback, and I like your approach, @millanp, regarding "Another option is I could explicitly turn this off for everything but the link light rail". I'd like to see the following changes before merging:

1. Implement all of the code review feedback that Amr offered
2. Only enable this feature when the region is Puget Sound and the agency in question is `40` (I think you should have access to that information...If not, then just check for matching region)

thanks!

johnhelmuth-st:
Rather than relying on parsing an ID whose format is not included in any specification and may change at any time,  I might recommend using the experimental GTFS-RT VehiclePosition's `multi_carriage_details` property, as a more reliable way to get the number of cars on a train.

Sound Transit publishes its light rail GTFS RT VehiclePosition feed via the Puget Sound region's OneBusAway server, which does not support the `multi_carriage_details` property, as far as I can tell. However, Sound Transit's upstream CAD/AVL system for their light rail service does produce the `multi_carriage_details` property, in preparation for including congestion information per-car. (ST's APC system still needs to be wired into that feed for the congestion information to be included.)

A way forward for this feature is to add support to the server's OBA APIs to include the `multi_carriage` property, and then use that, instead of the vehicle ID, to count the number of cars in the train.

aaronbrethorst:
@johnhelmuth-st If the OBA REST API server added support for ingesting and publishing this data, would ST update their deployed OBA server version to include that?

johnhelmuth-st:
I am checking with the vendor that supports the OBA service for us.

johnhelmuth-st:
@aaronbrethorst  We will get the code deployed to pass through the `multiple_carriage_details` GTFS RT feeds for Trip Updates and Vehicle Positions once it is available.

amrhossamdev:
@millanp any updates?

millanp:
Should I proceed with Aaron's idea of activating the feature only for region 40?

amrhossamdev:
@aaronbrethorst Hi, Aaron. Can we have your opinion on that?

aaronbrethorst:
Yeah, go for it.

millanp:
Sorry for the delay here! I addressed the code review comments and restricted the feature to the Puget Sound region. I couldn't find an easy way to extract the agency ID from the arrival info, but hopefully this is OK. Let me know if there are more issues.

onebusaway-android/src/main/res/values-pl/strings.xml
```diff
@@ -170,6 +170,14 @@
             minut wcześniej
         </item>
     </plurals>
+    <plurals name="stop_info_car_count" tools:ignore="MissingQuantity">
```
amrhossamdev, line 173: Why is there `MissingQuantity`?

amrhossamdev (approved):
Well done @millanp :clap: please look at my comments before merging.

onebusaway-android/src/main/res/values-pl/strings.xml
millanp: This follows the pattern of all of the previous `<plurals>` blocks. It seems like the Polish translations in OneBusAway are all somewhat flawed. Not sure if this is worth correcting in this pull request.

millanp:
I don't think I can merge it myself, but I'm not sure. Also @aaronbrethorst @amrhossamdev do you think we can get this out to the Play Store soon? I'm excited to see it in the wild!

amrhossamdev:
@millanp it will be uploaded to the beta channel soon :rocket:

PR #1405, opened 2025-04-05
Fix the notification channel description typo in createNotificationChannels(). Channel3's description is set on channel2. ALSO, Used UTF-8 charset when converting the custom API URL to bytes in reportAnalytics() to avoid charset issues.

### LINKED ISSUE #1407: Notification channel description typo in createNotificationChannels(). Channel3's description is set on channel2
Here, channel3 is created, but the description is set to channel2. That's a typo. So channel3 would have no description, and channel2's description would be overwritten. This would cause the destination alerts notifications to have an incorrect or missing description, and the arrival reminders might show the wrong description.

Das-R10:
1)In the method createNotificationChannels() the original code accidentally set the description for the destination alerts channel (channel3) on channel2. This fix ensures the description is properly assigned to channel3, making the notification channel's purpose clear to users in their device settings.

2)In the reportAnalytics() method, there's a call to digest.update(getCustomApiUrl().getBytes()). Using getBytes() without specifying a charset can lead to inconsistent behavior on different devices. It's better to use StandardCharsets.UTF_8.

Fixes #1407 

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -70,6 +70,7 @@ import edu.usf.cutr.open311client.models.Open311Option;
 
 import static com.google.android.gms.location.LocationServices.getFusedLocationProviderClient;
 import static org.onebusaway.android.util.UIUtils.setAppTheme;
+import java.nio.charset.StandardCharsets;
 
 public class Application extends MultiDexApplication {
 
@@ -614,7 +615,7 @@ public class Application extends MultiDexApplication {
             MessageDigest digest = null;
             try {
                 digest = MessageDigest.getInstance("SHA-1");
-                digest.update(getCustomApiUrl().getBytes());
+                digest.update(getCustomApiUrl().getBytes(StandardCharsets.UTF_8));
                 customUrl = getString(R.string.analytics_label_custom_url) +
                         ": " + getHex(digest.digest());
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
```diff
@@ -614,7 +615,7 @@ private void reportAnalytics() {
             MessageDigest digest = null;
             try {
                 digest = MessageDigest.getInstance("SHA-1");
-                digest.update(getCustomApiUrl().getBytes());
+                digest.update(getCustomApiUrl().getBytes(StandardCharsets.UTF_8));
```
Das-R10, line 618: Used StandardCharsets.UTF-8 charset when converting the custom API URL to bytes in reportAnalytics() to avoid charset issues. #

amrhossamdev: Can you explain possible issues that might happen ?

amrhossamdev: @Das-R10

amrhossamdev:
waiting for your replay @Das-R10

Das-R10:
> waiting for your replay @Das-R10

> When we use getBytes() without specifying a charset, it relies on the platform's default charset.This can lead to different hash values when generating analytics data, making it unreliable. That's why I suggested to use UTF-8.

@amrhossamdev

amrhossamdev (approved):
Well done @Das-R10

PR #1435, opened 2025-04-15
Fix situation dialog incorrect parsing

### LINKED ISSUE #1434: Preserve Line Breaks in Alerts
When viewing an alert with line breaks, the app does not show the line breaks and instead places all of the text into a single paragraph.

Actual alert as entered:

Image

Image

Alert as shown in app:

Image

Image

amrhossamdev:
Fixes #1434

## UI

image

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/SituationDialogFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/SituationDialogFragment.java
@@ -168,7 +168,8 @@ public class SituationDialogFragment extends DialogFragment {
 
         if (descTxtView != null) {
             if (!TextUtils.isEmpty(desc)) {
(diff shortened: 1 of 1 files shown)


amrhossamdev:
Hi, @devinbraun. I have fixed the problem. Can you verify that the message is correctly populated? I think it's updated!

devinbraun:
HI @amrhossamdev, I'm not sure how to test except from a downloaded version from the Play Store.

devinbraun:
The screenshot looks right.  Can you show a screenshot from the detour for stop ID 11180 which is the "Rt. 43 nightly detour" from my original screenshot for the bug report?

amrhossamdev:
@devinbraun 
image
image

devinbraun:
Looks great, thank you!

aaronbrethorst (approved):
looks good 👍

Issue #1463, opened 2026-01-21
Update deprecated support libraries to androidX versions


theKimSingh:
The android app seems to run on deprecated com.android.support libraries that should be upgraded to their androidX variants (for stability/compatibility reasons). There are 3 deprecated dependencies that need to be updated. The first 2 can be upgraded easily to their newer variants:

- [ ] android.arch.lifecycle:common-java8:1.1.1
- [ ] android.arch.lifecycle:extensions:1.1.1

However, these following libraries need manual updating and building of their dependencies:

- [ ] com.github.OneBusAway:android-fab:2.4.1 
- [ ] com.github.amlcurran.showcaseview:library:5.4.3
- [ ] com.sothree.slidinguppanel:library:3.3.0 

The last 2 libraries either need to be switched out to libraries with newer dependencies or forked and updated manually.

<img width="913" height="235" alt="Image" src="https://github.com/user-attachments/assets/f0ce6663-ad79-4f2f-94b0-bc89d29e3133" />

aaronbrethorst:
@theKimSingh don't use the "Fix #XYZ" tag in a PR if you don't want to close an issue.

PR #1473, opened 2026-02-18
Updated SlidingUpPanel to maintained version (Fix: #1463)

theKimSingh:
As referenced in (#1463)

### DIFF
--- a/onebusaway-android/proguard-project.txt
+++ b/onebusaway-android/proguard-project.txt
@@ -62,7 +62,7 @@
 # Keep Open311 library (needed for unit tests)
 -keep class edu.usf.cutr.open311client.** { *; }
 
-# for AndroidSlidingUpPanel and compiling with API 28 (see https://github.com/OneBusAway/onebusaway-android/issues/944#issuecomment-446340531)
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Kim, thanks for the quick follow-up on these changes.

## Previous Review Items

### Addressed

1. **`setOverlayContent()` migration** — Updated from the deprecated `setOverlayed(true)` to `setOverlayContent(true)`. Looks great.

2. **Proguard comment** — Updated to reference the new `com.github.hannesa2:AndroidSlidingUpPanel` coordinates.

### Accepted As-Is

3. **Library version (4.6.1 vs 4.7.1)** — Your explanation is reasonable. I dug into the 4.7.x releases and can see that between 4.6.1 and 4.7.1, the library was entirely converted from Java to Kotlin and its compile target was bumped to Java 17 — so there are legitimate compatibility considerations beyond just the Gradle wrapper version. Staying on 4.6.1 achieves the primary goal of migrating to the maintained fork, and a future version bump can be a separate effort if needed.

Issue #1479, opened 2026-02-25
Move from AppCompat theming to MaterialComponents


theKimSingh:
We need to migrate over the themes in order to help update the app to a modern state and fix numerous visual glitches

PR #1485, opened 2026-02-28
Updated to MaterialComponents, added proper dark theme

### LINKED ISSUE #1482: ActionBar Tabs are deprecated
ActionBar navigation mode has been deprecated and has issues with theming. See https://github.com/OneBusAway/onebusaway-android/blob/main/onebusaway-android/src/main/java/org/onebusaway/android/ui/MyRoutesActivity.java. Should be a simple fix for an intermediate programmer.

theKimSingh:
Added related theme colors. Removed "forceDarkAllowed" heuristic. Updated top bar padding to match theme colors. (https://github.com/OneBusAway/onebusaway-android/issues/1479). Was forced to fix #1482, #1492 as well as part of this fix. 

Tested for custom theming as well


aaronbrethorst (changes requested):
Hey Kim, thanks for taking on the MaterialComponents migration and dark theme support. This is substantial, meaningful work -- replacing the `forceDarkAllowed` heuristic with proper `values-night` resources is the right approach, and the ViewPager2 migration for the recent stops/routes tabs is a nice modernization. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Build-breaking XML syntax error in `load_more_arrivals_button.xml`**: There's a `// TOODS` line after the closing `</LinearLayout>` tag. `//` is not valid XML comment syntax, and content after the root element will cause AAPT2 to fail the build. Please remove this line entirely.

2. **White-label brands agencyX and agencyY are broken by the color rename**: The PR renames `theme_primary` to `theme_primary_light` and makes `theme_primary` an alias. Several colors in `values/colors.xml` now reference `@color/theme_primary_light` directly (e.g., `navdrawer_text_color_selected`, `navdrawer_icon_tint_selected`, `header_stop_info_ontime`, `stop_info_ontime`). However, agencyX and agencyY only override `theme_primary` -- they don't define `theme_primary_light`. This means those accent colors will fall through to OBA's green `#4CAF50` instead of the brand's red/blue. Either add `theme_primary_light` overrides to agencyX/agencyY, or keep the shared color references pointing at `@color/theme_primary` and only use `theme_primary_light` where it's specifically needed for light-mode contexts.

3. **`restoreDefaultTab()` is never called in the new `MyRecentStopsAndRoutesActivity.onCreate()`**: The old code called `restoreDefaultTab()` at the end of `onCreate()` to restore the user's previously selected tab. The new code sets up ViewPager2 and TabLayout but never calls it. The method has been properly updated to work with ViewPager2 positions, but it's now dead code. Add `restoreDefaultTab();` after the `TabLayoutMediator.attach()` call.

4. **`colorOnSurface` set to `@color/theme_primary` in the light theme will turn all body text green**: In `values/themes.xml`, `colorOnSurface` is set to the brand's primary color (green `#4CAF50` for OBA). In Material Components, `colorOnSurface` controls the default color for text and icons on surfaces -- setting it to green will make TextViews, input labels, hint text, and many widget defaults render as green text. The dark theme correctly sets this to `@android:color/white`. Change the light theme's `colorOnSurface` to a standard dark text color like `@color/body_text_1`, or remove it to use the MaterialComponents default.

## Important

5. **Debug/TODO markers left in committed code**: `styles.xml` has `<!-- Form styles TOODS -->` on line 73, a commented-out style property on line 75, and `<!--  // STODO: -->` on line 111. These indicate unfinished work. The `FormBackground` style is now empty (its only item is commented out), which may change appearance for any views using it. Please resolve these -- either finalize the style or remove it.

6. **Inconsistent alert dialog theme parents between light and dark**: The light `customAlertDialogTheme` inherits from `Theme.AppCompat.Light.Dialog.Alert` while the dark version inherits from `ThemeOverlay.MaterialComponents.MaterialAlertDialog`. This means light-mode dialogs get AppCompat styling (with a hardcoded white background) while dark-mode dialogs get Material styling. Update the light-mode dialog to also use a MaterialComponents parent for consistency.

7. **Inconsistent theme parents between light and dark**: The light theme uses `Theme.MaterialComponents.Light.DarkActionBar` while the dark uses `Theme.MaterialComponents.DayNight.DarkActionBar`. Using `DayNight` in `values-night/` works, but the standard pattern is to use `DayNight` in both and let the resource qualifier handle switching. This also means the two theme files have divergent attribute sets (the dark theme defines `actionBarStyle` and other items the light theme doesn't), creating maintenance burden.

8. **Guard `getSupportActionBar()` against null**: In `MyRecentStopsAndRoutesActivity.java`, `HomeActivity.java`, and `TripPlanActivity.java`, `getSupportActionBar().setElevation(0)` is called without a null check. `getSupportActionBar()` can return null. `ReportActivity.java` in the same PR already does this correctly with `if (getSupportActionBar() != null)`. Please apply the same pattern consistently.

## Fit and Finish

9. **Unused imports**: `MyRecentStopsAndRoutesActivity.java` has four unused imports left from the old ActionBar code (`Resources`, `ActionBar`, `Toolbar`, `ContextCompat`). `ListFragment.java` has an unused `RelativeLayout` import.

10. **Empty method overrides should be removed**: After removing the `setBackgroundColor` calls, `MyListFragmentBase.onViewCreated()` and `MyRemindersFragment.onViewCreated()` are now empty methods that only call `super`. These should be removed entirely.

11. **Missing newlines at end of file**: Several XML files are missing trailing newlines (`activity_recent_stops_routes.xml`, `values-night/themes.xml`, `navdrawer_item_selectable.xml`, `settings_toolbar.xml`, `simple_list_item.xml`, `load_more_arrivals_button.xml`).

12. **`setEmptyText()` layout logic in `ListFragment.java` could be simpler**: The `instanceof` check chain for `FrameLayout.LayoutParams` vs `LinearLayout.LayoutParams` is overly cautious -- `mStandardEmptyView` is always in a `FrameLayout`. And computing pixel widths from screen width doesn't respond to configuration changes. Consider using padding or margins in `onCreateView()` instead of recalculating on every `setEmptyText()` call.

13. **Duplicate night-mode color entries**: `values-night/colors.xml` redefines `route_line_color_default` and `route_text_color_default` with values identical to the light-mode defaults. If they don't need to differ, remove them to reduce maintenance overhead.

Since this PR has conflicts with the base branch, you'll need to rebase against `main` before we can merge. If you're not comfortable with rebasing, this blog post might help: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

Thanks again, and I look forward to merging this change.

PR #1488, opened 2026-03-03
Fix: Real-time updates now start correctly when trip time is missing

### LINKED ISSUE #1486: Real-time trip monitoring fails silently when trip start time is missing
## Summary
The real-time trip monitoring service stops working without warning if it cannot determine the trip's start time, leaving the user without arrival or delay notifications.

## Steps to reproduce
1. Plan a multi-modal trip in the app.
2. Select an itinerary to start real-time monitoring.
3. If the itinerary data is incomplete (e.g., missing a valid `DATE_TIME` in the parameters), the service will encounter the bug during its initialization check.

## Expected behavior
The app should either successfully schedule a background task to monitor the trip or, if the data is invalid, log an error and attempt to recover the information or notify the user that monitoring could not be started.

## Observed behavior
When the "Start Time" for a trip is missing from the request parameters, the `RealtimeService` incorrectly reports that the update has been "rescheduled" for a future time. In reality, no future update is scheduled, and the current update cycle is abandoned. This results in:
* No background monitoring of the bus/train status.
* No arrival or departure notifications.
* No error messages or indication to the user that anything is wrong.

## Device and Android version
*(N/A - This is a logic bug identified in the source code analysis)*

## Possible fixes
The issue is located in `RealtimeService.java` (around line 138-140).

In `rescheduleRealtimeUpdates(Bundle bundle)`:
```java
if (start == null) {
    // To avoid NPE, return true to say that it's been rescheduled, 
    // but don't actually reschedule it
    return true; 
}
```

Returning `true` here causes the calling code in `onHandleIntent` to skip starting the real-time updates, effectively killing the monitoring process. The service should instead return `false` or handle the recovery of the missing trip data.

Amit-Matth:
## Description
The app was occasionally failing to start real-time updates for bus trips if certain timing information was missing. Instead of just giving up, the app should start checking for updates right away. This PR fixes #1486 .

## Fix
- Updated the real-time service logic so it doesn't quietly stop when a trip's start time isn't found.
- Added a new automated test to verify the fix.
- Standardized the copyright headers in the new test file to match the project's style.

## Why is this needed?
Users were missing out on real-time arrival info when the trip data wasn't perfect. This change makes the app more reliable by ensuring monitoring starts even if some metadata is missing.

## How I tested it
- I ran the project's unit tests using Gradle.
- I verified that the new test case specifically covering this issue passes successfully.

### Checklist

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/directions/realtime/RealtimeServiceTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/directions/realtime/RealtimeServiceTest.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2026 Amit-Matth
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.onebusaway.android.directions.realtime;
+
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.onebusaway.android.directions.util.OTPConstants;
+import org.onebusaway.android.directions.util.TripRequestBuilder;
+import org.opentripplanner.api.model.Itinerary;
+
+import java.util.ArrayList;
+import java.util.Date;
+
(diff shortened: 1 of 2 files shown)


Amit-Matth:
Hey @aaronbrethorst, thanks for the detailed and helpful review! I’ve addressed all the critical and "fit and finish" points you raised:
1. **NPE Risk**: I've updated `getItinerary(Bundle)` to be null-safe and added a guard in `onHandleIntent` to ensure `startRealtimeUpdates` is only called if a valid itinerary is found.  
2. **Diagnostic Logging**: Added the suggested `Log.w` to the `else` block in `rescheduleRealtimeUpdates`, which will provide bundle keys to help debug #790 and #791 in the future.  
3. **Copyright Header**: Fixed the header in `RealtimeServiceTest.java` to match the `2016 Cambridge Systematics, Inc.` attribution.  
4. **End-to-End Safety Test**: Added `testOnHandleIntentWithEmptyBundleDoesNotCrash()` to verify that the fix prevents the NullPointerException in a real scenario.  
5. **Test Base Class**: Cleaned up `RealtimeServiceTest` by removing the `ObaTestCase` dependency and switching to a standard `AndroidJUnit4` runner.

Issue #1483, opened 2026-02-26
Upgrade app to use Material 3


theKimSingh:
I believe a material 3 upgrade could be amazing for the app, and draw in new users. Aside from deprecated components (see #1481), the biggest thing that is holding the app back is the theming, as the current version of the app uses AppCompat theming and custom themes may break. Regardless, I have created a working version of the app here: https://github.com/theKimSingh/onebusaway-android/tree/viable-m3 using colors provided by Google's Material Theme Builder.

Issue #1240, opened 2024-09-05
Disable or remove travel behavior features


aaronbrethorst:
The Travel Behavior functionality is causing us trouble with Google—they are rejecting app updates because of this (unused) feature. Let's try to disable the feature, and, if that proves insufficient, archive the code in a branch and delete it from the project.

<img width="695" alt="image" src="https://github.com/user-attachments/assets/f920280c-611c-4dbb-bfc4-0e614092ca45">

PR #1496, opened 2026-03-07
created m3 app, adding theming details 

theKimSingh:
Fix (#1483). 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 .DS_Store
 /build/
 settings.local.json
+.kotlin/
--- a/REBRANDING.md
+++ b/REBRANDING.md
@@ -87,11 +87,11 @@ src/newBrandName/
     │   ├── strings.xml          # App name (required)
     │   ├── colors.xml           # Theme colors (required)
     │   └── do_not_translate.xml # API keys, URLs (required)
-    ├── values-es/               # Optional: localized app name
-    │   └── strings.xml
     └── mipmap-*/
-        ├── ic_launcher.png
-        └── ic_launcher_round.png
+    │   ├── ic_launcher.png      # App icon (all densities)
+    │   └── ic_launcher_round.png
+    └── values-night/
+        └── colors.xml          # Dark Theme colors (required)
 ```
 
 ### Step 3: Configure Resources
@@ -124,6 +124,8 @@ All other branded strings use `%1$s` placeholders that automatically substitute 
     <color name="stop_info_ontime">#4CAF50</color>
 </resources>
 ```
+
+To override material 3 colors, please checkout the [theme builder](https://material-foundation.github.io/material-theme-builder).
 
 #### do_not_translate.xml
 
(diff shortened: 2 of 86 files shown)


aaronbrethorst (changes requested):
Kim, thanks for pushing through the feedback and force-pushing the fixes — the time picker crash, 24-hour format, string localization, and AlertDialog import issues are all resolved. The M3 migration is shaping up well. There's one new bug to fix and a few items from the first round that still need attention.

## Critical

### 1. `setColorFilter` called with resource ID instead of resolved color

**`ArrivalsListHeader.java:399`** — `mExpandCollapse.setColorFilter(R.color.body_text_1)` passes a resource ID (an integer like `0x7f060XXX`) directly to `setColorFilter(int)`, which interprets it as a raw ARGB color value. This will produce an incorrect tint at runtime — likely a semi-transparent blue/green instead of the intended body text color.

The very next line does it correctly:

```java
mAlertView.setColorFilter(mView.getResources().getColor(R.color.header_text_color));
```

Fix:

```java
mExpandCollapse.setColorFilter(mView.getResources().getColor(R.color.body_text_1));
```

Or use `ContextCompat.getColor()` to be consistent with modern API usage.

## Important

### 2. FQCNs used inline in PreferencesActivity (from prior review)

**`PreferencesActivity.java:332-336`** — The `onPostCreate` surface color block uses three fully-qualified class names inline:

```java
int surfaceColor = com.google.android.material.color.MaterialColors.getColor(
        this,
        com.google.android.material.R.attr.colorSurface,
        android.graphics.Color.WHITE
);
```

Add imports for `MaterialColors` and `Color` and use short names. `MaterialAlertDialogBuilder` from the same package family is already imported in this file, so the inconsistency stands out.

### 3. Dead code modifications add diff noise (from prior review)

Two files modify commented-out or dead code without functional benefit:

- **`PreferencesActivity.java:531-532`** — The commented-out `showOptOutDialog()` was changed from `AlertDialog.Builder` to `MaterialAlertDialogBuilder` with a FQCN. This method has been disabled since issue #1240. Either leave it alone or delete it entirely.
- **`TravelBehaviorManager.java:34`** — `@Nullable` import added, but it's only used in commented-out code (line 299). The existing `android.app.AlertDialog` import on line 18 is similarly dead. Don't add new imports for dead code.

### 4. No-op style attributes in `styles.xml`

**`styles.xml:21-22`** — The `NavDrawerListItem` style has:

```xml
<!--        <item name="android:shape">oval</item>-->
<item name="android:radius">100dp</item>
```

`android:radius` is a `ShapeDrawable` XML attribute, not a View style attribute. Applied to a ListView style, it's silently ignored at runtime. The commented-out `android:shape` was also non-functional. Remove both lines.

## Verdict

**Request changes.** The critical `setColorFilter` bug will produce an incorrect tint color at runtime and needs to be fixed. The important items are cleanup from the prior round. Once those are addressed, this should be in good shape.

PR #1513, opened 2026-03-13
Decouple map abstraction from Google Maps and add MapLibre flavor

### LINKED ISSUE #1446: Switch to MapLibre
Hello,
Have you guys looked at using MapLibre and a tile server such as [openfreemap.org](https://openfreemap.org/)? This would save costs for my research.
Thanks

amanjn38:
Fixes #1446 

### DIFF
--- a/onebusaway-android/src/maplibre/java/org/onebusaway/android/map/maplibre/MapLibreMapFragment.java
+++ b/onebusaway-android/src/maplibre/java/org/onebusaway/android/map/maplibre/MapLibreMapFragment.java
@@ -76,6 +76,7 @@ import android.graphics.drawable.Drawable;
 import android.location.Location;
 import android.os.Bundle;
 import android.os.Handler;
+import android.os.Looper;
 import android.provider.Settings;
 import android.text.TextUtils;
 import android.util.Log;
@@ -302,7 +303,10 @@ public class MapLibreMapFragment extends SupportMapFragment
         if (savedInstanceState != null) {
             initMapState(savedInstanceState);
         } else {
-            Bundle args = getActivity().getIntent().getExtras();
+            Bundle args = null;
+            if (getActivity() != null) {
+                args = getActivity().getIntent().getExtras();
+            }
             if (args == null) {
                 args = new Bundle();
             }
@@ -909,7 +913,7 @@ public class MapLibreMapFragment extends SupportMapFragment
     // StopOverlay.OnFocusChangedListener
     // ============================================================================================
 
-    final Handler mStopChangedHandler = new Handler();
+    final Handler mStopChangedHandler = new Handler(Looper.getMainLooper());
 
     @Override
     public void onFocusChanged(final ObaStop stop, final HashMap<String, ObaRoute> routes,
@@ -933,10 +937,10 @@ public class MapLibreMapFragment extends SupportMapFragment
 
     @Override
     public void onLocationChanged(Location l) {
-        // MapLibre LocationComponent handles rendering automatically when activated;
-        // we still inform controllers about location updates.
         if (mControllers != null) {
-            // LocationHelper fires this; nothing more to do for the map display.
+            for (MapModeController controller : mControllers) {
+                controller.onLocation();
+            }
         }
     }
 
@@ -957,8 +961,8 @@ public class MapLibreMapFragment extends SupportMapFragment
     // ============================================================================================
 
     protected void showDialog(int id) {
-        MapDialogFragment.newInstance(id, this)
-                .show(getFragmentManager(), MapDialogFragment.TAG);
+        MapDialogFragment.newInstance(id)
+                .show(getChildFragmentManager(), MapDialogFragment.TAG);
     }
 
     public void showLocationPermissionDialog() {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Hey Aman, thanks for tackling the MapLibre integration — this is a significant architectural change and the overall direction is excellent. The `ObaMapFragment` interface with `asFragment()` and the `BuildConfig`-driven factory is a clean abstraction, and the manifest overlay to suppress Google Maps metadata is a nice touch. The stop overlay with directional icons works well. Before we can merge this, I need you to address a few items:

## Critical

### 1. Static fragment reference in `MapDialogFragment` causes memory leak and stale-reference crash

`MapLibreMapFragment.java:1004`

`MapDialogFragment` stores the parent fragment in a `private static MapLibreMapFragment mMapFragment` field. After a configuration change (screen rotation), the framework recreates the `DialogFragment`, but the static field still points to the old, destroyed fragment. This causes a memory leak and potential crashes when dialog button handlers call methods like `zoomToRegion()` or iterate `mControllers` on the stale reference.

The Google `BaseMapFragment` has the same pattern (it's pre-existing tech debt), but since this is new code, let's do it right. Use `getParentFragment()` inside the dialog callbacks instead of a static field:

```java
static MapDialogFragment newInstance(int dialogType) {
    MapDialogFragment f = new MapDialogFragment();
    Bundle args = new Bundle();
    args.putInt(DIALOG_TYPE_KEY, dialogType);
    f.setArguments(args);
    f.setCancelable(false);
    return f;
}

// In button handlers, retrieve the parent dynamically:
MapLibreMapFragment mapFragment = (MapLibreMapFragment) getParentFragment();
if (mapFragment != null && mapFragment.isAdded()) {
    mapFragment.zoomToRegion();
}
```

And show the dialog with `getChildFragmentManager()` so that `getParentFragment()` works:

```java
protected void showDialog(int id) {
    MapDialogFragment.newInstance(id)
            .show(getChildFragmentManager(), MapDialogFragment.TAG);
}
```

This also fixes item 4 below (`getFragmentManager()` deprecation).

### 2. NPE in `StopOverlay` — listener and `mMarkerData` not null-checked

`StopOverlay.java:480, 484, 492`

Three null-pointer crash paths reachable through normal user interaction:

- **Line 480** (`doFocusChange`): `mOnFocusChangedListener.onFocusChanged(...)` called without null-check. Crashes if a stop marker is tapped before the listener is set.
- **Line 484** (`removeFocus`): `mMarkerData.getFocus()` called without null-check. Crashes if the user taps the map before any stops have loaded (since `removeMarkerClicked` is wired up in `initMap`).
- **Line 492** (`removeFocus`): Same listener null-check issue.

Fix: Add null guards to match the defensive pattern used elsewhere:

```java
private void doFocusChange(ObaStop stop) {
    mMarkerData.setFocus(stop);
    if (mOnFocusChangedListener != null) {
        HashMap<String, ObaRoute> routes = mMarkerData.getCachedRoutes();
        mOnFocusChangedListener.onFocusChanged(stop, routes, stop.getLocation());
    }
}

private void removeFocus(LatLng latLng) {
    if (mMarkerData != null && mMarkerData.getFocus() != null) {
        mMarkerData.removeFocus();
    }
    Location location = (latLng != null) ? MapHelpMapLibre.makeLocation(latLng) : null;
    if (mOnFocusChangedListener != null) {
        mOnFocusChangedListener.onFocusChanged(null, null, location);
    }
}
```

### 3. `onLocationChanged()` is a no-op — controllers never notified of location updates

`MapLibreMapFragment.java:935-941`

The `onLocationChanged` method has an empty `if`-block that checks `mControllers` but does nothing with it. In the Google flavor, location updates trigger `controller.onLocation()`, which is how `StopMapController` knows to load stops near the user's current position. Without this, the MapLibre flavor will silently fail to refresh stops as the user moves.

```java
@Override
public void onLocationChanged(Location l) {
    if (mControllers != null) {
        for (MapModeController controller : mControllers) {
            controller.onLocation();
        }
    }
}
```

## Important

### 4. `getFragmentManager()` is deprecated

`MapLibreMapFragment.java:961`

`Fragment.getFragmentManager()` is deprecated and can return null. This is resolved by the fix in item 1 (switching to `getChildFragmentManager()`).

### 5. `new Handler()` without explicit Looper is deprecated

`MapLibreMapFragment.java:912`

The no-arg `Handler()` constructor is deprecated since API 30. Replace with:

```java
final Handler mStopChangedHandler = new Handler(Looper.getMainLooper());
```

### 6. Dead API-level checks

`StopOverlay.java:130, 136`

The `Build.VERSION_CODES.JELLY_BEAN_MR1` (API 17) checks are always true since `minSdkVersion` is 21. Remove the `if` guards and keep the timing code inline.

### 7. `getActivity()` can be null in async style callback

`MapLibreMapFragment.java:305`

`initMap` is called from the `onMapReady` -> style loaded callback, which is asynchronous. If the fragment detaches before the style loads, `getActivity()` returns null and `getActivity().getIntent().getExtras()` will NPE. Use a null guard or `requireActivity()`.

## Fit and Finish

### 8. Answering your questions from the PR description

- **BuildConfig field approach**: The reflection-based factory works. A slightly more robust alternative would be placing a concrete factory class in each flavor's source set (avoids R8/ProGuard keep-rule needs and catches typos at compile time), but the current approach is fine for now and can be revisited later.
- **Quantum colors in `colors.xml`**: Defining them locally is the right call — they're referenced by layouts that both flavors share. Good solution.
- **MapLibre SDK 11.5.2**: That's a reasonable version. No objection here.

PR #1521, opened 2026-03-14
Add save, share, and reuse trip plans feature (#1110)

### LINKED ISSUE #1110: Save, Reuse, and Share Planned Trips
Feedback from a user:

> After planning a trip you should be allowed to save it, name it, put it in favorites and share it.

diveshpatil9104:
## Summary

Implements **Save, Share & Reuse Trip Plans** (closes #1110).

Users can now:
- **Save** a planned trip with a custom name and optional favorite flag
- **Share** trip details (route, times, duration) via any app (Messages, Email, etc.)
- **View saved trips** from the navigation drawer
- **Re-open** a saved trip to view its itinerary
- **Favorite** trips with a star toggle
- **Delete** trips via long-press with confirmation

### What Changed

- **New menu actions** in Trip Plan screen: "Save trip" and "Share trip"
- **Save dialog** with trip name input and favorite checkbox
- **Share intent** formats trip details (depart/arrive times respecting 24h setting, duration, route)
- **Saved Trips screen** accessible from nav drawer, with RecyclerView list and empty state
- **Room database migration** (v2 → v3) adding `saved_trips` table
- **New files**: `SavedTripEntity`, `SavedTripDao`, `SavedTripsManager`, `ItineraryJsonConverter`, `SavedTripsActivity`, `SavedTripsFragment`, 4 layout XMLs

### Architecture

Follows existing codebase patterns:
- Room Entity/DAO/Manager pattern (mirrors `RecentStopsManager`)
- Gson for `Itinerary` serialization/deserialization
- Nav drawer integration matching existing items
- `include_toolbar` + `0dp/weight` layout pattern
- White-label safe (`%1$s` + `getString(R.string.app_name)`)

## Screen Recording

https://github.com/user-attachments/assets/ae7fa355-25d3-4dea-858c-92bb4dee9bb2

## Test Plan

-  Save a trip with custom name — appears in Saved Trips list
-  Save a trip as favorite — star icon filled
-  Share a trip — share sheet shows formatted trip text
-  Open Saved Trips from nav drawer — list displays correctly
-  Tap saved trip — re-opens in Trip Plan screen
-  Toggle favorite — star updates immediately
-  Long-press delete — confirmation dialog, trip removed
-  No trip planned → "Save trip" shows "No trip to save" toast
-  Empty state — "No saved trips" message when list is empty

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -263,6 +263,7 @@ dependencies {
     googleImplementation 'com.google.android.libraries.places:places-compat:2.6.0'
     // Autocomplete text views with clear button for trip planning
     implementation 'com.google.android.material:material:1.12.0'
+    implementation 'androidx.recyclerview:recyclerview:1.3.1'
     // Explicit dependency on Gson is now apparently needed
     implementation 'com.google.code.gson:gson:2.10.1'
     // Unit tests - seems like this is still necessary w/ Android X even though useLibrary is declared earlier
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/SavedTripsFragment.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/SavedTripsFragment.java
@@ -237,7 +237,7 @@ public class SavedTripsFragment extends Fragment {
                 favoriteIcon.setOnClickListener(new View.OnClickListener() {
                     @Override
                     public void onClick(View v) {
-                        int pos = getAdapterPosition();
+                        int pos = getBindingAdapterPosition();
                         if (pos != RecyclerView.NO_POSITION) {
                             onFavoriteClicked(trip, pos);
                         }


aaronbrethorst (changes requested):
Hey Divesh, thanks for addressing the feedback from the first review -- the from/to address passthrough, the stale position fix, and the `withToggledFavorite()` helper all look correct. A few things remain before we can merge.

## Critical Issues (0 found)

None -- the from/to address passthrough is working correctly now.

## Important Issues (2 found)

### 1. `getAdapterPosition()` is still deprecated -- and `getBindingAdapterPosition()` IS available

`SavedTripsFragment.java:240`

In the previous review you mentioned that `getBindingAdapterPosition()` isn't available because the project uses a pre-1.2.0 RecyclerView. However, Gradle dependency resolution shows RecyclerView **1.3.1** (pulled in transitively by `material:1.12.0`), so `getBindingAdapterPosition()` is available. Please replace:

```java
int pos = getAdapterPosition();
```

with:

```java
int pos = getBindingAdapterPosition();
```

### 2. Fire-and-forget coroutines in `SavedTripsManager` silently swallow exceptions

`SavedTripsManager.kt:41, 97, 110`

`saveTrip()`, `toggleFavorite()`, and `deleteTrip()` launch coroutines with `CoroutineScope(Dispatchers.IO).launch { ... }` and have no exception handler. If the DAO call throws (e.g., a SQLite constraint violation), the exception is silently swallowed -- the user sees "Trip saved" toast but the trip isn't actually saved.

At minimum, wrap the DAO calls in a try/catch and log the error. Better yet, consider using a shared `CoroutineExceptionHandler`:

```kotlin
private val exceptionHandler = CoroutineExceptionHandler { _, throwable ->
    Log.e(TAG, "Database operation failed", throwable)
}

// Then in each launch:
CoroutineScope(Dispatchers.IO + exceptionHandler).launch {
    dao.insert(trip)
    callback?.invoke(id)
}
```

This also applies to `toggleFavorite` and `deleteTrip`, where a failed write means the UI is now out of sync with the database.

## Suggestions (1 found)

### 1. `saveTripToDatabase` still uses positional constructor for `SavedTripEntity`

`TripPlanActivity.java:547-550`

The same fragility issue flagged for `onFavoriteClicked` (now fixed via `withToggledFavorite()`) still exists here -- `SavedTripEntity` is constructed with 11 positional args. If the entity's constructor order changes, this call site will silently pass wrong values. Consider using Kotlin named arguments (from a Kotlin call site) or creating a builder/factory method on `SavedTripEntity`.

This isn't blocking, but it's worth tracking as a follow-up.

## Verification of Previous Feedback

| Previous Item | Status |
|---|---|
| Critical #1: from/to address passthrough | Fixed -- `onTripClicked` now builds `CustomAddress` objects and uses `TripRequestBuilder` |
| Important #1: deprecated `getAdapterPosition()` | **Not fixed** -- author believed RecyclerView was pre-1.2.0, but resolved version is 1.3.1 |
| Important #2: stale position in `confirmDelete` | Fixed -- now uses `mTrips.indexOf(trip)` with bounds check |
| Important #3: fragile positional constructor in `onFavoriteClicked` | Fixed -- uses `withToggledFavorite()` helper |
| Suggestion #1: duplicate nav drawer icon | Fixed -- changed to `ic_drawer_star` |
| Suggestion #2: share text omits date | Fixed -- format is now `MMM d, HH:mm` / `MMM d, hh:mm a` |

## Strengths

- All critical and most important items from the first review were addressed correctly
- The `TripRequestBuilder` + `Source.NOTIFICATION` approach for re-opening trips is clean
- `withToggledFavorite()` is a nice, idiomatic Kotlin helper
- The `indexOf(trip)` fix for stale position is correct since `SavedTripEntity` is a data class with proper `equals()`
- Nav drawer icon change differentiates Saved Trips from Plan a Trip

## Recommended Action

1. Replace `getAdapterPosition()` with `getBindingAdapterPosition()` (Important #1)
2. Add exception handling to fire-and-forget coroutines (Important #2)
3. Consider the positional constructor suggestion as a follow-up

diveshpatil9104:
Thanks Aaron! issues addressed:

  **Important 1** — I ran `./gradlew :onebusaway-android:dependencies --configuration obaGoogleDebugCompileClasspath | grep recyclerview` and the resolved version is 1.1.0 (not 1.3.1). The 1.3.1 version from **material:1.12.0** is not being pulled in transitively  Gradle resolves to 1.1.0. `getBindingAdapterPosition()` was added in 1.2.0, so it does not compile. Keeping getAdapterPosition() for now. Happy to bump the RecyclerView dependency in a follow-up PR if you'd like.

  Important 2 — Fixed. Added a shared CoroutineExceptionHandler with `Log.e() ` to all three fire-and-forget coroutines (`saveTrip`, `toggleFavorite`, `deleteTrip`). Database failures are now logged instead of silently swallowed.

  Suggestion 1 — Noted for follow-up. Will track the positional constructor cleanup separately.

diveshpatil9104:
My bad on the RecyclerView version 🥲 you were right! Added explicit recyclerview:1.3.1 to build.gradle and switched to getBindingAdapterPosition(). All done.

PR #1525, opened 2026-03-15
Don't crash fetching experiments before setting a region

hufman:

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/io/request/survey/ObaStudyRequest.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/io/request/survey/ObaStudyRequest.java
@@ -7,6 +7,7 @@ import androidx.annotation.NonNull;
 
 import org.onebusaway.android.R;
 import org.onebusaway.android.app.Application;
+import org.onebusaway.android.io.elements.ObaRegion;
 import org.onebusaway.android.io.request.RequestBase;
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Hey Walter, nice catch on this null-region crash -- clean and minimal fix.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (1 found)

### 1. Pre-existing: `URI` is a static field that can retain stale values

`ObaStudyRequest.java:29`

`URI` is declared `private static`, which means it's shared across all `Builder` instances. If a `Builder` runs successfully for Region A, then a second `Builder` runs while `region == null` (hitting the early return), `URI` still holds Region A's URL. The subsequent `build()` call would create a request using the stale URI.

This isn't introduced by your PR -- it's pre-existing -- but it's worth noting since you're in this code. Filed #1539 to track making `URI` an instance field, which would also make the early-return behavior more predictable (the request would get a null URI and fail gracefully in `StudyRequestTask.doInBackground`'s catch block).

## Strengths

- Correctly identifies the root cause: `getCurrentRegion()` can return null before region discovery completes
- Minimal, focused change -- doesn't over-engineer the fix
- Follows the existing pattern (mirrors the `baseUrl == null` guard on the next line)
- The crash path is fully eliminated: even if `URI` ends up null, `StudyRequestTask.doInBackground` catches exceptions, but preventing the NPE at the source is the right call

## Recommended Action

Ship it.

PR #1544, opened 2026-03-20
Fix #1537 (Item 4): Add accessibility support for arrival badges

### LINKED ISSUE #1537: Starred stops arrival badges: follow-up improvements
## Context

Follow-up items from PR #1520 (inline arrival badges on starred stops). These were identified during review but are not blocking the initial implementation.

### 1. Layout duplication — use `<include>` instead of copying `stop_list_item.xml`
`starred_stop_list_item.xml` duplicates `stop_list_item.xml` and adds the arrivals container. Use `<include layout="@layout/stop_list_item" />` to keep layouts in sync. (From PR review round 1.)

### 2. Error vs. empty state indicator
When API calls fail, arrival badges silently disappear — indistinguishable from "no upcoming arrivals." Show a subtle error indicator (e.g., small "!" icon or "Could not load" text) so users know to check their connection rather than assuming no buses are coming. (From PR review round 1.)

### 3. Serial network requests in `StarredStopsArrivalsLoader`
`loadInBackground()` makes one HTTP request per starred stop sequentially. With many starred stops on a slow network, total load time scales linearly. Consider parallelizing with an `ExecutorService` or delivering partial results.

### 4. Accessibility for arrival badges
Badge `TextView`s inside the `HorizontalScrollView` may not be optimally navigable for screen reader users. Consider adding `android:importantForAccessibility` attributes and ensuring `contentDescription` is set on badges.

## Related
Closes: n/a
PR: #1520

Gitkbc:
I've implemented the accessibility issues for arrival badges (Item 4 of #1537).

Instead of TalkBack just reading out the tiny badge text, I updated the StarredStopsAdapter to generate a full sentence description—something like "Route 10, arriving in 5 minutes." I also silenced the HorizontalScrollView container so the screen reader doesn't get stuck on it, making it much smoother to swipe through the actual arrival times.

Tested with TalkBack on a physical device — before the fix it read  "list", after the fix it correctly announces each badge.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/StarredStopsAdapter.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/StarredStopsAdapter.java
@@ -142,6 +142,16 @@ class StarredStopsAdapter extends SimpleCursorAdapter {
             }
             badge.setText(context.getString(R.string.starred_stop_arrival_badge,
                     routeName, etaText));
+            String contentDesc;
+            if (eta <= 0) {
+                contentDesc = context.getString(
+                        R.string.arrival_badge_content_description_now, routeName);
+            } else {
+                contentDesc = context.getString(
+                        R.string.arrival_badge_content_description_min, routeName, (int) eta);
+            }
+            badge.setContentDescription(contentDesc);
+            badge.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_YES);
 
             badge.setTextSize(BADGE_TEXT_SIZE_SP);
             badge.setTextColor(ContextCompat.getColor(context, android.R.color.white));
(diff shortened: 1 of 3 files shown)


Gitkbc:
@aaronbrethorst Sir,

While testing this, I noticed that when a stop has multiple arrivals for the same route, TalkBack reads "Route 128, arriving in 8 minutes" then "Route 128, arriving in 33 minutes" — the repeated route name feels a bit redundant both visually and for screen reader users.

Would it be worth showing the route name once and labeling subsequent badges as "Bus 1", "Bus 2"? Also the TalkBack will be "Arriving: Route 128, Bus 1 in 6 minutes, followed by Bus 2 in 33 minutes." I've attached a quick hand-sketch of what I mean. If this sounds like a good enhancement, I’m happy to present a digital preview so we can see exactly how it looks before coding it. Happy to explore this as a follow-up if it sounds useful.

image

Issue #12, opened 2012-08-31
Add location provider status to bug report email


paulcwatts:
This will be helpful in diagnosing when people email me about GPS problems. I may be able to add a lat/lon/accuracy as well.

Issue #11, opened 2012-08-30
Search route hint has wrong background


paulcwatts:
This is the hint text on the "Search" tab of the "My Routes" activity.

I never noticed when the default background was white, but on ICS/JB it's off white it has this strange white-on-off white.

Issue #10, opened 2012-08-29
Use transit-specific icon in the notification tray


paulcwatts:
I use the "reminder/notification" icon which looks like a generic "notification" icon. We should use something that more specifically calls out as OneBusAway, like the old bus icon.

PR #1554, opened 2026-04-14
Animated vehicle markers with speed-based extrapolation

bmander:
## Summary

Adds real-time vehicle marker animation on the map between AVL data updates, using a probabilistic speed model to extrapolate position. Vehicle markers now move smoothly along the route polyline at their estimated speed, rather than teleporting when new data arrives every ~15s. Also adds a new trip details map view that visualizes the full probability distribution over the vehicle's predicted location.

## How to review

This PR is organized into **8 logical commits**. Each commit builds on the previous and is independently meaningful. To review this efficiently, step through the commits in order via the "Commits" tab:

1. **Expose trip schedule distances and add geometry utilities** — Maps new fields from the OBA API (`distanceAlongTrip`, `lastKnownDistanceAlongTrip`) and adds helpers (haversine distance matching the OBA server formula, segment lookups, elapsed-time formatting).
2. **Modernize AnimationUtil and extract StampedPolylineFactory** — Removes pre-ICS animation code paths (minSdk is 21) and extracts polyline-arrow stamp creation from `BaseMapFragment` into a reusable factory.
3. **Add probability distribution library and Polyline utility** — Pure math: `ProbDistribution` interface with Gamma, Dirac, frozen, and mixture implementations; `Polyline` class for O(log n) distance-based interpolation. JVM-testable.
4. **Add extrapolation engine with gamma and schedule-replay strategies** — Speed model (H34 two-gamma mixture) and the per-trip `Extrapolator` that picks the appropriate strategy (real-time or schedule replay) based on available data.
5. **Add TripDataManager, coroutine-based pollers, and ThrottledFrameLoop** — Data layer: per-trip cache, `RoutePoller`/`TripPoller` replacing the old Loader/Handler mechanism, and a Choreographer wrapper for throttled frame animation.
6. **Extract VehicleMapController and animate markers between data updates** — Refactors the monolithic `VehicleOverlay` into `VehicleMapController`, `VehicleMarkerState`, `VehicleIconFactory`, and `VehicleInfoWindowAdapter`. Wires per-frame extrapolation into the route map.
7. **Add trip map view with route overlay and distance estimate visualization** — New `TripMapFragment` showing a single trip's route, animated vehicle, and a colored polyline segment visualizing the probability distribution over vehicle location.
8. **Wire up trip details UI, data views, and speed estimation UI** — `TripDetailsActivity` map/list toggle, developer data views (trajectory graph + table), and supporting layouts/strings/styles.

## Test plan

- [x] `./gradlew test` — JVM unit tests pass (distributions, haversine, trip gating, schedule replay, gamma model)
- [x] `./gradlew assembleObaGoogleDebug` — compiles clean
- [x] `./gradlew installObaGoogleDebug` — installs and launches on device
- [ ] Manual: observe smooth vehicle animation on the route map
- [ ] Manual: open a trip's detail map and verify the route polyline, vehicle marker, and distance estimate visualization
- [ ] Manual: toggle between map/list in trip details
- [ ] Manual: verify MapLibre flavor still builds (`./gradlew assembleObaMaplibreDebug`)

## Notes

- This is intentionally **not** a single squashed commit (per the contributing guidelines) because the scope is large and the commit-per-layer structure is how it was designed to be reviewed. Happy to squash on merge if that's preferred.

Implements #1555

## Summary by CodeRabbit

* **New Features**
  * Trip map: single-trip map with vehicle overlay, stamped route polylines, stop markers, and per-trip estimate overlays
  * Trajectory graph and vehicle-location data activity for history, table view, and debug
  * Selectable vehicle markers with animated transitions and info windows

* **Improvements**
  * Better marker icons, directional rendering, and a fast-estimate marker/icon set
  * Map styling (light/dark) and menu/UX tweaks in trip details (map/list toggle)


aaronbrethorst (changes requested):
Brandon — this is a remarkable piece of work. A 9k-line PR could easily be a mess, and this one is the opposite: the math layer is tested to a standard you usually see in dedicated libraries (the H34 mixture worked-example, the PDF-integrates-to-one assertions, `BisectTest`'s timeout guards on every degenerate input), the 888-line `VehicleOverlay` decomposes into well-bounded units, per-frame allocation discipline is real (`subPolylineMapInto`, the scratch `Location`, reused arrays in `DistanceEstimateOverlay`), and the commit-per-layer organization made the review tractable. The choreographer-driven main-thread invariant is a great design call — it lets `Trip` hold plain mutable fields and lets you skip a synchronization story you'd otherwise need to write and test.

Two themes pulled me back from a clean merge:

1. **Global singleton + multi-screen state.** `TripDataManager.INSTANCE` is shared across the route map, the trip details screen, and the data view. A few code paths assume they own it (e.g. `RouteMapController.clearCurrentState()` wipes the whole registry; in-flight fetches don't get cancelled). Under realistic use, screens will step on each other.
2. **Lifecycle gaps.** The choreographer loop and the pollers can outlive their host (fragment hidden via `FragmentManager`, async map-ready arriving post-pause, fetch results resurrecting cleared trips).

Those are the real ones. The rest are tractable cleanups.

I'm going to ask for changes before this merges — the global-state corruption in particular has a clear path to user-visible weirdness (stale markers, animations referencing dead Trip objects). Once those land, this is shipping-quality.

---

## Critical Issues (3 found)

### 1. `RouteMapController.clearCurrentState()` wipes the global trip registry
- `onebusaway-android/src/main/java/org/onebusaway/android/map/RouteMapController.java:150` — `TripDataManager.INSTANCE.clearAll()` runs every time the user switches routes on the home map. But `TripDataManager` is a singleton shared with `TripDetailsActivity`, `TripMapFragment`, and `VehicleLocationDataActivity`. If any of those screens is in the back stack with a `Trip` reference, that `Trip` is now orphaned — the next `getOrCreateTrip(tripId)` call creates a *new* one, but `VehicleMarkerState.trip` is not refreshed in `VehicleMapController.updateVehicle()` (`VehicleMapController.java:149-160`), so the per-frame extrapolation loop reads from the now-stale instance.
- Fix: stop calling `clearAll()` here. The `MAX_TRACKED_TRIPS = 100` LRU eviction will handle this naturally, or do a targeted eviction of trips matching the previous route. Also refresh `vehicle.trip` to the current `getOrCreateTrip(tripId)` in `VehicleMapController.updateVehicle`.

### 2. `clearAll()` doesn't cancel in-flight fetches; ghost trips can resurrect
- `onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/TripDataManager.kt:370-375` together with `ensureFetched` (305-333). `clearAll()` empties the registry and the pending sets, but tasks already running on `fetchExecutor` continue. When they complete, `mainHandler.post { onSuccess(...) }` calls `getOrCreateTrip(tripId).schedule = ...` or `... .polyline = ...`, re-creating the trip you just cleared with only a schedule/shape and no status/history. Combined with #1, this is a recipe for partially-populated phantom trips persisting across user actions.
- Fix: track an epoch counter and have the `onSuccess` block check it, or maintain a cancellation set that pending tasks consult before applying results.

### 3. `Trip.kt:39` documents a synchronization scheme that doesn't exist
- The KDoc on `Trip` reads `Thread safety: all mutable state is accessed under TripDataManager's @Synchronized lock.` There is no `@Synchronized` anywhere in `TripDataManager` — `grep` finds only this Trip.kt comment. The real invariant ("main thread only") is documented on `TripDataManager.kt:40-45`, and the two file-level docs are now contradicting each other.
- This isn't just a stale comment — it's actively misleading. A maintainer who reads `Trip.kt` first will reasonably conclude they can call `Trip` methods from a background thread (since "the lock handles it"), which would race against everything else. Fix the docstring to reference the main-thread contract.

---

## Important Issues (9 found)

### 4. `TripMapFragment` can start its poller and frame loop while paused
- `TripMapFragment.kt:172-182` — `onOverlaysReady` unconditionally starts both the extrapolation controller and a new `TripDetailsPoller`. It's invoked via `TripMapOverlayFactory.create`, which can fire either inline (cached shape) or via `TripDataManager.ensureShape`'s main-thread callback after a network fetch. Either path can resolve after `onPause` ran — the user backgrounds the app while the map is still loading, or rotates mid-fetch.
- Net effect: poller and choreographer run while the fragment is paused. Add an `isResumed()` guard, or set `activated = true` here and let `onResume()` start the moving parts.

### 5. `onMapReady` doesn't guard against post-detach invocation
- `TripMapFragment.kt:103-114`. `getMapAsync(this)` is async; `onMapReady` can fire after `onDestroyView`. The first `requireContext()` (line 108) or the `requireArguments()` inside `tripId` (line 72) will throw `IllegalStateException` in that window. Same exposure in `activate()` at line 145.
- Early-return if `!isAdded || view == null`.

### 6. `VehicleOverlay` frame loop keeps running when the host fragment is hidden
- `VehicleOverlay.java:149-155` and `ThrottledFrameLoop.kt`. The loop only stops on `clear()` or when `mActivity.isDestroyed()`. `HomeActivity` hides the map fragment via `FragmentManager.hide()` when the user navigates to starred stops/etc; the activity isn't destroyed and the choreographer keeps firing `updatePositions()` on markers that aren't visible.
- Plumb a stop into `BaseMapFragment.onPause`/`onHidden`.

### 7. `Trip.getOrCreateExtrapolator()` silently locks in the wrong strategy
- `Trip.kt:142-152`. Strategy selection runs once and is cached for the trip's lifetime. If `routeType` is null when `extrapolate()` is first called — which happens whenever `recordStatus` runs before `recordTripsForRouteResponse` populates the route metadata — the trip is permanently bound to `GammaExtrapolator` even if a later poll learns it's actually a rail line. Subway/rail extrapolation silently degrades to the bus model for trips with the wrong race ordering.
- Either invalidate the cached extrapolator when `routeType` changes, defer construction until `routeType` is known, or document the invariant on `getOrCreateExtrapolator` and `Trip.kt:36`. (Today the class-level KDoc on `Trip.kt:36-37` also claims a third "schedule-only fallback" strategy that doesn't exist — `getOrCreateExtrapolator` only ever returns Gamma or ScheduleReplay.)

### 8. `VehicleMapController.updatePositions` catches `RuntimeException` — too broad
- `VehicleMapController.java:317-324`. The comment says this is for `require()` failures in the gamma model, but `RuntimeException` is wide enough to swallow NPE from a race with `clear()`, `ConcurrentModificationException` if `mStates` is mutated mid-iteration, and `IllegalStateException` from the Maps API on a removed marker. At 20fps, a structural bug would surface as `W` log spam and degraded animation — quietly enough that nobody would notice in production.
- Narrow this to `IllegalArgumentException` (which is what `require()` throws) and let the rest propagate. Same pattern in `TripExtrapolationController.kt:43-52`.

### 9. `Pollers.kt` swallows everything and retries forever at full cadence
- `Pollers.kt:65-67, 115-117, 145-147`. `catch (e: Exception)` is broad enough to swallow Jackson parse errors, `IllegalStateException` from `Application.get()`, NPEs — anything that isn't an `Error`. Combined with the unconditional 10s `delay()`, a persistent failure (bus parked in a tunnel; misconfigured server; broken response shape) means six requests per minute, for the entire lifetime of the screen, with the user staring at a spinner that never resolves.
- Narrow the catch to `IOException` plus a separate clause for parse/state errors, add exponential backoff with a cap, and surface failures back to the UI. `TripDetailsListFragment.java:218,281-282,514-523` has no escape hatch today — if every poll fails, the loading spinner shows indefinitely. `fetchTripDetailsOnce` (lines 129-149) explicitly documents itself as "does not notify callers on success or failure"; pull-to-refresh failures vanish silently.

### 10. `AffineTransformDistribution` accepts negative or zero `scale` silently
- `AffineTransformDistribution.kt:22-38`. For `scale < 0`, `pdf` returns 0 (invalid: should be a flipped distribution), `cdf` becomes a step at `offset`, and `quantile` is non-monotone in `p` — violating the `ProbDistribution` contract. For `scale == 0`, you get a degenerate point mass that doesn't integrate to 1. Today the only caller (`GammaExtrapolator.kt:57`) gets a non-negative scale because `Trip.kt:131` guards `dtMs < 0`, but the type itself doesn't enforce its contract.
- Add `require(scale > 0)` (or accept `0` and explicitly route to a `DiracDistribution(offset)`). The other distributions in this package guard their preconditions and are exemplary; bring this one up to that bar.

### 11. `Trip` exposes all state via public `var`; parallel-array invariants live in convention
- `Trip.kt:41-72`. `history`, `fetchTimes`, `localFetchTimes` are `val mutableListOf<…>()` — the reference is final but the contents are publicly mutable. `schedule`, `serviceDate`, `polyline`, `routeType`, `lastActiveTripId`, `tripDetailsResponse` are all public `var`. The three lists must stay index-aligned (truncated together, grown together) and only `recordStatus` is a well-formed producer — but that contract is implicit. Any caller can break it.
- At minimum, make the three lists `private val` and expose `List<>` views (or use `HistorySnapshot` exclusively). Make the cache fields settable only via package-private setters or via the manager.

### 12. `HistorySnapshot` is undermined by per-field accessors
- `TripDataManager.kt:159-187`. The whole point of `HistorySnapshot` is that two separate `getHistory()` / `getFetchTimes()` calls can see different sizes if a poll lands between them. But the per-field accessors (`getHistory`, `getFetchTimes`, `getLocalFetchTimes`, `getHistorySize`) are still public and several are unused by production code (only used in tests).
- Either delete the per-field accessors and force callers to use `HistorySnapshot`, or document the snapshot type as transitional in a way that doesn't outlive its label.

---

## Suggestions (8 found)

### 13. `onTripMapActivationFailed` collapses six failure modes into one generic toast
- `TripDetailsActivity.java:240-246` and `TripMapFragment.kt:145-170`. `TripMapOverlayFactory` enumerates five distinct reasons (`MISSING_SCHEDULE`, `MISSING_REFERENCES`, `TRIP_NOT_IN_REFERENCES`, `MISSING_SHAPE_ID`, `SHAPE_FETCH_FAILED`) plus the "no cached details" case in `activate()`. All six get the same toast: "Trip map data is not available yet." For `SHAPE_FETCH_FAILED` the right answer is probably a retry, not falling back to list view. Plumb the reason through `onTripMapActivationFailed(reason)` so distinct problems get distinct UX.

### 14. Stale test class name `SpeedEstimatorTest`
- `onebusaway-android/src/androidTest/java/org/onebusaway/android/extrapolation/test/SpeedEstimatorTest.kt:35-37`. KDoc claims "Tests for the speed estimation framework classes" but no `SpeedEstimator` class exists; the tests actually exercise `TripDataManager` and the extrapolation result types. Rename to something like `TripDataManagerExtrapolationTest` and update the KDoc.

### 15. `Polyline.kt:23-25` complexity claim overstates `subPolyline`
- `interpolate` and `bearingAt` are O(log n); `subPolylineInto`/`subPolylineMapInto` are O(log n + k) where k is the number of vertices in the requested range. The KDoc says "O(log n) via binary search" for `subPolyline` — true for the lookups, false for the walk. Fix the comment or add a separate claim.

### 16. `TripDataManager.changes` emission contract is overstated
- `TripDataManager.kt:70`. The KDoc says "Emits Unit whenever any mutation method runs" but `notifyChanged()` is gated on actual state changes — duplicate `recordStatus`, no-op `putSchedule(null, …)`, zero-serviceDate `putServiceDate`, etc., all run without emitting. Adjust to "whenever a mutation method actually changes registry state."

### 17. `TripDataManager.ensureFetched` failure retry behavior needs at least a log bump
- `TripDataManager.kt:296-333`. The "no data" path logs at `D` (line 328). A persistently failing schedule fetch means a route's gamma extrapolator silently returns `MissingSchedule` forever, the user sees a stuck marker, and there's nothing in `adb logcat` above `D` to point at it. Bump to `W` and consider a per-trip failure marker so debug overlays can surface it. Same theme as #9.

### 18. `Polyline` leaks its internal point list
- `Polyline.kt:26`. Takes a `List<Location>` and exposes it as public `val points`. Callers passing a `MutableList<Location>` (or `Location` instances they mutate) can desync `cumulativeDistances` from `points` silently. Defensive-copy in the constructor (`points.toList()`) and require `size >= 2` via `require()`.

### 19. `MainScope().launch` in pollers is fire-and-forget
- `Pollers.kt:50, 100, 135`. Each poller creates its own scope; nothing ties them to a lifecycle. The "caller owns lifecycle" contract is fine when honored, but it's also exactly the kind of thing #4 above slipped through. Consider attaching to `viewLifecycleOwner.lifecycleScope` (or making the poller a `DefaultLifecycleObserver`) so this is structurally enforced.

### 20. `TripDataManager` LRU keepalive comment lists too few methods
- `TripDataManager.kt:56-59` calls out `recordStatus / getOrCreateTrip / isScheduleCached` but every read that goes through `getTrip()` promotes the entry (`getHistory`, `getAnchor`, `getSchedule`, `getPolyline`, etc., all promote). A future maintainer who calls `getAnchor` on a long-unobserved trip will be surprised to find they just kept it alive. Just say "every access via `getTrip`/`getOrCreateTrip`."

---

## Strengths

- Probability/math layer is the strongest part of this PR. `GammaDistribution`, `DiracDistribution`, and `FrozenDistribution` use constructor `require()` checks to make illegal states unrepresentable. The H34 worked-example in `GammaSpeedModelTest:73-78` pins down a real-world number; `BisectTest`'s `@Test(timeout = 1000)` on every degenerate input prevents future infinite-loop regressions; `extremely low speed does not crash` (line 201-210) locks in the mixture fallback.
- The `VehicleMapController.java:362-367` reference-equality trick on `anchor` to detect "fresh data" is the natural signal and avoids epsilon comparisons. Comment explains the *why*, which is the right level.
- Per-frame allocation discipline in `DistanceEstimateOverlay` (reused `DoubleArray`, `latLngBuffer`, `scratchLocation`) and `Polyline.subPolylineMapInto` is exactly right for a 20Hz loop.
- The commit organization made review tractable — eight commits, each one mappable to a layer. Don't squash if you don't have to.
- Defensive numerics in `bisect` (iteration caps, NaN/Inf guards, documented fallback semantics) are model.
- The sealed `ExtrapolationResult` carries enough debug fidelity that future log messages and crash dumps will tell you *why* extrapolation didn't run, even if every caller currently treats it as binary. That's an investment that pays back.

---

## Recommended Action

1. Fix the global-state corruption: stop the indiscriminate `clearAll()` in `RouteMapController` (#1), guard fetch results against post-clear resurrection (#2), and update `Trip.kt:39` to match reality (#3).
2. Tighten the fragment/poller lifecycles: guard `onOverlaysReady`/`onMapReady` against post-pause invocation (#4, #5), and stop the choreographer loop on hide/pause (#6).
3. Fix the extrapolator strategy lock-in (#7) and narrow the `RuntimeException` catches (#8).
4. Make the poller failure path observable: narrow exceptions, add backoff, surface to UI (#9). Either fix `fetchTripDetailsOnce`'s silent contract or document it loudly.
5. Tighten the types: `AffineTransformDistribution` invariants (#10), `Trip` field visibility (#11), `HistorySnapshot` exclusivity (#12).
6. Suggestions are independent — pick them up as you go.

Verdict: **request changes.**

The architecture is right. The math is solid. The remaining issues are the kind that surface in production rather than CI, and they're all addressable without restructuring anything you've built.

amrhossamdev:
That's amazing !, can you send me apk to test that out ? ( currently my machine have problem with android studio )

bmander:
Sure! Here's a debug APK built from this branch (commit 0bb532f0):

📦 **[onebusaway-animated-markers-debug.apk](https://github.com/bmander/onebusaway-android/releases/download/animated-markers-test-1/onebusaway-animated-markers-debug.apk)** (25 MB)

(Release page: https://github.com/bmander/onebusaway-android/releases/tag/animated-markers-test-1)

A couple of install notes:
- It's a debug-signed `obaGoogleDebug` build, so you may need to allow installs from unknown sources.
- If you have the Play Store version of OneBusAway installed, you'll need to uninstall it first — the signature mismatch will block the install.

To see the animated markers in action, open a map with active vehicles (or tap a vehicle → trip details) and watch the markers move between AVL updates. Would love to hear how it feels on a real device!

aaronbrethorst (approved):
This looks terrific. Amazing work, @bmander

PR #1563, opened 2026-05-16
Fix arrivals onBackPressed() on Android 13+

robertgodfrey:
## Problem

Pressing back while viewing a stop's arrivals exited the app instead of dismissing the panel and returning to the map.

## Root cause
  
HomeActivity overrode `onBackPressed()` to handle panel dismissal, but this override was silently bypassed on Android 13+ (API 33+). Modern Android routes back presses through `WindowOnBackInvokedDispatcher` → `AndroidX` `OnBackPressedDispatcher`, whose fallback calls `ComponentActivity.super.onBackPressed()` directly — skipping subclass overrides entirely. The app has used `androidx.activity:activity-ktx:1.9.3` for about 2 years, which made this the behavior on all currently-supported devices.

## Fix

Migrate the back press logic from `onBackPressed()` into an `OnBackPressedCallback` registered with `getOnBackPressedDispatcher()` in `onCreate()`. This is the API that both the AndroidX dispatcher and the OS-level `OnBackInvokedDispatcher` actually invoke, and it works correctly on all API levels (not just 13+).

## Behavior

  - Panel expanded/anchored → collapse to header
  - Panel collapsed → dismiss arrivals, return to map
  - No panel showing → exit as before

Before:

https://github.com/user-attachments/assets/4cb8f538-efec-42cf-ac91-551fd7a573f5

After:

https://github.com/user-attachments/assets/9b857614-7b02-49b7-a413-8730c2ea781f

## Summary by CodeRabbit

* **Chores**
  * Updated back button navigation handling to use modern Android framework patterns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/onebusaway-android/pull/1563?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.java
@@ -115,6 +115,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 
+import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
Rob, this is a really nicely scoped fix. The root-cause writeup in the PR description is genuinely illuminating — pinning the silent bypass on `WindowOnBackInvokedDispatcher` → `OnBackPressedDispatcher` → `ComponentActivity.super.onBackPressed()` skipping subclass overrides is exactly the kind of explanation that makes a one-file diff easy to land with confidence. The Before/After videos are also a great touch.

I cross-checked the migration against the official Android docs (Provide custom back navigation; Add support for the predictive back gesture) and the implementation matches the documented pattern step-for-step.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (1 found)

- The fallback inside `handleOnBackPressed()` uses the documented `setEnabled(false); getOnBackPressedDispatcher().onBackPressed();` pattern, which permanently disables the callback after first fallback [HomeActivity.java:428-429]. In practice this is fine for `HomeActivity` because the only time the fallback runs is when no panel is showing and the activity is about to finish, so the disabled state never matters. Worth being aware of as a subtle behavior of this idiom, but no change needed here.

## Follow-up (out of scope, tracked separately)

The same Android 13+ bug almost certainly affects two other activities still using the deprecated `onBackPressed()` override:

- `TripPlanActivity.java:261-267` — has panel-collapse logic structurally identical to the bug fixed here.
- `InfrastructureIssueActivity.java:722-726` — its post-back spinner reset is being silently skipped on API 33+.

I've opened **[#1566](https://github.com/OneBusAway/onebusaway-android/issues/1566)** to track these migrations so they don't get lost.

thanks for knocking this out!

Issue #1582, opened 2026-06-25
Deep-link route latch can drop rapid back-to-back distinct intents


aaronbrethorst:
## Problem

The staged deep-link route is a `StateFlow<String?>` latch (`HomeViewModel.deepLinkRoute`), set before `setContent` and cleared by `DeepLinkEffect` after navigating. Rapid, *distinct* incoming intents (e.g. FCM A → B → A, or pinned-shortcut bursts) can overwrite the latch before the collector consumes the previous value, dropping a navigation.

The same-route case is already guarded; only rapid distinct routes are at risk. Probability is low given how reminders/shortcuts fire in practice.

## Proposed fix

Use a `Channel<String>` (queued) instead of a `StateFlow<String?>` (latched) for deep-link routes so each intent is delivered exactly once, or add a serial guard so a staged route can't be overwritten before consumption.

## Context

Follow-up from the high-level architecture review of the single-Activity Compose modernization (`pr-stack/b1-compose-host`).

PR #1584, opened 2026-06-27
Replace HomeActivity downcasts with Hilt EntryPoints; remove the Navigator latch

### LINKED ISSUE #1579: Replace findActivity()-as-HomeActivity dependency recovery with Hilt EntryPoints / hiltViewModel
## Background

The Compose modernization (single-Activity + Navigation-Compose) lands on the correct architecture overall, but dependency recovery for non-HOME destinations is done by casting the Compose `LocalContext` to the concrete activity and reading its injected members:

```kotlin
val activity = LocalContext.current.findActivity() as HomeActivity
// ...reads activity.prefsRepository, activity.arrivalsViewModelFactory, etc.
```

This pattern repeats across ~8 sites: `ArrivalsDestinations`, `TripDestinations`, `SettingsDestinations`, `ReportDestinations`, `ExtraDestinations`, and the `SettingsRehomeEffect`/`PaymentWarningDialog` host effects. Only the HOME destination has its dependencies threaded in via `HomeDestinationDeps`.

## Why it's worth changing

- **Tight coupling:** every feature graph depends on the concrete `HomeActivity` and its field layout.
- **Type-safety gap:** `as HomeActivity` is an unchecked cast that hard-crashes if the host is ever not `HomeActivity`.
- **Testability:** a destination can't be exercised in isolation without a full Activity; an injected fake would be cleaner.

This is the one place the PR steps off the idiomatic path it otherwise follows. The codebase already uses the idiomatic alternatives elsewhere — `@EntryPoint` interfaces (`LocationEntryPoint`, `RegionEntryPoint`) and assisted factories — so this is "incompletely applied," not unknown.

## Proposed direction

- For shared singleton-ish deps (`prefsRepository`, `regionRepository`, `locationRepository`): expose via `@EntryPoint` interfaces resolved from the application context, the same way `LocationEntryPoint`/`RegionEntryPoint` already work.
- For the assisted `arrivalsViewModelFactory` and the activity-scoped feature ViewModels: use `hiltViewModel(viewModelStoreOwner = activity)` or `hiltViewModel()` scoped to the back-stack entry, instead of reaching through the activity.

Goal: remove the `as HomeActivity` cast from feature graphs entirely.

## Context

Follow-up from the high-level architecture review of the single-Activity Compose modernization (`pr-stack/b1-compose-host`). Not a merge blocker — filed as the top post-merge refactor.

bmander:
Closes #1579.

## What

Dependency recovery for non-HOME Compose destinations cast `LocalContext` to the concrete `HomeActivity` to read its injected members (`findActivity() as HomeActivity`), repeated across ~8 feature graphs. This replaces every cast with the idiomatic alternatives already used elsewhere in the codebase, then does the two coupled cleanups the cast removal exposed.

### #1579: remove the `as HomeActivity` casts

- **Shared singleton deps** (`prefsRepository`, the assisted `arrivalsViewModelFactory`) now resolve via `@EntryPoint` interfaces from the application context — new `PreferencesEntryPoint` and `ArrivalsViewModelFactoryEntryPoint`, mirroring the existing `LocationEntryPoint`/`RegionEntryPoint`.
- **Activity-scoped feature ViewModels** use `hiltViewModel()` (`ArrivalsDestinations`, `SettingsDestinations`, `TripDestinations`, `ExtraDestinations`).

No `as HomeActivity` cast remains in any feature graph.

### Region resolution extraction

Region resolution + the forced region picker move out of `HomeViewModel` into a reactive `RegionPickerViewModel`/`RegionPickerHost`, hosted at the `setContent` root so the picker overlays whichever screen triggered re-resolution (not just HOME). The OTP-version reset moves to `AdvancedSettingsViewModel`. Both are constructor-injected and unit-tested (`RegionPickerViewModelTest`, `RegionOtpResetTest`).

### Navigator removal

The `@Singleton Navigator` route-latch (and `NavigatorEntryPoint` + its test) is removed in favor of idiomatic Navigation-Compose:

- In-app navigation uses hoisted `NavController` lambdas — `NavController.navigateFromHome` captures the pop-to-HOME + `launchSingleTop` invariant once.
- External intents flow through a per-Activity launch-intent channel (`MutableStateFlow<Intent?>` consumed by `LaunchIntentEffect`) rather than a process-scoped latch. `IntentRouteMapper` is unchanged.
- The launcher facades (`TripInfoLauncher`/`NightLightLauncher`) survive only as the standalone activity-alias default. Reminder editing carries a structured `ReminderEditorArgs` (route assembled at the NavController boundary). `HomeCallbacks` holds `HomeActivityActions` whole so an added Activity action can't be silently dropped at the combine site.

## Testing

- `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin :onebusaway-android:compileObaGoogleDebugJavaWithJavac` — green.
- `./gradlew :onebusaway-android:testObaGoogleDebugUnitTest --tests "*IntentRouteMapperTest"` — green; new `RegionPickerViewModelTest` / `RegionOtpResetTest` added.
- On-device pass recommended before merge for the highest-risk paths: a reminder deep link while backgrounded, "show on map" from a pushed list, and the forced region picker over the Advanced-settings screen.

## Summary by CodeRabbit

* **New Features**
  * Added smoother “show on map” actions from arrivals, saved lists, and search results.
  * Added a new manual region selection experience when multiple regions are available.
  * Added support for editing trip reminders from arrivals and saved items.

* **Bug Fixes**
  * Improved app launches from notifications, links, and shortcuts so screens open more reliably.
  * Fixed backup restore and settings flows so region updates refresh correctly.
  * Updated survey, donation, and night-light flows to open consistently from the app.


aaronbrethorst (changes requested):
Brandon, this is the kind of refactor I wish more PRs looked like. There are no correctness bugs here. I'm landing on **Request Changes** for two cheap, in-scope items below — one test-coverage regression and one actively-misleading comment — both tied directly to code this PR moved.

---

## Critical Issues (0 found)

None. The downcasts are fully removed (no `findActivity() as HomeActivity`, `showRouteOnMap`, `focusStopOnMap`, or `arrivalsViewModelFactory` host-reads remain in `src/main`), no dangling references to removed members (`onRegionChosen`, `stageDeepLinkRoute`, `DeepLinkEffect`, `HomeDialog`) survive in live code, and the navigation rewiring is sound.

---

## Important Issues (2 found)

**1. Lost regression coverage: `AdvancedSettingsViewModel.setExperimentalRegions()` wiring is now untested.**
`AdvancedSettingsViewModel.kt:105-118`. The OTP-reset *rule* (`resetOtpVersionOnRegionChange`) was nicely extracted as a pure suspend fun and is well covered by `RegionOtpResetTest` (all four status arms + the deferred-await edge). But the deleted `HomeViewModelTest` cases (`the experimental-regions toggle resets the OTP API version…`) used to guard the *end-to-end wiring*, and that wiring moved into `setExperimentalRegions` with no test following it. Nothing now verifies that the method (a) calls `regionRepository.clear()` only when `!enabled && regionWasExperimental`, (b) persists the toggle pref, or (c) actually invokes `resetOtpVersionOnRegionChange` with `refresh()`'s result at all — a future edit could delete the entire `viewModelScope.launch { … }` block and every test stays green. `RegionOtpResetTest.kt`'s own KDoc even references a `[AdvancedSettingsViewModelTest]` that doesn't exist. `FakeRegionRepository` already supports this (`refreshResult` / `emitState` / `chosen`). Please add the `AdvancedSettingsViewModelTest`.

**2. Stale comment names the deleted `SurveyWebViewLauncher` and describes a flow this PR removed.**
`ExtraDestinations.kt:69-71` (a file this PR edits): the comment says the survey web view is *"Reached in-app from the home survey overlay (SurveyWebViewLauncher facade → HomeActivity → translator)."* This PR deletes `SurveyWebViewLauncher` entirely; `SurveyOverlay` now calls `onOpenSurvey(url)` → `navController.navigateFromHome(NavRoutes.surveyWebView(url))` — a direct in-process navigation, no facade/translator round-trip. A maintainer grepping `SurveyWebViewLauncher` from this comment finds nothing. Update it to describe the `onOpenSurvey`/`NavRoutes.surveyWebView` path.

---

## Suggestions (fit-and-finish)

- **MapReveal KDoc invents a symbol that never existed.** `MapReveal.kt:22-24` says it replaces *"the former `Navigator` route-latch"* — there is no `Navigator` type anywhere (now or on `main`; grep confirms this comment is its only occurrence). The real predecessor was `HomeViewModel.deepLinkRoute` / `stageDeepLinkRoute` and its `DeepLinkEffect` consumer. Since this is brand-new KDoc, worth naming the real thing before it ossifies. (The PR *title*'s "Navigator latch" is fine as informal prose, but a `[Symbol]`-style KDoc reference reads as a type name.)
- **Filename no longer matches contents.** `SurveyWebViewLauncher.kt` now contains only `SurveyWebViewScreen` (the launcher object is gone). Rename to `SurveyWebViewScreen.kt` so finding #2's dead class name isn't resurrected by the filename.
- **AOSP import ordering** (CLAUDE.md mandates AOSP style). `HomeNavHost.kt` has `androidx.annotation.StringRes` (line ~25) after the `androidx.compose.runtime.*` block, and `android.content.Intent` + the `RESULT_MAP_*` imports interleaved within the `org.onebusaway` block (lines ~64, 73-76). Cosmetic but easy.
- **Two out-of-scope provenance comments now dangle** at deleted types: `DonationCard.kt:201` ("Moved here from `HomeDialogs`…") and `HelpViewModel.kt:33` ("Split out of the shared `HomeDialog`…"). Neither was edited here, but this refactor deletes `HomeDialog`/`HomeDialogs`. Drop the dead referent if you're in the neighborhood.
- **`revealStopOnMap` never nulls its lat/lon keys** (`MapReveal.kt:55-59` writes three, `HomeNavHost.kt` only nulls `STOP_ID`). Harmless — the trigger key gates application and all three are always rewritten together — but for symmetry you might null all three on consume. Optionally, collapse the stop reveal's 3-key cluster into a single `@Parcelable StopReveal` so a partial write can't silently no-op, and mark the `RESULT_MAP_*` consts `internal`.
- **Latent stale-callback footgun.** In `SurveyOverlay.kt:90` and `DonationCard.kt:77`, the long-lived `LaunchedEffect(viewModel, lifecycleOwner)` collector closes over `onOpenSurvey` / `onLearnMore`. Not a live bug — both callbacks are `remember`-stable upstream in `HomeNavHost` — but `rememberUpdatedState` would make it robust to a future caller passing an unremembered lambda. (CodeRabbit flagged the same.)
- **Pre-existing, not introduced here:** `ReminderEditorArgs.stopSequence = 0` (`NavRoutes.kt`) uses `0` as the "absent" sentinel, which collides with a valid first-stop sequence (`TripInfoRepository.kt:135` treats a real `0` as "read from DB"). The `0`-means-absent convention already lived in `NavRoutes.tripInfo` on `main`; `ReminderEditorArgs` just mirrors it. Worth a future cleanup to nullable numerics, but out of scope for this PR.

---

## Non-issues I checked (so you don't have to re-litigate)

- **SavedStateHandle reveal consume** — correct: each `LaunchedEffect` is keyed on its trigger, reads, then set-nulls; the re-emit re-runs with a null key and `?.let` no-ops. No recomposition re-fire, no infinite loop, process-death-safe.
- **Launch-intent channel** — `Intent` has no `equals()` override (identity equality), so `MutableStateFlow` conflation never drops a distinct `onNewIntent`; rotation re-seed is gated on `savedInstanceState == null`.
- **OTP await across a forced pick** — safe in practice: the picker is non-dismissible, so the user can't pop the Advanced-settings destination (clearing its VM) while `state.filterIsInstance<Active>().first()` is pending. There's a *very* narrow process-death-mid-pick silent-drop, but it's no worse than the old in-memory `settingsTogglePending` flag — not a regression.
- **`EntryPointAccessors.fromApplication`** — correct accessor for a `SingletonComponent`; `get()` runs once per VM creation inside the `viewModelFactory` initializer, not per recomposition.
- **Nested `with(callbacks){ with(activityActions){ } }`** — no member-name collisions; every unqualified reference binds unambiguously. The "hold `HomeActivityActions` whole" decision is a good guard against silently dropping a new Activity action.
- **`HelpAction` AGENCIES split** — the `onHelpAction` `when` is exhaustive over an `enum` with no `else`, so adding a navigational action forces a *compile error*, not a silent drop. (One review pass flagged this as a silent-drop risk; the exhaustive `when` makes it compile-safe.)

---

## Strengths

- The `@EntryPoint` + companion `get(context)` pattern is exemplary — hides `EntryPointAccessors` behind one typed method, decouples destinations from the concrete host, and the KDoc explains *why* `ArrivalsViewModel` stays `@AssistedInject`.
- Extracting `resetOtpVersionOnRegionChange` as a pure `internal suspend fun` purely to make it unit-testable, then testing all four branches plus the deferred-await edge the old VM tests never reached.
- Honest test deletions: each removed `HomeViewModelTest` case maps to logic that genuinely left the class, with a breadcrumb comment pointing where it went.
- Fixes a real latent `ClassCastException` for standalone hosts as a side effect of removing the casts.
- The KDoc throughout is load-bearing and (mostly) accurate — a rarity in a refactor this size.

---

## Recommended Action

1. Add `AdvancedSettingsViewModelTest` covering `setExperimentalRegions()` (Important #1).
2. Fix the stale `SurveyWebViewLauncher` comment in `ExtraDestinations.kt:69-71` (Important #2).
3. Optional but cheap: the `Navigator` KDoc reference, the `SurveyWebViewLauncher.kt` rename, and the AOSP import ordering.
4. Re-request and I'll give it a quick second pass — this is very close.

aaronbrethorst (approved):
Brandon, this is a genuinely excellent piece of surgery. Untangling ~8 feature graphs from `findActivity() as HomeActivity` is the kind of refactor that usually leaves a trail of subtle breakage, and instead you've landed a clean, internally-consistent migration. The KDoc density is doing real work here — the rationale comments made this reviewable.

I went deep on the parts most likely to hide bugs — the reveal pattern, the intent channel, the deferred OTP reset, the EntryPoint scoping — and they're all correct. Nothing here blocks the merge.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None. The high-risk mechanisms were verified sound:

- **`SavedStateHandle` map reveal** (`MapReveal.kt`, `HomeNavHost.kt:117-145`) — `revealStopOnMap` writes `STOP_ID`/`LAT`/`LON` synchronously before `popBackStack`; HOME's `LaunchedEffect` keys on `STOP_ID` (written first), reads lat/lon non-reactively, then nulls all three so it neither re-fires on recomposition nor loops after process-death restore. `getBackStackEntry(NavRoutes.HOME)` is safe because HOME is the start destination.
- **`LaunchIntentChannel` / `LaunchIntentEffect`** — correctly fixes #1582: the UNLIMITED buffer retains the `onCreate`-staged cold-launch intent, `LaunchedEffect(navController)` collects the single `receiveAsFlow` exactly once, side effects run before navigation, and the `savedInstanceState == null` guard prevents re-fire on config change.
- **Deferred OTP reset** (`AdvancedSettingsViewModel.resetOtpVersionOnRegionChange`) — on `NeedsManualSelection`, state is `NeedsManualChoice` when `filterIsInstance<Active>().first()` begins, so it correctly waits for the user's pick rather than resolving on a stale value. The `Changed`-path reset is idempotent.
- **Hilt EntryPoints** resolve from the singleton component, valid for the `ArrivalsViewModel.Factory`, `PreferencesRepository`, `RegionRepository`, and `LocationRepository`. The `as HomeActivity` removal is complete; the `@Singleton` Navigator latch is fully gone with no orphaned references.

## Suggestions (tracked in #1602, none blocking)

1. **Silent `RegionState.Failed` (pre-existing).** A catastrophic region load surfaces no UI and no log — `refreshRegions()` only reacts to `RegionStatus.Changed`. This predates your PR, but you introduce `RegionState.Failed` as a first-class sentinel and render it nowhere, so it's the natural seam to add a retryable error affordance + logging later.
2. **Partial stop-reveal is dropped *and* consumed** (`HomeNavHost.kt:132-145`). If `STOP_ID` is present but lat/lon null, focus is silently skipped while all three keys are nulled. Only reachable on a corrupted/half-restored handle (the producer always writes all three atomically), so it's latent — a `Log.w` in the `else` would make it findable.
3. **Unbounded await** in `resetOtpVersionOnRegionChange`'s `NeedsManualSelection` branch — bounded by `viewModelScope` and the non-dismissible picker in practice, but a refresh that lands on `Failed` instead of `Active` never resets. Worth a `withTimeoutOrNull` or a KDoc note.
4. **Polish:** a symmetric typed *read* helper for the `RESULT_MAP_*` keys (collapse the stringly-typed produce/consume contract into one file); `FLAG_ACTIVITY_NEW_TASK` or a KDoc precondition on the launcher facades' `startActivity`; `require(...)` non-blank guards on `ReminderEditorArgs`; a `RegionPickerViewModel` test for `NeedsManualChoice → Resolving/Failed` clearing the picker.
5. **One doc nit introduced here:** `HomeDestinationDeps` KDoc still says it carries "the callbacks" — the field was renamed `callbacks` → `activityActions` and now carries only the Activity-bound subset. Trivial to fold in.

## Strengths

- The two-holder `HomeCallbacks` / `HomeActivityActions` split tracks the real seam ("needs NavController" vs "needs Activity services") and holding `activityActions` *whole* means a newly-added Activity action can't be silently dropped at the combine site — a genuine compile-time guard.
- Extracting `applyExperimentalRegionsToggle` / `resetOtpVersionOnRegionChange` as Context-free `internal suspend` functions made the formerly-coupled OTP logic testable in isolation; coverage got *better* as a direct result of the refactor.
- `LaunchIntentChannel<T>`'s generic parameter earns its keep — it makes the #1582 no-drop/in-order contract unit-testable on the JVM without an Android `Intent`, and the burst test (`a, b, a` before any drain) is exactly the regression guard that matters.
- `RegionPickerViewModel` is the best-encapsulated VM in the set: private mutable state, `choose` as the sole mutator delegating to the repository, and a documented reason for the manual `collect` over `stateIn(Eagerly)`.

## Recommended Action

1. Merge as-is.
2. Follow-ups tracked in #1602.

PR #1585, opened 2026-06-27
Cache map marker BitmapDescriptors by logical key in GoogleMapRenderer

### LINKED ISSUE #1580: Cache and recycle map marker bitmaps in GoogleMapRenderer
## Problem

`GoogleMapRenderer` allocates marker `BitmapDescriptor`s on the per-frame reconcile path and never recycles them:

- Vehicle icons are re-created via `BitmapDescriptorFactory.fromBitmap(...)` on every heading-octant change (reconcile runs at ~20 Hz).
- Trip markers lazily allocate `BitmapDescriptor`s that are never released, including in `dispose()`.

The arrow stamp is cached correctly, but the vehicle/trip bitmaps are not.

## Impact

Memory pressure during long route/trip-focus sessions with frequent heading changes — on a busy route (30+ vehicles) this is hundreds of allocations/sec. On devices with a small native heap this can drive GC stalls / native OOM.

## Proposed fix

- Cache vehicle bitmaps keyed by `(direction octant, isRealtime)` so reconcile reuses descriptors instead of reallocating.
- Track allocated bitmaps and `recycle()` them in `dispose()`.

## Context

Follow-up from the high-level architecture review of the single-Activity Compose modernization (`pr-stack/b1-compose-host`). The map seam itself is clean; this is the highest-value performance follow-up.

bmander:
## Problem

`GoogleMapRenderer` allocated marker `BitmapDescriptor`s on the per-frame reconcile path and never released them (#1580):

- Vehicle icons were re-created via `BitmapDescriptorFactory.fromBitmap(...)` on every heading-octant change (reconcile runs at ~20 Hz).
- Trip-focus markers held lazily-allocated `BitmapDescriptor`s never released, including in `dispose()`.

Investigating also surfaced a compounding root cause: `VehicleBitmaps`' underlying bitmap `LruCache` holds only **15** entries, so a busy route (30+ vehicles) *thrashes* it — every lookup misses and re-creates the `Bitmap`. Keying a descriptor cache by `Bitmap` identity would inherit that churn, which is why this caches by a **stable logical key** instead.

## Fix

- **`BitmapDescriptorCache<V>`** (new, Google flavor): an `LruCache` keyed by a stable logical id whose bitmap supplier runs **only on a miss** — so a steady-state hot path does no bitmap decode/wrap at all. Generic over `V` so a test can inject a counting fake. Cleared in `dispose()` to release native textures.
- **`VehicleBitmaps.iconKey()`**: the stable per-icon key (type + heading octant + schedule-deviation color), sharing type/color resolution with `vehicleBitmap()` (cablecar→tram normalization lifted into one helper so the bitmap and key paths can't drift).
- **`GoogleMapRenderer`** routes vehicle and trip-focus icons through the cache; the trip-focus `by lazy val`s become cached lookups released on `dispose()`. Cache sized for a busy route's working set.

Source bitmaps live in shared static caches and are deliberately **not** recycled (other renderer instances / the maplibre flavor reuse them); dropping descriptor references is the correct release.

## Verification

**Allocation regression test** (`VehicleIconAllocationTest`, instrumented): drives the real icon path over a 38-vehicle HART snapshot and asserts descriptor allocations are bounded by the distinct icons and **independent of frame count**.

> 38 vehicles, **5092 icon requests** over a 240-frame session → **5092** `fromBitmap` allocations before, **24** after.

**On-device before/after** (live route, 120 s busy-route session, Pixel 7 Pro):

| Metric | Before (pre-fix) | After (fixed) |
|---|---|---|
| Native heap growth | **+3,112 KB** | **+40 KB** (flat) |
| Total PSS growth | +31,820 KB | −29,659 KB |

The pre-fix build climbs ~3 MB of native heap in two minutes on a busy route; the fixed build stays flat — the accumulation that drives the GC stalls / native-OOM the issue warns about.

## Follow-up (out of scope here)

The maplibre flavor still re-decodes and re-wraps icons per octant-flip against the same 15-entry shared LRU. A cross-flavor fix (raise the shared `VehicleBitmaps` cache to cover a route's working set, and/or a flavor-neutral wrapper cache) is the more durable resolution but is outside #1580's `GoogleMapRenderer` scope.

Closes #1580

## Summary by CodeRabbit

* **New Features**
  * Improved map marker rendering for vehicles and trip stops, with smoother updates during live map refreshes.
  * Added smarter icon reuse so the app can handle frequent map changes more efficiently.

* **Bug Fixes**
  * Reduced unnecessary icon recreation, helping prevent performance issues during longer map sessions.
  * Kept vehicle icons consistent across updates, including heading and status-based changes.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
@@ -1,0 +1,164 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.map
+
+import android.content.Context
+import android.util.Log
+import androidx.test.InstrumentationRegistry.getTargetContext
+import androidx.test.runner.AndroidJUnit4
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.onebusaway.android.extrapolation.extrapolatedVehicles
+import org.onebusaway.android.io.request.ObaTripsForRouteResponse
+import org.onebusaway.android.map.googlemapsv2.BitmapDescriptorCache
+import org.onebusaway.android.map.render.VehicleBitmaps
+import org.onebusaway.android.map.render.VehicleMarker
+import org.onebusaway.android.mock.Resources
+
+/**
+ * The before/after allocation guard for #1580. `GoogleMapRenderer` re-stamps a gliding vehicle's icon on
+ * every heading-octant change (the ~20Hz reconcile path), so before [BitmapDescriptorCache] it minted a
+ * fresh `BitmapDescriptor` + native texture per flip — hundreds/sec on a busy route.
+ *
+ * This drives the *real* icon path (the cached [VehicleBitmaps.vehicleBitmap] feeding the descriptor
+ * cache) over a simulated route session and asserts the property the fix introduces: the number of wrapper
+ * allocations is bounded by the distinct icons and is **independent of how many frames run** — whereas the
+ * naive per-flip allocation count grows with the session. The descriptor factory is a counting fake (the
+ * cache's only allocator), so this needs no real `GoogleMap`.
+ */
+@RunWith(AndroidJUnit4::class)
+class VehicleIconAllocationTest {
+
+    private val context: Context get() = getTargetContext()
+
+    // A real, busy trips-for-route snapshot (38 vehicles across many HART routes, with full route refs so
+    // icon type/color resolve) — the "busy route" the issue calls out.
+    private fun response(): ObaTripsForRouteResponse =
+        Resources.readAs(
+            context,
+            Resources.getTestUri("trips_for_route_hart_5"),
+            ObaTripsForRouteResponse::class.java,
+        )
+
+    /** The live vehicles for every route the snapshot serves, mapped to render markers as RouteMapController does. */
+    private fun vehicles(response: ObaTripsForRouteResponse): List<VehicleMarker> {
+        val routeIds = response.trips
+            .mapNotNull { it.status?.activeTripId }
+            .mapNotNull { response.getTrip(it)?.routeId }
+            .toSet()
+        return extrapolatedVehicles(response, routeIds, nowMs = 1_000_000L) { null }
+            .map { v ->
+                VehicleMarker(
+                    activeTripId = v.status.activeTripId,
+                    point = v.point,
+                    isRealtime = VehicleBitmaps.isLocationRealtime(v.status),
+                    status = v.status,
+                    fixTimeMs = v.fixTimeMs,
+                    bearing = v.bearing,
+                )
+            }
+    }
+
+    /** Tallies for one replay: total icon requests, descriptors minted, and bitmaps decoded. */
+    private class Counts {
+        var requests = 0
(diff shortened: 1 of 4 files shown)


aaronbrethorst (approved):
Brandon — this is a genuinely sharp piece of work. The diagnosis goes a level deeper than the symptom: you didn't just stop re-wrapping descriptors, you noticed that keying a descriptor cache by `Bitmap` identity would have inherited the 15-entry source LRU's thrash and quietly defeated itself, so you keyed by a stable *logical* id instead. The allocation regression test with real before/after numbers (5092 → 24) and the on-device native-heap measurements (+3,112 KB → +40 KB flat) make the case concrete. The docstrings explain *why* throughout, which is exactly where comments earn their keep.

I verified the three things most likely to bite on a change like this, and all three hold:

- **Init ordering is safe.** `descriptorCache` is declared after `vehicleEstimate`/`fastEstimate`/`dataAgeEstimate`, but their `{ ...Icon() }` lambdas are only invoked on first show (inside `TripEstimateMarker.update`), long after construction — no uninitialized-field hazard.
- **Thread-safety is fine.** `BitmapDescriptorCache.get()` is a non-atomic check-then-act, but the whole `renderDynamic`/`renderStatic`/`dispose` chain runs on the main thread (Compose frame callbacks), and gms marker mutation is main-thread-only regardless. The old `by lazy` was synchronized, but it was guarding a race that couldn't occur, so there's no regression.
- **cablecar→tram parity holds.** Lifting the normalization out of `getBitmap` into `vehicleType()` keeps the bitmap path and the new key path agreeing — both feed the same normalized type through `createBitmapCacheKey` in the same argument order. Behavior is unchanged.

CI is green, including the instrumented test on API 33.

## Critical Issues (0 found)

None. The shipping code is correct.

## Important Issues (0 found)

None that block merge. The test-strengthening notes below are about catching *future* regressions, not defects in this PR.

## Suggestions (follow-up, non-blocking)

These are all tracked in #1590 — none gate this PR.

**Test could pass through some real regressions.** The allocation guard is a real before/after test, but a few gaps mean some keying bugs would stay green:

- The `iconKey()` ⟺ `vehicleBitmap()` agreement — the actual correctness invariant the refactor introduces — is never asserted. A key collision would only *lower* the allocation count, so every current assertion still passes. [VehicleIconAllocationTest.kt]
- The `allocations <= 8 * vehicles.size` bound is loose: a regression keying per-`activeTripId` would hit exactly 38×8=304 and pass. Asserting the *sharing* property (e.g. `allocations < vehicles.size`) would actually exercise the cache's value. [VehicleIconAllocationTest.kt:149]
- `BitmapDescriptorCache` eviction and `clear()` have zero coverage (`CACHE_SIZE = 1024` is picked so eviction never fires), yet the docstring makes load-bearing claims about both. The cache is pure Kotlin — these belong in a fast JVM unit test. [VehicleIconAllocationTest.kt:160-162]
- Color/deviation key variation and the cablecar→tram branch aren't exercised by the buses-only fixture.

**Comment polish (the docs this PR sells itself on):**

- Drop the hardcoded "VehicleBitmaps' is 15 entries" — it duplicates `MAX_CACHE_SIZE` from another file and will rot silently. [BitmapDescriptorCache.kt:30]
- KDoc `[bitmap]` and `[maxSize]` won't resolve as Dokka links (a method param and a non-property ctor param). [BitmapDescriptorCache.kt:32,35]
- "decode/tint/wrap" conflates the bitmap supplier with the separate `wrap` lambda. [BitmapDescriptorCache.kt:32]
- Worth documenting the main-thread-only precondition on the class.
- Pre-existing: `directionIndex` Javadoc says "(or the undirected slot)" but `getHalfWindIndex(..., 8)` can only return 0..7. [VehicleBitmaps.java:116]

**Style:** annotate the icon helper return types (`vehicleEstimateIcon`/`fastEstimateIcon`/`dataAgeIcon`/`tripStopIcon`) to match `tripCircleIcon: BitmapDescriptor` and the CLAUDE.md explicit-return-type preference. [GoogleMapRenderer.kt:508-522]

## Strengths

- Root-cause diagnosis (logical-key vs identity-key, given the bounded source LRU) rather than a surface fix.
- Measured impact, both synthetic (allocation test) and on-device (native heap).
- `dispose()` clears the cache, with a correct note on why source bitmaps are deliberately *not* recycled.
- `V : Any` is a deliberate, correctness-bearing bound (a nullable `V` would make a cached value indistinguishable from a miss), and the generic is a legitimate test seam that decouples from the Maps SDK.
- Honest, scoped follow-up section calling out the maplibre flavor as out of scope.

## Verdict

**Merge.** The production change is correct, well-scoped, documented, and CI-green. The test-strengthening and comment items are worth doing but don't belong gating this PR — tracked in #1590.

PR #1586, opened 2026-06-27
Deduplicate trip polylines by shapeId (#1581)

### LINKED ISSUE #1581: Evict cached trip polyline when its trip is evicted from TripState LRU
## Problem

`TripState` caps tracked trips at `MAX_ENTRIES = 100` (access-ordered LRU), but the per-trip polyline cached alongside each trip is not evicted when the trip is evicted. Polylines (ordered location lists with precomputed cumulative-distance arrays) can be sizable for rail/rapid-transit shapes.

## Impact

Slow, unbounded native/heap growth over a long-lived app session as distinct trips are observed. Not a crash; gradual memory creep.

## Proposed fix

Evict the cached polyline when its trip is evicted from the LRU (or give polylines their own bounded cache).

## Context

Follow-up from the high-level architecture review of the speed-estimation trip view (`pr-stack/b1-compose-host`). The cold-flow / SingleFlight data layer is otherwise sound.

bmander:
Closes #1581.

## Background

#1581 reported that the per-trip polyline cached in `TripState` could outlive its trip's eviction from the `TripStateCache` LRU, causing unbounded heap growth.

Tracing the code: the polyline is a **field of `TripState`**, and `TripState` is the value stored in the LRU — so it's already GC'd when the trip is evicted (confirmed: no static retainer, the fetcher's `SingleFlight` clears its entry on completion, and `ensureShape` only writes the polyline back into the trip's own state). Polyline memory was therefore already bounded by the 100-trip LRU, not unbounded.

The **real, weaker** problem — and the issue's parenthetical alternative ("give polylines their own bounded cache") — is that polylines were keyed by `tripId`, not `shapeId`. Many distinct trips share one route/shape but each held its **own** `Polyline` copy, so a full LRU could retain up to ~100 duplicate copies of what is often a handful of distinct (and, for rail, sizable) shapes.

## Change

- New bounded, synchronized, access-ordered `ShapeCache` keyed by shapeId (mirrors `TripStateCache`'s pattern; the data layer stays Android-free / JVM-testable).
- `ensureShape` now resolves one **shared** `Polyline` instance per shape and records that same instance on every trip on the route, fetching each shape once. Concurrent first-misses already coalesce in the fetcher's `SingleFlight` to the same instance, so they stay consistent.
- The polyline still rides inside `TripState` (read API for consumers is unchanged; it still evicts with its trip); `ShapeCache` collapses duplicates and bounds the distinct-instance count. Its bound is coupled to `MAX_TRACKED_TRIPS` so it can never evict a shape a live trip still references (which would re-introduce a duplicate).

## Measured impact (on-device, Pixel 7 Pro)

Real `DefaultTripObservationRepository`, 100 distinct trips all on one shapeId, a fresh 2,000-point polyline minted per fetch:

| Metric | Before | After | Δ |
|---|---|---|---|
| Shape fetches | 100 | 1 | −99 |
| Distinct `Polyline` instances retained | 100 | 1 | −99 |
| Retained Java heap | ~22.2 MB | 288 KB | −98.7% |

Savings scale with concurrent trips-per-shape and shape size (highest on a single busy/rail route view).

## Tests

- `ShapeCacheTest` — hit/miss and LRU eviction (pure JVM).
- `TripObservationRepositoryTest` — new case asserting two trips on the same shapeId trigger a single fetch and share the same `Polyline` instance.

## Summary by CodeRabbit

* **New Features**
  * Added smarter shape handling so repeated trips can reuse the same route geometry instead of loading duplicates.
  * Introduced a bounded cache to keep recently used shapes available and automatically evict older ones.

* **Bug Fixes**
  * Improved consistency when multiple trips share the same shape, helping ensure the same route line is reused across views and updates.

* **Tests**
  * Added coverage for cache hit/miss behavior, eviction, and shared shape reuse across trips.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/ShapeCache.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/ShapeCache.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2024-2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.extrapolation.data
+
+import org.onebusaway.android.util.Polyline
+
+/**
+ * Bound matches the trip LRU's [MAX_TRACKED_TRIPS]: distinct shapes can't outnumber the trips that
+ * reference them, so a cache this size never evicts a shape a live trip still holds — that
+ * guarantee is what keeps the dedup complete (a smaller bound could evict a still-referenced shape
+ * and force a sibling trip to re-fetch a second copy). Orphaned shapes are likewise retained no
+ * longer than the same ceiling.
+ */
+internal const val MAX_CACHED_SHAPES = MAX_TRACKED_TRIPS
+
+/**
+ * A bounded, access-ordered LRU of [Polyline] shapes keyed by shapeId, owned by
+ * [DefaultTripObservationRepository]. Trips that ride the same route share one shapeId, so caching
+ * by shape — rather than letting every trip's [TripState] hold its own copy — collapses those into
+ * a single shared [Polyline] instance, deduplicating the (for rail, sizable) shapes across the up
+ * to [MAX_TRACKED_TRIPS] trips the store tracks.
+ *
+ * Synchronized for the same reason [TripStateCache] is: lookups and writes both promote, so the
+ * access-order map mutates on read, and callers arrive on arbitrary threads — all access must be
+ * serialized.
+ */
+internal class ShapeCache {
+
+    private val shapes =
+            object : LinkedHashMap<String, Polyline>(16, 0.75f, /* accessOrder = */ true) {
+                override fun removeEldestEntry(
+                        eldest: MutableMap.MutableEntry<String, Polyline>
+                ): Boolean = size > MAX_CACHED_SHAPES
+            }
+
+    /** The cached shape for [shapeId], or null if it was never cached (or has been evicted). */
+    @Synchronized
+    fun get(shapeId: String): Polyline? = shapes[shapeId]
+
+    /** Stores [polyline] under [shapeId], promoting it in the retention order. */
+    @Synchronized
+    fun put(shapeId: String, polyline: Polyline) {
+        shapes[shapeId] = polyline
+    }
+}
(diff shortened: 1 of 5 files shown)


aaronbrethorst (approved):
Hi Brandon — Nice work.

The runtime is sound and the memory bound that motivated #1581 holds regardless of cache behavior. The findings below are all about documentation accuracy and test coverage — none block the merge, and I've filed them as a tracked follow-up (#1591) so they don't get lost.

## Non-blocking findings (tracked in #1591)

- **The `MAX_CACHED_SHAPES` doc overstates a guarantee the eviction policy doesn't keep** [`ShapeCache.kt:20-25`]. The comment claims that sizing the cache equal to `MAX_TRACKED_TRIPS` means it "never evicts a shape a live trip still holds." The counting premise (distinct live shapes ≤ live trips ≤ 100) is true, but the conclusion doesn't follow:
  1. **Orphaned shapes occupy slots.** When a trip is evicted from `TripStateCache`, nothing removes its shape from `ShapeCache`. So the 100 slots are shared between live-referenced shapes and orphans — the "shapes ≤ trips" argument ignores the orphans.
  2. **Live shapes are never re-promoted.** `ensureShape` early-returns at `TripObservationRepository.kt:133` once a trip carries its polyline, so `shapeCache.get(shapeId)` is never called again for that live trip. Its access-order position reflects fetch recency, not trip liveness — a long-watched live trip's shape can age to eldest while a recently-fetched orphan survives, and a sibling then re-fetches a duplicate.

  The impact is bounded and benign (one extra copy; memory stays bounded by the 100-trip ceiling), so it's not a functional bug — but the comment is load-bearing design rationale, so it's worth softening to describe best-effort dedup rather than a completeness guarantee. The adjacent "orphaned shapes are likewise retained no longer than the same ceiling" sentence reads as if orphans yield to live shapes — worth dropping or rewording too.

- **The SingleFlight coalescing comment omits its confinement precondition** [`TripObservationRepository.kt:134-136`]. The "concurrent first-misses coalesce ... and resolve to the same instance" claim is accurate for overlapping in-flight misses, but `SingleFlight` is explicitly *not* thread-safe and depends on all callers arriving on a single confined thread (`SingleFlight.kt:33-35`). A half-clause noting coalescing holds for callers on the fetcher's confined dispatcher would close the gap. (Pre-existing — `tripSchedule` already uses the same path — so not a regression here.)

- **A few `ensureShape` branches lack tests.** The TripState short-circuit at `TripObservationRepository.kt:133` is deletable today without any test failing (the dedup test exercises two *different* trips, so it never hits that path). Also worth adding: a different-shapeId case (`shapeCalls == 2`, distinct instances) to pin the keying behavior, and a null-fetch case (`fetcher.shape` returns null → caches nothing → later call refetches) to protect the negative path.

## Strengths

- Honest, well-traced problem reframing with on-device measurements.
- Data layer stays Android-free and JVM-testable; `ShapeCache` cleanly mirrors `TripStateCache`'s synchronized, access-ordered pattern.
- The eviction test (`ShapeCacheTest.kt:42`) correctly promotes via `get()` before overflowing, so it verifies access-order — not just insertion-order — eviction.
- Consistent `assertSame` usage throughout, the right call for an instance-reuse feature.
- AOSP style, license headers, and 100-column wrapping all match the neighboring files; the `MAX_TRACKED_TRIPS` extraction is clean.
- New unit tests pass green (`BUILD SUCCESSFUL`).

## Verdict

1. Merge as-is — the runtime is correct and the memory bound holds.
2. Address the doc reword and test gaps in follow-up #1591 when convenient.

PR #1587, opened 2026-06-27
Queue staged deep-link routes so rapid distinct intents aren't dropped

### LINKED ISSUE #1582: Deep-link route latch can drop rapid back-to-back distinct intents

bmander:
Fixes #1582

## Problem

The staged deep-link route was a `StateFlow<String?>` latch (`HomeViewModel.deepLinkRoute`), set before `setContent` and cleared by `DeepLinkEffect` after navigating. Rapid, *distinct* incoming intents (e.g. FCM A → B → A, or pinned-shortcut bursts) could overwrite the latch before the collector consumed the previous value, dropping a navigation. A `StateFlow` conflates rapid sets, so the intermediate values are lost.

## Fix

Replace the single-value `StateFlow<String?>` latch with an `UNLIMITED`-buffered `Channel<String>`, exposed as `deepLinkRoutes: Flow<String>`:

- Each staged route is delivered **exactly once, in order** — distinct back-to-back intents are no longer collapsed.
- `UNLIMITED` buffering preserves the original "route staged in `onCreate` before the NavHost composes isn't lost" guarantee, and means `stageDeepLinkRoute`'s `trySend` never has to drop.
- `DeepLinkEffect` now drains the flow instead of observing a latch, so the `onDeepLinkRouteConsumed()` clear step is gone (a queue removes items on receive — there's no latch to reset).

This follows the existing `Channel + receiveAsFlow()` one-shot-event convention already used by `SettingsViewModel`, `AdvancedSettingsViewModel`, and `TripInfoViewModel`.

The sibling `showWelcomeTutorial` latch is deliberately left as a `StateFlow<Boolean>` — it's a valueless, idempotent edge-trigger with nothing to drop, so a latch remains the simpler correct choice (documented inline).

## Tests

Adds three `HomeViewModelTest` regression tests that lock in the queue semantics:
- rapid distinct burst (A → B → A) is delivered exactly once, in order
- a route staged before the collector subscribes is not lost
- a null staged route (intent with no mapped destination) enqueues nothing

Confirmed the burst test fails against the pre-fix latch (`expected:<[a, b, a]> but was:<[a]>`) and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deep links now queue and process in order, reducing the chance of missed or overwritten navigation requests.
  * Routes triggered before the app finishes loading are now more reliably handled.
  * Empty deep-link requests are ignored, avoiding unnecessary navigation behavior.

* **Tests**
  * Added coverage for ordered delivery, queued routes, and ignoring empty requests.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.kt
@@ -121,7 +121,7 @@ class HomeActivity : AppCompatActivity() {
             val navController = rememberNavController()
             AccessibilityAnalyticsEffect()
             HomeAnalyticsEffect(viewModel.analyticsEvents)
-            DeepLinkEffect(navController, viewModel.deepLinkRoute, viewModel::onDeepLinkRouteConsumed)
+            DeepLinkEffect(navController, viewModel.deepLinkRoutes)
             // The welcome tutorial (now the Compose green welcome + map-stop spotlight sequence) is
             // started by HomeScreen off the same showWelcomeTutorial latch — no host effect needed.
             SettingsRehomeEffect(navController)
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/HomeNavHost.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/HomeNavHost.kt
@@ -39,6 +39,7 @@ import androidx.navigation.compose.composable
 import androidx.navigation.navArgument
 import org.onebusaway.android.map.compose.TripMapScreen
 import com.google.firebase.analytics.FirebaseAnalytics
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.StateFlow
 import org.onebusaway.android.R
@@ -148,24 +149,21 @@ fun HomeNavHost(
 }
 
 /**
- * Consumes a staged deep-link route (the [HomeViewModel.deepLinkRoute] latch) once the NavHost is ready
- * (and on each `onNewIntent`): navigates to it, popping up to HOME, then clears the latch via
- * [onConsumed]. Lifted verbatim from the former inline `onCreate` effect.
+ * Drains the queue of staged deep-link routes ([HomeViewModel.deepLinkRoutes]) once the NavHost is ready
+ * (and as each `onNewIntent` enqueues more): navigates to each, popping up to HOME. The queue (vs. the
+ * former single-value latch) delivers rapid, distinct back-to-back intents exactly once each (#1582).
  */
 @Composable
 internal fun DeepLinkEffect(
     navController: NavHostController,
-    deepLinkRoute: StateFlow<String?>,
-    onConsumed: () -> Unit,
+    deepLinkRoutes: Flow<String>,
 ) {
-    val pending by deepLinkRoute.collectAsStateWithLifecycle()
-    LaunchedEffect(pending) {
-        pending?.let { route ->
+    LaunchedEffect(navController) {
+        deepLinkRoutes.collect { route ->
             navController.navigate(route) {
                 popUpTo(NavRoutes.HOME) { inclusive = false }
                 launchSingleTop = true
             }
-            onConsumed()
         }
     }
 }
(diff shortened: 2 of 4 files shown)


aaronbrethorst (approved):
Brandon, nice work. I verified the load-bearing concurrency claims directly:
- **Single consumer.** `deepLinkRoutes` is collected at exactly one site (`HomeActivity.kt:124` → `DeepLinkEffect`, `HomeNavHost.kt:161`), so `receiveAsFlow()`'s single-consumer contract holds.
- **No in-flight drop on recreation.** `navController.navigate { … }` is non-suspending, so there's no suspension point between `receive()` and the next `receive()` where cancellation could pull a route off the channel and then lose it. On a config change the `Channel` survives in the ViewModel and the new composition drains whatever's buffered. No loss, no double-nav.

## Critical Issues (0 found)
None.

## Important Issues (0 found)
None block merge. The items below are deliberate follow-ups.

## Suggestions (5 found)
- **Lifecycle awareness changed.** The old latch collected via `collectAsStateWithLifecycle()` (pauses below `STARTED`); the new raw `LaunchedEffect(navController) { collect {} }` will navigate even while the activity is `STOPPED`. Every staging path runs while foregrounded or arrives via `onNewIntent` (which foregrounds the activity), so this almost certainly never bites — but it's a real semantic shift. If you want the prior behavior, wrap the collect in `repeatOnLifecycle(STARTED)` / `flowWithLifecycle`. `HomeNavHost.kt:161`
- **"distinct" understates the guarantee.** The field comment and `DeepLinkEffect` KDoc say the queue delivers "rapid, *distinct* back-to-back intents exactly once each," but the queue preserves *all* routes in order including duplicates — and your own `a, b, a` test relies on that. Dropping "distinct" avoids a future reader assuming back-to-back duplicates get deduped. `HomeViewModel.kt:115`, `HomeNavHost.kt:153`
- **`trySend` comment slightly overstated.** "trySend always succeeds on an UNLIMITED channel" is true only because this channel is never closed (`trySend` also fails on a closed channel). A one-word qualifier ("never-closed UNLIMITED channel") future-proofs it against someone adding a `close()` in `onCleared`. `HomeViewModel.kt:187`
- **`DeepLinkEffect` KDoc omits `launchSingleTop`.** The KDoc mentions `popUpTo(HOME)` but not `launchSingleTop = true`, which can no-op a navigation to a route already on top. A half-sentence keeps the "delivered exactly once" (collection layer) and "may not re-navigate" (nav layer) behaviors from being conflated. `HomeNavHost.kt:151`
- **Two cheap regression tests worth adding.** (1) Consecutive duplicates (`a`, `a`) → assert `listOf("a", "a")` — the cleanest #1582 repro, since a StateFlow latch drops the repeat via distinct-until-changed, and it pins that the queue does *not* dedupe. (2) Re-collection after cancellation: collect → `cancel()` → `stageDeepLinkRoute("x")` → collect again → assert `["x"]`, which locks in both the buffered-across-recreation guarantee and the `receiveAsFlow`-not-`consumeAsFlow` choice. A Compose/Robolectric test of `DeepLinkEffect`'s navigate contract is a nice-to-have beyond that. `HomeViewModelTest.kt:334`

## Strengths
- Clear problem statement with the concrete failure (FCM A→B→A latch conflation) and a tested before/after (`expected:<[a, b, a]> but was:<[a]>`).
- Correct, deliberate asymmetry: queue for the valued, droppable deep-link routes; latch retained for the idempotent welcome-tutorial trigger, documented inline.
- Removed the stale "Lifted verbatim from the former inline `onCreate` effect" line — good comment hygiene, that was rot referencing a transitional state.
- `(#1582)` references in both the field comment and the KDoc give durable "why" context.
- Tests follow the file's established `runTest` / launched-collector-with-`job.cancel()` pattern and target the two right scenarios (burst ordering, pre-subscription buffering).

## Verdict
1. Merge as-is.
2. Track the lifecycle-awareness decision and the extra regression tests as a follow-up — filed as #1592.

Issue #1583, opened 2026-06-25
Extrapolation layer robustness/correctness hardening (SingleFlight confinement, interpolation bounds, clock skew, test gaps)


aaronbrethorst:
Tracking a few low-severity robustness/correctness items found in the speed-estimation extrapolation layer during the architecture review. None are merge blockers.

- **`SingleFlight` main-thread confinement is implicit** (`TripObservationFetcher` / `SingleFlight`): the `inFlight` map relies on main-thread confinement documented only in a comment. A future off-thread caller would corrupt it. Add a thread assertion in `SingleFlight.run`, or use a thread-safe map.
- **`interpolateScheduleTime` boundary** (`TripTrajectory`): `distanceMeters in d0..d1` is inclusive on both ends, so a point exactly on a segment boundary can double-count. Use an exclusive upper bound (`d0..<d1`) or an epsilon.
- **Client/server clock skew in the trajectory view** (`TripTrajectory.kt`, already a `TODO`): `nowMs` uses the local clock while the axis is server-clock; skew shifts the "now" line. Cosmetic (data/debug view). Compute an offset from the latest server/local time pair and thread it through `buildTrajectory()`.
- **Test gaps:** add an end-to-end `TripTrajectoryViewModel` store-hydration test (collect → record → refresh → verify), and an explicit `SingleFlight` concurrent-dedup test.

## Context

Follow-up from the high-level architecture review of the speed-estimation trip view (`pr-stack/b1-compose-host`). The extrapolation math and data layer are well-architected and well-tested for pure logic; these are hardening items.

PR #1588, opened 2026-06-27
Make SingleFlight thread-safe with a ConcurrentHashMap

bmander:
Addresses the **SingleFlight main-thread confinement** item from #1583 (the first of several robustness items tracked there — this PR covers only that one; the interpolation-bounds, clock-skew, and remaining test-gap items are out of scope here).

## Problem

`SingleFlight`'s in-flight map was a plain `HashMap`, safe only because every `run()` call and the scope's dispatcher were confined to one thread (in practice the main thread). That invariant was documented only in a comment, so a future off-thread caller would silently corrupt the map.

## Fix

Replace the `HashMap` with a `ConcurrentHashMap` and coalesce racing callers via its atomic `computeIfAbsent`. This removes the confinement invariant at the root rather than guarding it with an assertion: there is no longer a confining-thread rule for a future caller to violate, so nothing to document, assert, or test against. The dedup contract is unchanged, and the overhead is negligible at this call volume (a handful of in-flight keys, at poll-tick frequency).

`computeIfAbsent` is also marginally cheaper than the old `getOrPut` on the insert path (one atomic map operation vs. a separate get + put).

The KDoc now scopes its guarantee accurately: concurrent callers may invoke `run` safely from any context, but the coalesced block still runs on the constructor `scope`'s dispatcher — not the caller's.

## Tests

Adds `SingleFlightTest` covering the coalescing contract:
- concurrent callers for one key share a single execution (the second joins the first's in-flight `Deferred` rather than re-running the block)
- an entry clears on completion, so a later call runs fresh

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of repeated requests so simultaneous calls for the same key more reliably share a single in-flight operation.
  * Enhanced concurrency safety to reduce the chance of duplicate work when calls happen at the same time.
* **Tests**
  * Added unit tests verifying shared execution for concurrent calls.
  * Added unit tests confirming a fresh execution runs after the previous one completes.
  * Added a test to ensure completed entries clear promptly when using an eager dispatcher.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/TripObservationFetcher.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/TripObservationFetcher.kt
@@ -68,7 +68,7 @@ class DefaultTripObservationFetcher @Inject constructor(
         @ApplicationContext private val context: Context
 ) : TripObservationFetcher {
 
-    /** Process-lifetime scope confined to the main thread; the SingleFlight maps live under it. */
+    /** Process-lifetime scope the coalesced fetches run on; the network calls hop to [fetchDispatcher]. */
     private val fetchScope = CoroutineScope(SupervisorJob() + Dispatchers.Main.immediate)
 
     /**
--- a/onebusaway-android/src/main/java/org/onebusaway/android/util/SingleFlight.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/util/SingleFlight.kt
@@ -16,8 +16,10 @@
 package org.onebusaway.android.util
 
 import android.util.Log
+import java.util.concurrent.ConcurrentHashMap
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.async
 
@@ -30,19 +32,23 @@ import kotlinx.coroutines.async
  * Failures resolve to null for every joined caller (logged once); retry policy is left to
  * callers.
  *
- * Confinement: not thread-safe. All [run] calls and [scope]'s dispatcher must be confined to a
- * single thread (in practice the main thread) — that confinement is what makes the unlocked map
- * safe.
+ * Concurrency-safe: the in-flight map is a [ConcurrentHashMap], whose atomic `computeIfAbsent`
+ * coalesces racing callers, so [run] may be invoked concurrently from any context. The coalesced
+ * block always runs on [scope]'s dispatcher, not the caller's.
  */
 class SingleFlight<K : Any, V : Any>(private val scope: CoroutineScope) {
 
-    private val inFlight = HashMap<K, Deferred<V?>>()
+    private val inFlight = ConcurrentHashMap<K, Deferred<V?>>()
 
     /** Runs [block] for [key] in [scope], or joins the in-flight execution for the same key. */
     suspend fun run(key: K, block: suspend () -> V?): V? =
             inFlight
-                    .getOrPut(key) {
-                        scope.async {
+                    // computeIfAbsent is atomic, so concurrent callers for one key share a Deferred.
+                    // The coroutine is LAZY: await() (below) starts it only after computeIfAbsent has
+                    // returned and stored the Deferred, so the finally/remove can never run
+                    // re-entrantly — even for an eager dispatcher and a non-suspending block.
+                    .computeIfAbsent(key) {
+                        scope.async(start = CoroutineStart.LAZY) {
                             try {
                                 block()
                             } catch (e: CancellationException) {
@@ -51,7 +57,6 @@ class SingleFlight<K : Any, V : Any>(private val scope: CoroutineScope) {
                                 Log.e(TAG, "Single-flight block failed for $key", e)
                                 null
                             } finally {
-                                // Runs on the scope's thread before any awaiter resumes.
                                 inFlight.remove(key)
                             }
                         }
--- a/onebusaway-android/src/test/java/org/onebusaway/android/util/SingleFlightTest.kt
+++ b/onebusaway-android/src/test/java/org/onebusaway/android/util/SingleFlightTest.kt
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2024-2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.util
+
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.async
+import kotlinx.coroutines.test.runCurrent
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+/**
+ * Covers [SingleFlight]'s coalescing contract: concurrent callers for one key share a single
+ * execution, while an entry clears on completion so a later call runs fresh.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+class SingleFlightTest {
+
+    @Test
+    fun `concurrent calls for the same key share one execution`() = runTest {
+        val singleFlight = SingleFlight<String, Int>(backgroundScope)
+        val firstEntered = CompletableDeferred<Unit>()
+        val release = CompletableDeferred<Unit>()
+        var executions = 0
+
+        val block: suspend () -> Int = {
+            executions++
+            firstEntered.complete(Unit)
+            release.await() // hold the in-flight execution open so the second caller has to join it
+            42
+        }
+
+        val a = async { singleFlight.run("k", block) }
+        firstEntered.await() // a is now running block and parked on release
+        val b = async { singleFlight.run("k", block) }
+        runCurrent()
+        release.complete(Unit)
+
+        assertEquals(42, a.await())
+        assertEquals(42, b.await())
+        assertEquals(1, executions) // b joined a's flight rather than running the block again
+    }
+
+    @Test
+    fun `a later call after completion runs the block again`() = runTest {
+        val singleFlight = SingleFlight<String, Int>(backgroundScope)
+        var executions = 0
+
+        // Entries clear on completion, so the second call is a fresh execution, not a joined one.
+        assertEquals(1, singleFlight.run("k") { executions++; 1 })
+        assertEquals(2, singleFlight.run("k") { executions++; 2 })
+        assertEquals(2, executions)
+    }
+
+    @Test
+    fun `an eager dispatcher with a non-suspending block clears the completed entry`() = runTest {
+        // Dispatchers.Unconfined runs the block synchronously to completion before async() returns.
+        // The block never suspends, so this exercises the path where the finally/remove could fire
+        // re-entrantly inside computeIfAbsent — which the LAZY start guards against. A stale entry
+        // would make the second call join the first's completed Deferred instead of running fresh.
+        val singleFlight = SingleFlight<String, Int>(CoroutineScope(Dispatchers.Unconfined))
+        var executions = 0
+
+        assertEquals(1, singleFlight.run("k") { ++executions })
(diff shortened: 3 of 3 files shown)


aaronbrethorst (changes requested):
Hi Brandon — this is a tidy, well-scoped piece of work, and I appreciate that the description draws a sharp line around *which* item from #1583 it covers and which it doesn't. The reasoning in the PR body ("remove the confinement invariant at the root rather than guard it with an assertion") is exactly the right instinct. The new tests are deterministic — gating with `CompletableDeferred` instead of `Thread.sleep` — and both pass locally (`testObaGoogleDebugUnitTest --tests SingleFlightTest`, green).

There's one real problem, though, and it's ironic given the PR's own thesis: the new code trades one silent invariant for a subtler one. Details below.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

**Launching the coroutine *inside* `computeIfAbsent` re-introduces a silent invariant** [`SingleFlight.kt:46-59`]

The `scope.async { ... }` is created inside the `computeIfAbsent` mapping function, and that async body's `finally { inFlight.remove(key) }` mutates the same map. If `block` ever completes **without suspending** under an **eager dispatcher** (`Dispatchers.Main.immediate` when already on the main thread, or `Dispatchers.Unconfined`), the async body runs to completion synchronously *before* `async` returns — so `inFlight.remove(key)` executes re-entrantly, while `computeIfAbsent` is still computing. Two consequences:

1. **Re-entrant `ConcurrentHashMap` mutation.** `computeIfAbsent` documents: *"The mapping function must not modify this map during computation."* Modern JDKs can throw `IllegalStateException: Recursive update`.
2. **Stale entry.** Even where it doesn't throw, the `remove` no-ops (nothing is stored yet), then `computeIfAbsent` installs the already-completed `Deferred`. It now lives in the map permanently, so every later `run(key)` returns the stale cached value and the block never runs again.

**Why it doesn't bite today:** every production block starts with `withContext(fetchDispatcher)` (`TripObservationFetcher.kt:118-119, 134-135`), which always suspends, so the `Deferred` is stored before the `finally` runs. I verified this — the code is correct *for its current caller*.

But that's precisely the issue. The PR's stated goal is to eliminate an unwritten confinement invariant "at the root" rather than document it. This replaces it with a *different* unwritten invariant — "the block must suspend before completing, or the scope's dispatcher must not be eager" — that's just as silent and more surprising. And the KDoc now advertises "`run` may be invoked concurrently from any context," which overstates what the implementation actually guarantees.

The fix is one line and makes the guarantee real for any dispatcher, suspending block or not:

```kotlin
.computeIfAbsent(key) {
    scope.async(start = CoroutineStart.LAZY) {
        // ... unchanged ...
    }
}
.await()   // await() starts the LAZY coroutine *after* it's already in the map
```

With `CoroutineStart.LAZY`, the body can't begin until `await()` is called — strictly after `computeIfAbsent` has returned and stored the `Deferred` — so the `finally`/`remove` can never run re-entrantly. That's the genuine root-level fix the PR is reaching for. Please pair it with a regression test using `Dispatchers.Unconfined` and a non-suspending block, which would have caught this directly.

## Suggestions (3 found)

- **Restore the ordering rationale you removed** [`SingleFlight.kt:55-57`]. Deleting `// Runs on the scope's thread before any awaiter resumes.` took away the one comment that explained *why* the cleanup is race-free. With the type now advertised as concurrency-safe, that rationale matters more, not less. (If you take the `LAZY` fix, a one-line note about why insertion-before-start is what makes removal safe would be even better.)
- **Add a failure-path test.** "Failures resolve to null for every joined caller (logged once)" is contractual but unverified: a throwing block → joined callers receive `null`, the entry clears, and the next call re-runs. A `CancellationException`-is-rethrown test would also pin down the subtle special-case at line 50-51 (which, for the record, is handled correctly — nice).
- **(Out of scope, noting only)** `run`'s `V?` return conflates "succeeded with null," "failed," and "joined a failure." Today's consumers (`tripSchedule`, `shape`) treat absence and failure identically, so it's fine — but worth a sentence in the KDoc since this PR is actively reworking that contract's docs. Not for this PR.

## Strengths

- `CancellationException` is re-thrown before the catch-all — avoids the classic coroutine bug of swallowing cancellation into a `null` result.
- Null-safety is correct by construction: the map holds `Deferred<V?>` (never null), nullability lives on the result, so no `ConcurrentHashMap` null-value pitfall.
- The error path is logged with context, not silent; the exception-to-null conversion is a documented, intentional contract.
- Clean AOSP style, correct import ordering, Apache header on the new test file. No style violations.
- Tests are deterministic and non-flaky — the `CompletableDeferred` gating is the right pattern.

## Recommended Action

1. Switch to `scope.async(start = CoroutineStart.LAZY)` and start via `await()` after insertion (the Important issue).
2. Add a regression test for the eager-dispatcher / non-suspending-block path (`Dispatchers.Unconfined`).
3. Consider the failure-path test and restoring the removed `finally` comment.
4. Re-run the review after the fix.

---

**Verdict: Request Changes.** The change is one line, but it's the heart of the PR: as written, the "concurrency-safe from any context" guarantee in the KDoc isn't one the code delivers, and the new silent invariant is exactly the kind of thing this PR set out to eliminate. The `CoroutineStart.LAZY` fix closes the hole and makes the documentation honest.

bmander:
[[Brandon says: heh Claude went ahead and added this comment without asking me - it's all just Claude talking to itself]]

Thanks for the careful review — you and CodeRabbit both caught the same thing, and you're right that launching the coroutine inside `computeIfAbsent` traded the confinement invariant for a subtler "the block must suspend, or the dispatcher mustn't be eager" one.

Applied the fix exactly as suggested:

- **`CoroutineStart.LAZY` + `await()`** ([`SingleFlight.kt`](https://github.com/OneBusAway/onebusaway-android/blob/71fc8a8c/onebusaway-android/src/main/java/org/onebusaway/android/util/SingleFlight.kt)). The body can't begin until `await()` runs, which is strictly after `computeIfAbsent` has returned and stored the `Deferred`, so the `finally`/`remove` can never run re-entrantly — for any dispatcher, suspending block or not. This makes the KDoc's "concurrency-safe from any context" guarantee real instead of caller-dependent.
- **Restored the ordering rationale** as a comment on the `computeIfAbsent` block, now explaining why insertion-before-start is what makes the removal race-free.
- **Regression test** for the eager-dispatcher / non-suspending path using `Dispatchers.Unconfined` — this fails against the old `scope.async { ... }` and passes with `LAZY`.

On the **failure-path test** suggestion: I tried it, but `SingleFlight` logs via `android.util.Log.e`, which is unmocked in this module's plain JVM unit tests (it deliberately avoids Robolectric/mocking — see `HomeViewModelTest`'s note). Rather than pull in mocking infra or flip the global `returnDefaultValues` just for this, I left a comment marking the failure path as instrumented-coverage territory, consistent with the existing convention. Happy to revisit if you'd prefer the infra.

The `V?` overload-of-meaning note is noted as out of scope for a follow-up.

`testObaGoogleDebugUnitTest --tests SingleFlightTest` is green.

aaronbrethorst (approved):
Brandon, this is a textbook example of fixing a problem at the root instead of papering over it. Deleting the confinement invariant entirely — rather than guarding it with an assertion nobody would ever trip in practice — is the right instinct, and the PR description lays out the reasoning so clearly that I could check each claim against the code and have it hold up. The `CoroutineStart.LAZY` move is the load-bearing subtlety here, and the fact that you anticipated the `computeIfAbsent` re-entrancy hazard *and* wrote a dedicated regression test for it (test 3) is exactly the level of care this kind of change demands.

I dug hard on the concurrency itself, since that's the whole point of the PR, and it's **correct**:

- **`computeIfAbsent` re-entrancy is genuinely prevented.** Creating a LAZY coroutine doesn't run its body, so `block()` and `finally { inFlight.remove(key) }` can't execute while the mapping function is on the stack. `.await()` starts it only after `computeIfAbsent` has returned and stored the `Deferred`. Without LAZY, an eager start on an inline dispatcher (production's `Main.immediate`, or `Unconfined`) with a non-suspending block would call `remove` re-entrantly inside `computeIfAbsent` and throw `IllegalStateException: Recursive update` on JDK 17. The guard is necessary, not decorative.
- **No stale-join race.** `remove(key)` lives in the `finally`, which completes-before the Job transitions to completed, which happens-before any awaiter resumes. So no caller can ever retrieve a *completed* `Deferred` from the map and join it — by the time the `Deferred` completes, its key is already gone. A caller either joins a live flight or creates a fresh one.
- **Scope-cancellation poisoning** (a cancelled LAZY async that never runs its `finally`, leaving a dead key) exists but is **not a regression** — the old eager code had the same property — and `fetchScope` is a process-lifetime `SupervisorJob` that's never cancelled, so it's effectively unreachable.

So the machinery is sound. My reservations are all about the comments and tests added in this same PR — and because this is subtle concurrency code where the comments *are* the spec, a comment that contradicts the code is a real defect, not a nitpick.

## Critical Issues (0 found)

None.

## Important Issues (2 found)

1. **The test-3 comment contradicts the LAZY code it describes.** [`SingleFlightTest.kt:73`] The comment states as fact: `// Dispatchers.Unconfined runs the block synchronously to completion before async() returns.` But the code under test is `scope.async(start = CoroutineStart.LAZY)` — with a LAZY start, `async()` returns a `Deferred` *without running the block at all*; the block runs only when `.await()` is later invoked. The sentence describes the eager-start counterfactual (the hazard the LAZY guard prevents) but phrases it as what actually happens here. A maintainer who reads this line and then sees the hardcoded `CoroutineStart.LAZY` two files over will be justifiably confused. Reword to make clear it's describing what *would* happen under an eager start, e.g. "...were the start eager rather than LAZY, the block would run inline and the finally/remove would fire re-entrantly inside computeIfAbsent." (The rest of the block, lines 74–76, is accurate and valuable — keep it.)

2. **"left to instrumented coverage" claims coverage that doesn't exist.** [`SingleFlightTest.kt:85-87`] I confirmed there is no instrumented test referencing `SingleFlight` anywhere under `src/androidTest` — the failure path isn't "left to instrumented coverage," it's simply untested repo-wide. The comment creates a false impression that the path is verified elsewhere. Two fixes, ideally both:
   - **Reword** so it stops asserting coverage that isn't there (e.g. "...is intentionally left untested here" rather than "left to instrumented coverage").
   - **The deferral is broader than the actual constraint.** Only the `Log.e`→null sub-branch touches the unmocked `android.util.Log`. The other two failure-path guarantees are plain-JVM testable *today* with no Log involved: (a) `CancellationException` propagates (it's rethrown at `SingleFlight.kt:54-55` *before* the `Log.e` line), and (b) the `finally`/remove clears the entry even on throw, so a later call runs fresh. Both are core safety guarantees of this class and both are deterministically assertable without mocking. Worth adding — they're cheap and they cover the contract the PR exists to protect.

## Suggestions (3 found)

- **"eager dispatcher" is imprecise.** [`SingleFlight.kt:49`] Eagerness is a property of `CoroutineStart`, not the dispatcher. The hazard needs an eager *start mode* plus a dispatcher that runs continuations inline/undispatched (`Unconfined`, or `Main.immediate` on the main thread). "an inline/undispatched dispatcher" is the precise phrasing. This is one of the most valuable comments in the diff, so it's worth getting exactly right.
- **Line exceeds AOSP's 100-column limit.** [`TripObservationFetcher.kt:71`] The new KDoc line is 107 chars. The content is accurate (a correct de-rotting of the old "confined to the main thread" comment) — just wrap it.
- **KDoc "always runs on [scope]'s dispatcher, not the caller's" slightly overstates for `Unconfined`.** [`SingleFlight.kt:36-37`] When `scope`'s dispatcher is `Unconfined` (as in test 3), the LAZY coroutine starts inline on whichever thread calls `await()` first. Technically `Unconfined` *is* scope's dispatcher so the letter holds, but "not the caller's" is a touch optimistic. Pedantic — flagging only because this PR's own test uses an `Unconfined` scope.

## Strengths

- Root-cause fix: removes the confinement invariant rather than guarding it, so there's nothing left for a future caller to violate.
- Test 3 is an exemplary regression guard — it pins the exact `LAZY`-start invariant, fails loudly under both the pre-PR and any intermediate implementation, and documents why.
- Deterministic coroutine tests done right: `runTest` + `backgroundScope` (correct lifecycle for a long-lived scope), `CompletableDeferred` handshakes + `runCurrent()` for ordering, no wall-clock flakiness.
- `computeIfAbsent` over `getOrPut` is both atomic and marginally cheaper on the insert path — a real, if small, win.
- Tests assert behavior (execution counts, fresh-vs-joined results), not implementation details.

## Recommended Action

The implementation is correct, so I'm merging this as-is. None of the findings touch runtime behavior — they're comment accuracy and a couple of cheap test additions — so they don't need to block the fix from landing. They're all addressed in a fast-follow PR rather than churning this one:

1. Reword the two misleading comments (`SingleFlightTest.kt:73` and the failure-path note).
2. Add the two JVM failure-path tests (cancellation propagation, entry-cleared-after-cancellation) that the LAZY guard's contract deserves.
3. Sweep the suggestions (terminology, the 107-char line wrap).

The non-cancellation `Log.e`→null branch stays out of scope (it needs `Log` mocking / an instrumented test); the follow-up's comment now says so honestly instead of claiming coverage that doesn't exist.

---

**Verdict: Merge.** The concurrency fix is correct and well-tested — the LAZY-start reasoning holds, there's no stale-join race, and the regression guard is exactly on point. The only findings are documentation precision and a couple of low-cost test additions, none of which affect behavior, so they're handled in a fast-follow rather than holding up the fix. Nice work.

PR #1589, opened 2026-06-27
Make interpolateScheduleTime segments half-open so boundaries don't double-count

bmander:
Addresses the **`interpolateScheduleTime` boundary** item from #1583 (one of several robustness items tracked there — this PR covers only that one; the SingleFlight item is in #1588, and the clock-skew and remaining test-gap items are out of scope here).

## Problem

`interpolateScheduleTime` tested each segment with `distanceMeters in d0..d1` — inclusive on both ends. A distance exactly on an interior stop boundary (`d1` of one segment equals `d0` of the next) matched **two** adjacent segments at once. Forward iteration disambiguated only by accident (the earlier segment won, returning that stop's *arrival*), so the result depended on iteration order rather than on a well-defined ownership rule.

## Fix

Give each segment the half-open span `[d0, d1)`, so a boundary distance lands in exactly one segment (the next one — reading the stop's *departure*) deterministically by construction. The final segment stays closed `[d0, d1]` so the trip's end distance still interpolates to the last stop's arrival rather than dropping to the `0` sentinel. The degenerate-segment guard (`d1 <= d0`, which also prevents a divide-by-zero) is lifted into an early `continue` for clarity.

## Behavior impact

- **Interior boundaries:** arrival → departure (the intended de-double-counting).
- **Strictly-interior distances, both endpoints, out-of-span, and < 2 stops:** unchanged — the existing tests pass untouched (two-stop schedules have no interior boundary).

## Tests

- An interior-boundary regression test (a distance exactly on the middle stop now reads the departure deterministically; the old inclusive code returned the prior segment's arrival).
- A final-endpoint test guarding that the closed last segment keeps the trip's end distance interpolatable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip time interpolation at stop boundaries so exact interior-stop distances resolve consistently.
  * Corrected final-stop handling to always return the last valid scheduled arrival time, even with trailing degenerate entries.
  * Added more robust handling for degenerate/overlapping schedules and non-finite distances.
* **Tests**
  * Expanded unit tests covering sentinel behavior, exact boundary mapping (first/interior/final), skipped degenerate segments, and segment construction rules.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
@@ -180,8 +180,16 @@ private fun extrapolationSeries(state: TripState, schedule: List<ScheduleStop>, 
     )
 }
 
-/** A strictly-increasing distance span `[d0, d1)` mapping to the server-clock time span `[t0, t1]`. */
-internal data class ScheduleSegment(val d0: Double, val d1: Double, val t0: Long, val t1: Long)
+/**
+ * A segment with strictly increasing distance (`d0 < d1`), spanning departure time [t0] at [d0] to
+ * arrival time [t1] at [d1]. Boundary ownership (which segment claims a distance exactly on [d0]/[d1])
+ * is not a property of the segment — it lives in [interpolateScheduleTime].
+ */
+internal data class ScheduleSegment(val d0: Double, val d1: Double, val t0: Long, val t1: Long) {
+    init {
+        require(d1 > d0) { "ScheduleSegment requires strictly increasing distance, got d0=$d0 d1=$d1" }
+    }
+}
 
 /**
  * The interpolatable segments of [schedule], in trip order, with strictly increasing distance.
@@ -210,22 +218,24 @@ internal fun scheduleSegments(schedule: List<ScheduleStop>): List<ScheduleSegmen
  * Pure, so it is unit-testable. Returns 0 when fewer than two stops exist or the distance falls
  * outside every interpolatable segment.
  *
- * Each segment owns the half-open span `[d0, d1)`, so a distance exactly on a stop boundary lands in
- * a single segment (the next one) instead of being ambiguously claimed by both. The terminal segment
- * is closed `[d0, d1]` so the trip's end distance still interpolates rather than dropping to the 0
- * sentinel. Operating on [scheduleSegments] (strictly increasing by construction) means there is no
- * degenerate case left to handle here.
+ * Each segment is half-open `[d0, d1)`, so a distance exactly on a stop boundary lands in a single
+ * segment (the next one) instead of being ambiguously claimed by both. The trip's final distance,
+ * which every half-open span excludes, is mapped after the scan to the last segment's arrival so the
+ * end of the trip still resolves rather than dropping to the 0 sentinel. Operating on
+ * [scheduleSegments] (strictly increasing by construction) means there is no degenerate case left to
+ * handle here.
  */
 fun interpolateScheduleTime(schedule: List<ScheduleStop>, distanceMeters: Double): Long {
     val segments = scheduleSegments(schedule)
-    for ((index, s) in segments.withIndex()) {
-        val withinUpper = if (index == segments.lastIndex) distanceMeters <= s.d1 else distanceMeters < s.d1
-        if (distanceMeters >= s.d0 && withinUpper) {
+    for (s in segments) {
+        if (distanceMeters >= s.d0 && distanceMeters < s.d1) {
             val fraction = (distanceMeters - s.d0) / (s.d1 - s.d0)
             return s.t0 + (fraction * (s.t1 - s.t0)).toLong()
         }
     }
-    return 0L
+    // Half-open spans exclude the trip's final distance; map that exact endpoint to the last arrival.
+    val last = segments.lastOrNull() ?: return 0L
+    return if (distanceMeters == last.d1) last.t1 else 0L
 }
 
 /**
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
@@ -191,7 +191,13 @@ fun interpolateScheduleTime(schedule: List<ScheduleStop>, distanceMeters: Double
     for (i in 1 until schedule.size) {
         val d0 = schedule[i - 1].distanceMeters
         val d1 = schedule[i].distanceMeters
-        if (distanceMeters in d0..d1 && d1 > d0) {
+        if (d1 <= d0) continue
+        // Each segment owns the half-open span [d0, d1), so a distance exactly on a stop boundary
+        // lands in a single segment (the next one) instead of double-counting into both. The final
+        // segment is closed [d0, d1] so the trip's end distance still interpolates rather than
+        // dropping to the 0 sentinel.
+        val withinUpper = if (i == schedule.lastIndex) distanceMeters <= d1 else distanceMeters < d1
+        if (distanceMeters >= d0 && withinUpper) {
             val fraction = (distanceMeters - d0) / (d1 - d0)
             val t0 = schedule[i - 1].departureMs
             val t1 = schedule[i].arrivalMs
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
@@ -180,36 +180,49 @@ private fun extrapolationSeries(state: TripState, schedule: List<ScheduleStop>, 
     )
 }
 
+/** A strictly-increasing distance span `[d0, d1)` mapping to the server-clock time span `[t0, t1]`. */
+internal data class ScheduleSegment(val d0: Double, val d1: Double, val t0: Long, val t1: Long)
+
+/**
+ * The interpolatable segments of [schedule], in trip order, with strictly increasing distance.
+ *
+ * Stops sharing a distance (GTFS shape_dist_traveled ties, or backward data) collapse rather than
+ * forming zero-length segments: the segment *entering* a distance ends at the first such stop's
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Hi Brandon — this is a satisfying little fix. Replacing the accidental "whoever-iterates-first-wins" behavior with an explicit, by-construction ownership rule for boundary distances is exactly the right instinct, and the half-open `[d0, d1)` / closed-final framing is the clean way to express it. The two new tests are genuinely excellent: the interior-boundary test asserts `7_000L` *specifically because* it disambiguates from the old `6_000L`, so it pins the contract rather than the implementation. The block comment explains *why* the asymmetry exists instead of restating the code, which is the kind of comment that resists rot.

There's one issue I'd like resolved before this merges, because it defeats one of the PR's own stated goals.

## Critical Issues (1 found)

**The trip-end distance drops to the `0` sentinel when the final segment is degenerate — a regression that contradicts this PR's own comment.** `TripTrajectory.kt:194,199-200`

The `if (d1 <= d0) continue` guard at line 194 runs *before* the closed-final-segment logic at line 199. The "final segment stays closed" protection is keyed on `i == schedule.lastIndex`, but if that last index pair is itself degenerate, it gets `continue`d past and the closed `<= d1` branch never executes. Every earlier segment uses the exclusive `distanceMeters < d1`, so the maximum distance value is excluded from the segment that reaches it and claimed by no later one — the loop falls through to `return 0L`.

Concrete case with trailing duplicate-distance stops (distances `0, 100, 100`):

```
schedule = [stop(0.0, ...), stop(100.0, arrive=6000), stop(100.0, arrive=10000)]
interpolateScheduleTime(schedule, 100.0)
  i=1 (0→100):   interior, 100 < 100 is false        → no match
  i=2 (100→100): d1 <= d0                             → continue
  → return 0L
```

The **old** code matched at `i=1` (`100 in 0..100 && 100 > 0`), interpolated with `fraction = 1.0`, and returned the arrival (`6000`). So the new code returns the `0` sentinel where the old code returned a valid time — a behavioral regression that directly defeats stated goal #1 ("the trip's end distance still interpolates rather than dropping to the `0` sentinel"). It also makes the comment at lines 196–198 untrue for this input.

This is reachable in practice: schedules are built from `st.distanceAlongTrip` (`TripTrajectory.kt:141`), and GTFS permits stops sharing `shape_dist_traveled`, so coarse or repeated distance data near the trip end is a realistic condition. When it happens, a vehicle at the end of the trip would get a `0` scheduled time and a garbage deviation label.

Suggested fix — tie the closed upper bound to the last *interpolatable* segment rather than the literal last index:

```kotlin
val lastInterpolatable = schedule.indices.lastOrNull { i ->
    i >= 1 && schedule[i].distanceMeters > schedule[i - 1].distanceMeters
}
// ...
val withinUpper = if (i == lastInterpolatable) distanceMeters <= d1 else distanceMeters < d1
```

This keeps the half-open interior behavior while ensuring the true terminal segment stays closed even when degenerate stops trail it. Please add a regression test (trailing duplicate-distance stops, querying the max distance) so this stays caught.

## Suggestions (3 found)

- **Add a symmetric first-stop boundary test.** `TripTrajectoryTest.kt`. The span is lower-*inclusive*, so a vehicle exactly at the trip origin (`interpolateScheduleTime(schedule, 0.0)`) should resolve to the first segment's start time, not the sentinel. This is the natural partner to the new final-stop test and guards the `distanceMeters >= d0` lower bound. Worth adding while you're in here.

- **Cover the new `continue` skip path for an *interior* degenerate segment.** `TripTrajectoryTest.kt`. A schedule like `[0, 100, 100, 200]` queried at `150.0` would lock in that the loop advances past a degenerate middle pair and interpolates in the later valid segment. The existing tests never construct a degenerate segment at all, so this new branch is currently exercised only indirectly.

- **Tighten "double-counting" wording.** `TripTrajectory.kt:196`. The loop returns on first match, so the old code didn't literally count a boundary twice — it silently handed ownership to the *earlier* segment. The test comment's framing ("ambiguously matching the prior segment's arrival") is more precise; consider mirroring it in the source comment (e.g. "instead of being ambiguously claimed by both").

## Strengths

- The interior-boundary test is contract-focused and would fail if anyone reverted the bound to inclusive `<=` — a real guard, not a tautology.
- Both new tests assert specific values chosen to disambiguate from the buggy result, with inline comments explaining the *why*.
- The block comment captures the rationale (deterministic ownership + preserving end-distance interpolation) rather than narrating the code.
- The change is backward compatible with the pre-existing tests, which still pass untouched (confirmed locally via `testObaGoogleDebugUnitTest`).

## Recommended Action

1. Fix the degenerate-final-segment regression (tie the closed bound to the last interpolatable segment) and add a trailing-duplicate-distance regression test.
2. Consider the symmetric first-stop and interior-degenerate tests while you're there.
3. Optionally tighten the "double-counting" comment wording.
4. Re-run the review after fixes.

aaronbrethorst (dismissed):
Hi Brandon — this is a tidy, well-scoped fix. The PR description is a model of how to frame a behavior change: it names the exact ambiguity (a boundary distance matching two adjacent segments, disambiguated only by iteration-order accident), states the new ownership rule precisely, and is honest about what's in and out of scope relative to #1583. The `scheduleSegments` extraction is the right move — collapsing the degenerate-pair guard, the divide-by-zero protection, and the boundary-ownership rule into one place that's independently testable. I re-derived the arithmetic in every new test and ran the suite locally (`TripTrajectoryTest`, green). The core logic is correct.

Two small things I'd like addressed before this goes in, plus a couple of optional suggestions.

## Critical Issues (0 found)

None. The half-open interior boundary and closed terminal-segment handling are correct, including the subtle "terminal = last *interpolatable* segment" case that keeps trailing-degenerate trips from falling through to the `0L` sentinel.

## Important Issues (2 found)

- **The `ScheduleSegment` KDoc claims a half-open span the type neither owns nor consistently obeys.** [TripTrajectory.kt:183] The one-liner reads `A strictly-increasing distance span [d0, d1) ...`, but (a) `ScheduleSegment` is just four numbers and enforces no interval semantics, and (b) its sole consumer treats the *terminal* segment as closed `[d0, d1]` (line 222) — which is the whole point of three of your new tests. A reader looking at the type in isolation would conclude `d1` is always exclusive, which is false for exactly the last segment. The `interpolateScheduleTime` KDoc just below (lines 213–215) gets this exactly right, which makes the type-level line the contradicting outlier. Reword it to state only what the type guarantees and leave boundary ownership where it actually lives, e.g. `A segment with strictly increasing distance (d0 < d1), spanning departure time t0 at d0 to arrival time t1 at d1.`

- **The lower-bound guard `distanceMeters >= s.d0` is dead-untested, and it protects a production-reachable silent-wrong-answer path.** [TripTrajectory.kt:223] The existing out-of-span test only covers a distance *above* the top of the span (150 on a `[0,100]` schedule). Nothing exercises a distance *below* the first segment's `d0`. If that guard were ever dropped in a refactor, a below-origin median (e.g. `-10.0`, or `40.0` on a schedule whose first stop is at distance `50`) would compute a negative fraction and return a time *before* the first departure instead of `0L` — a silent wrong time, not a crash. This is reachable: `extrapolationSeries` feeds the distribution median straight in (line 179), and a median below the first scheduled stop is plausible early in a trip. Add a test asserting `interpolateScheduleTime(schedule, -10.0) == 0L`, ideally also one with a non-zero first `d0`.

## Suggestions (3 found)

- **Lift the terminal-closed case out of the loop.** [TripTrajectory.kt:219-229] The `withinUpper` ternary recomputes `segments.lastIndex` every iteration to control a single point — whether `distance == last.d1` maps to `t1` instead of the sentinel. That point can be handled once, after a uniform half-open scan, which also reads exactly like the KDoc narrates it:
  ```kotlin
  fun interpolateScheduleTime(schedule: List<ScheduleStop>, distanceMeters: Double): Long {
      val segments = scheduleSegments(schedule)
      for (s in segments) {
          if (distanceMeters >= s.d0 && distanceMeters < s.d1) {
              val fraction = (distanceMeters - s.d0) / (s.d1 - s.d0)
              return s.t0 + (fraction * (s.t1 - s.t0)).toLong()
          }
      }
      // Half-open spans exclude the trip's final distance; map that exact endpoint to the last arrival.
      val last = segments.lastOrNull() ?: return 0L
      return if (distanceMeters == last.d1) last.t1 else 0L
  }
  ```
  Behavior-identical, including the backward-data overlap precedence (the endpoint check stays *after* the loop, so an earlier overlapping segment still wins first-match). Optional — the current form is correct, just busier.

- **Consider enforcing the strict-ordering invariant on the type.** [TripTrajectory.kt:184] An `init { require(d1 > d0) { ... } }` would make "strictly increasing by construction" literally true rather than documentation-backed, turning any future out-of-factory construction with `d1 <= d0` into a loud failure instead of a silent divide-by-zero downstream. Low priority given `internal` visibility and the single factory.

- **`scheduleSegments` "strictly increasing" wording.** [TripTrajectory.kt:185] True per-segment, but with non-monotonic/backward distance data the list as a whole can overlap (0→100→50→150 yields `[0,100)` and `[50,150)`). `interpolateScheduleTime` resolves overlaps deterministically by first-match — identical to pre-PR behavior, so not a regression — but the global "strictly increasing" phrasing is slightly imprecise. Minor.

## Strengths

- The de-double-counting is genuinely fixed: distance exactly on an interior stop now reads the departure deterministically by construction, not by iteration-order luck.
- Test coverage is thorough and DAMP — each test names the behavior it pins and explains *why* a specific number is expected. The run-collapse test asserting both resulting segments pins the entire "first-arrival-in, last-departure-out" contract in one place.
- The `scheduleSegments` KDoc documents the non-obvious GTFS `shape_dist_traveled` tie behavior — exactly the kind of "why" future maintainers need.

## Recommended Action

1. Reword the `ScheduleSegment` KDoc so it no longer asserts a `[d0, d1)` interval the type doesn't own (important).
2. Add a below-origin / negative-distance test asserting the `0L` sentinel (important).
3. Optionally take the loop simplification and the `require(d1 > d0)` guard.
4. Re-run `TripTrajectoryTest` after the changes.

**Verdict: Request Changes** — both important items are quick, and both are tightly coupled to this exact change (a misleading contract on the new type, and a missing guard test for a path the new code makes reachable), so they're better fixed here than deferred.

PR #1594, opened 2026-06-27
Fix client/server clock skew in the trajectory now line

bmander:
## Summary

Addresses the third item in #1583: client/server clock skew in the trajectory view (`TripTrajectory.kt`).

The trajectory view's "now" line is laid out on a **server-clock** time axis, but `buildTrajectory()` was handed `System.currentTimeMillis()` (a **local** clock). Any client/server skew therefore shifted the now line off the latest observation and corrupted the schedule-deviation label (computed as now minus scheduled time, mixing clock domains).

## What changed

- Add `TripState.toServerClock(localTimeMs)`, which lifts a local-clock instant onto the server clock using the skew measured at the anchor (its paired `anchorTimeMs` / `anchorLocalTimeMs`). Falls back to no shift when no anchor data exists yet (skew unmeasurable).
- `buildTrajectory()` converts the local "now" through it for everything plotted (the now line, bounds, and the extrapolation overlay's `nowMs`), while `TripState.extrapolate()` still receives the local time it expects — it compares against the anchor's *local* clock, so feeding it server time would mis-classify fresh data as stale.
- The conversion lives on `TripState`, the type that already pairs both clock domains, so the skew has a single source of truth.

This removes the standing `TODO` carried over from the upstream view.

## Tests

- `TripStateTest`: `toServerClock` passthrough with no anchor, and a +5s skew lift.
- `TripTrajectoryTest`: the now line stays on the local clock when no skew is known, and is lifted onto the server clock by the anchor skew.

All affected unit tests pass (`./gradlew :onebusaway-android:testObaGoogleDebugUnitTest`).

Refs #1583 (does not close it — the other items in that issue remain).

## Summary by CodeRabbit

* **New Features**
  * Time-based trajectory calculations now align the current position marker with the server clock when timing information is available.
  * Added a helper to translate local device time into the server-time domain for more consistent plotting.

* **Bug Fixes**
  * Improved handling when no timing anchor is available, keeping timestamps unchanged instead of shifting them.
  * Updated trajectory rendering so the “now” point stays within the displayed time range using the corrected clock source.

* **Tests**
  * Added coverage for both anchored and unanchored time-conversion behavior.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/TripState.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/TripState.kt
@@ -173,6 +173,15 @@ data class TripState(
     fun withSchedule(schedule: ObaTripSchedule): TripState =
             if (this.schedule == null) copy(schedule = schedule) else this
 
+    /**
+     * Converts a local-device-clock instant to the server clock, using the skew measured at the
+     * anchor (its paired [anchorTimeMs]/[anchorLocalTimeMs]). Returns [localTimeMs] unchanged when
+     * no anchor data exists yet, since no skew can be measured. Lets consumers plot a local "now"
+     * against server-clock data (schedule, observations) without it drifting under clock skew.
+     */
+    fun toServerClock(localTimeMs: Long): Long =
+            if (anchorLocalTimeMs > 0) localTimeMs + (anchorTimeMs - anchorLocalTimeMs) else localTimeMs
+
     fun extrapolate(queryTimeMs: Long): ExtrapolationResult {
         val currentAnchor = anchor ?: return ExtrapolationResult.NoData
         val lastDist = currentAnchor.distanceAlongTrip ?: return ExtrapolationResult.NoData
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/dataview/TripTrajectory.kt
@@ -61,10 +61,10 @@ data class TripTrajectory(
     val extrapolation: ExtrapolationSeries? = null,
     val bounds: DataBounds = DataBounds(0.0, 0.0, 0L, 0L),
     /**
-     * "Now" the graph was built at — the horizontal now line, rising over time. NOTE: the caller
-     * passes a local clock (System.currentTimeMillis), but the time axis is laid out in server-clock
-     * values, so client/server skew shifts the now line off the latest observation. (Carried over
-     * from the upstream view's TODO; fix by offsetting via a history entry's server/local time pair.)
+     * "Now" the graph was built at — the horizontal now line, rising over time. Already in the
+     * server-clock domain the time axis is laid out in: [buildTrajectory] converts the caller's
+     * local clock using the anchor's server/local skew, so the now line tracks the latest
+     * observation even when the device clock drifts from the server's.
      */
     val nowMs: Long = 0L,
 )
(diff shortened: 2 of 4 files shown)


aaronbrethorst (approved):
Nice work, Brandon — this is a clean, surgical fix for a genuinely subtle class of bug. Putting the conversion on `TripState`, the one type that already pairs both clock domains, gives the skew a single source of truth, and the decision to keep feeding `extrapolate()` the *local* clock while lifting only what gets plotted is exactly right. The doc comments carry their weight, and removing the inherited upstream TODO is fully justified now that the thing it warned about is actually fixed.

I verified the arithmetic end to end, including the downstream payoff you called out: `TripTrajectoryScreen.kt:277` computes `formatDeviationLabel((series.nowMs - series.scheduleAtMedianMs) / 1000)`, and with `series.nowMs` now server-clock (via `toServerClock`) and `scheduleAtMedianMs` already server-clock, that label is correct for the first time. The 42 affected unit tests pass locally.

This is a **Merge**. The fix is correct and ready to land. There's one real test gap — the new tests guard the now line but not the extrapolation overlay, which is where the deviation-label bug actually lives — but it's a coverage follow-up rather than a defect in this change, so I've filed it as #1604 rather than block on it.

## Critical Issues (0 found)

None. The clock-skew arithmetic and domain handling are correct.

## Important Issues (1 found — tracked in #1604, not blocking)

- **The overlay `nowMs` lift — the deviation-label payoff — has no test coverage** [TripTrajectory.kt:181]. Both new `TripTrajectoryTest` cases build state via `TripState("trip1").copy(anchorTimeMs=…, anchorLocalTimeMs=…)` with no `anchor`/`history`, so `extrapolate()` returns `NoData` and `extrapolation` is null. That means line 181 (`nowMs = state.toServerClock(nowMs)`) is never exercised: reverting it to `nowMs = nowMs` — the pre-fix behavior — passes every test in this PR. Since the corrupted schedule-deviation label is the user-facing symptom you set out to fix, a future test should produce a non-null `ExtrapolationSeries` (drive a real anchor through `withStatus` so extrapolation succeeds) and assert `extrapolation.nowMs` is lifted by the skew, ideally also asserting the lifted now line lands within `trajectory.bounds`. Tracked in **#1604**.

## Suggestions (3 found — also folded into #1604)

- **Cover negative skew (server behind local)** [TripStateTest.kt:51, TripTrajectoryTest.kt:126]. Every new test uses `+5s` (server ahead). The subtraction branch — where `toServerClock` must *lower* the instant — is never exercised, and a flipped sign in a future refactor would slip through. Cheap to add.
- **Tighten the `nowMs` KDoc** [TripTrajectory.kt:64-67]. "Already in the server-clock domain the time axis is laid out in:" is a garden-path clause that needs a re-read, and "tracks the latest observation" slightly overclaims — the now line rises *ahead* of the anchor by design; it shares the observations' clock frame rather than sitting on the latest one. Consider: "Already in the server-clock domain the time axis uses. [buildTrajectory] converts the caller's local clock via the anchor's server/local skew, so the now line stays in the same clock frame as the observations even when the device clock drifts from the server's."
- **Phrase the `toServerClock` precondition to match the guard** [TripState.kt:178-179]. The prose says "when no anchor data exists yet," but the code guards on `anchorLocalTimeMs > 0` — the same invariant `extrapolate()` uses at line 188. Tying the wording to the actual field (and cross-referencing `extrapolate`) makes the shared guard explicit. Also worth a glance: line 183 is ~104 columns, just over the AOSP 100-column limit — in line with a few existing long lines here, so optional.

## Strengths

- Single source of truth for the skew, placed on the type that owns both clock domains.
- Correctly preserves the local-clock contract for `extrapolate()` while lifting only plotted values — the dual-domain split is the easy thing to get wrong, and you got it right.
- Test comments are exemplary: concrete numbers ("server 105_000 / local 100_000 → server runs 5s ahead") make the skew direction unambiguous and self-documenting.
- The fix transparently corrects the deviation label downstream without touching that code.

## Recommended Action

1. Merge as-is — the fix is correct and the tests are green.
2. Pick up the test-coverage follow-up (overlay path, bounds containment, negative skew) and the two comment-clarity tweaks in **#1604**.

PR #1595, opened 2026-06-27
Add trajectory store-hydration and SingleFlight concurrent-dedup tests

bmander:
Addresses the **test gaps** item from #1583 (the extrapolation-layer hardening tracker). This PR adds the missing tests only — it does **not** close the issue, since the other items there (SingleFlight thread assertion, `interpolateScheduleTime` boundary, clock-skew offset) are separate code changes.

## What's added

- **End-to-end `TripTrajectoryViewModel` store-hydration test** — exercises the full `collect → record → refresh → verify` path: the ViewModel collects the details stream, a recorded snapshot hydrates the store, and `refresh()` rebuilds the UI state from it. The pre-existing ViewModel tests used a static-snapshot fake and never covered this integration.
- **`SingleFlight` concurrent-dedup tests** (new `SingleFlightTest`) — concurrent same-key calls share a single execution, distinct keys run independently, and an in-flight entry self-clears so a later call runs fresh. Driven on a single virtual thread (the confinement the unlocked map requires) and kept on the success path so `android.util.Log` is never hit (no Robolectric needed).

## Cleanup folded in

- Adds a shared `testTripStatus()` JVM fixture in the `testing` package and migrates `TripStateTest`'s file-private `ObaTripStatus` stub onto it, removing the duplicate (incl. the duplicated `Unsafe`/`FAKE_LOCATION` trick).

## Verification

All affected unit tests pass: `SingleFlightTest` (3), `TripTrajectoryViewModelTest` (4), `TripStateTest` (27) — 0 skipped, 0 failures.

## Summary by CodeRabbit

* **Tests**
  * Added a JVM `testTripStatus` helper to standardize creation of trip status test data and reduced duplicated per-test implementations.
  * Updated trip state tests to use the shared helper for building trip status instances.
  * Expanded trip trajectory ViewModel tests with a hydrating fake repository that refreshes UI state from a trip-details stream.
  * Strengthened `SingleFlight` tests with clearer edge-case documentation and a new assertion that distinct keys don’t coalesce.

### DIFF
--- a/onebusaway-android/src/test/java/org/onebusaway/android/extrapolation/data/TripStateTest.kt
+++ b/onebusaway-android/src/test/java/org/onebusaway/android/extrapolation/data/TripStateTest.kt
@@ -15,7 +15,6 @@
  */
 package org.onebusaway.android.extrapolation.data
 
-import android.location.Location
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertSame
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
@bmander this has a merge conflict. please rebase!

PR #1601, opened 2026-06-28
Deep-link queue (#1582) follow-ups: lifecycle-aware collect, regression tests, comment polish

### LINKED ISSUE #1592: Deep-link queue (#1582 follow-ups): lifecycle-aware collection + extra regression tests + comment polish
Follow-ups from review of #1587 (queued deep-link routes). None are blocking; the merged implementation is correct.

### Lifecycle awareness
`DeepLinkEffect` (`HomeNavHost.kt:161`) now collects `deepLinkRoutes` via a raw `LaunchedEffect(navController) { collect {} }`. The previous latch used `collectAsStateWithLifecycle()`, which pauses below `STARTED`; the new collect will navigate even while the activity is `STOPPED`. In practice every staging path runs while foregrounded or arrives via `onNewIntent` (which brings the activity forward), so this is unlikely to bite — but if we want to preserve the prior semantics, wrap the collect in `repeatOnLifecycle(STARTED)` / `flowWithLifecycle`. Decide and document the intended behavior.

### Additional regression tests (`HomeViewModelTest`)
- Consecutive-duplicate routes (`a`, `a`) → assert `listOf("a", "a")`. This is the cleanest #1582 repro (a StateFlow latch drops the repeat via distinct-until-changed) and pins that the queue does not dedupe.
- Re-collection after cancellation: collect into job A, `cancel()`, `stageDeepLinkRoute("x")`, collect into job B → assert B receives `["x"]`. Locks in the buffered-across-recreation guarantee and the `receiveAsFlow` (not `consumeAsFlow`) choice.
- Follow-up: a Compose/Robolectric test of `DeepLinkEffect`'s `navigate { popUpTo(HOME); launchSingleTop }` contract (the queue-delivers-duplicates vs. `launchSingleTop`-dedupes interaction).

### Comment polish (`HomeViewModel.kt`, `HomeNavHost.kt`)
- The word "distinct" in the `deepLinkRoutes` field comment and `DeepLinkEffect` KDoc understates the guarantee — the queue preserves all routes in order, duplicates included (the PR's own `a,b,a` test relies on this). Drop "distinct."
- "trySend always succeeds on an UNLIMITED channel" (`stageDeepLinkRoute`) is true only because the channel is never closed; reword to note that.
- `DeepLinkEffect` KDoc omits `launchSingleTop = true`; add a half-sentence so the "exactly once" delivery claim isn't conflated with nav-layer dedupe.

bmander:
Closes #1592. Non-blocking follow-ups from the #1587 review; the merged queue implementation was correct.

### Lifecycle awareness
**Decision: gate the collect on `repeatOnLifecycle(STARTED)`.** This restores the pre-#1587 `collectAsStateWithLifecycle` pause-below-STARTED semantics so `DeepLinkEffect` never navigates while the activity is STOPPED. The UNLIMITED channel buffers any route staged during that window and replays it when the lifecycle returns to STARTED, so the gate costs no deliveries. Documented in the `DeepLinkEffect` KDoc.

### Regression tests (`HomeViewModelTest`)
- **Consecutive-duplicate routes** (`a`, `a`) → `listOf("a", "a")` — the cleanest #1582 repro (a StateFlow latch drops the repeat via distinct-until-changed); pins that the queue does not dedupe.
- **Buffered between collectors** — collect into job A, cancel, `stageDeepLinkRoute("x")`, collect into job B → B receives `["x"]`; locks in the buffered-across-recreation guarantee and the `receiveAsFlow` (not `consumeAsFlow`) choice.

The further Compose/Robolectric test of `DeepLinkEffect`'s `navigate { … launchSingleTop }` contract is left out: there's no Compose/Robolectric harness in `src/test` today, and standing one up is out of scope for this non-blocking polish. Filing it as a separate follow-up if desired.

### Comment polish
- Dropped the misleading "distinct" from the `deepLinkRoutes` field comment and `DeepLinkEffect` KDoc — the queue preserves all routes in order, duplicates included.
- Reworded `stageDeepLinkRoute`'s `trySend` note: it never drops because the channel is UNLIMITED *and* never closed.
- Documented `launchSingleTop = true` in the `DeepLinkEffect` KDoc as the nav-layer dedupe, distinct from the queue's exactly-once delivery.

### Testing
`./gradlew :onebusaway-android:testObaGoogleDebugUnitTest --tests "...HomeViewModelTest"` — green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved in-app navigation reliability by only processing incoming actions when the app is actively started.
  * Prevents navigation from continuing while the app is in the background, helping avoid unexpected screen changes.
  * Navigation events are now better handled when the app resumes, reducing the chance of missed or out-of-order actions.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/HomeNavHost.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/HomeNavHost.kt
@@ -33,7 +33,9 @@ import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.compose.LifecycleEventEffect
+import androidx.lifecycle.compose.LocalLifecycleOwner
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.NavController
 import androidx.navigation.NavHostController
 import androidx.navigation.NavType
@@ -238,6 +240,10 @@ fun HomeNavHost(
  * intents to the Activity, which exists before/around the composition. The channel is an UNLIMITED queue
  * (not a latch) so rapid, distinct back-to-back intents are each delivered exactly once rather than
  * overwriting one another before the collector consumes them (#1582).
+ *
+ * The collect is gated on `repeatOnLifecycle(STARTED)` so we never navigate while the activity is STOPPED.
+ * The UNLIMITED channel buffers any intent submitted during that window and replays it when the lifecycle
+ * returns to STARTED, so the gate costs no deliveries (#1592).
  */
 @Composable
 internal fun LaunchIntentEffect(
@@ -245,10 +251,13 @@ internal fun LaunchIntentEffect(
     launchIntents: Flow<Intent>,
     onSideEffects: (Intent) -> Unit,
 ) {
-    LaunchedEffect(navController) {
-        launchIntents.collect { i ->
-            onSideEffects(i)
-            IntentRouteMapper.routeForIntent(i)?.let { navController.navigateFromHome(it) }
+    val lifecycleOwner = LocalLifecycleOwner.current
+    LaunchedEffect(navController, lifecycleOwner) {
+        lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+            launchIntents.collect { i ->
+                onSideEffects(i)
+                IntentRouteMapper.routeForIntent(i)?.let { navController.navigateFromHome(it) }
+            }
         }
     }
 }


bmander:
**Heads up: the scope of this PR has shrunk substantially after rebasing onto `main`.**

`main` has since refactored past most of what this branch did. The deep-link mechanism moved out of `HomeViewModel` (the `_deepLinkRoutes` channel + `stageDeepLinkRoute`) and into `HomeActivity`'s `LaunchIntentChannel` + `IntentRouteMapper`, and the `#1582` queue-not-latch fix is **already in `main`** via that channel. As a result, the rebase resolved to:

- **Dropped** the re-added `HomeViewModel.deepLinkRoutes` / `stageDeepLinkRoute` — they no longer exist on `main`.
- **Dropped** the five `HomeViewModelTest` regression tests — they exercised that deleted API and wouldn't compile.
- **Kept** the one improvement `main` lacked: lifecycle-gating the launch-intent collect, ported onto `main`'s `LaunchIntentEffect`.

So this PR is now a focused **13+/4-** change: wrap `LaunchIntentEffect`'s `launchIntents.collect` in `repeatOnLifecycle(STARTED)` so navigation never fires while the activity is STOPPED (the UNLIMITED channel buffers and replays on the next STARTED, so no deliveries are dropped). Both the main and unit-test source sets compile.

One gap to flag: there's no longer regression coverage for the lifecycle gate, since the old VM-level tests are gone and `LaunchIntentEffect` is a Composable. A dedicated Compose/instrumented test could cover it if desired.

The commit message has been updated to match the new scope, and the branch force-pushed (`--force-with-lease`).

— Claude, on behalf of @bmander

Issue #1617, opened 2026-07-02
Backup restore restarts the process to work around dual DB ownership


bmander:
> Note: this refers to the in-progress storage-modernization work (retiring the legacy `ObaProvider` ContentProvider into a unified Room `AppDatabase`). Some referenced files may not be on `main` yet.

## Background

After the storage-modernization cutover, the Room `AppDatabase` singleton has **two owners** that can diverge:
- the Hilt object graph (`DatabaseModule.provideAppDatabase`, `@Singleton`), and
- the `DatabaseProvider` object (its own `@Volatile INSTANCE`).

They agree at startup (the Hilt provider sources its instance from `DatabaseProvider.getDatabase()`), but diverge the moment `DatabaseProvider.closeDatabase()` runs and a new instance is built — which happens on a **backup restore** (`Backup.restoreRoomBackup` swaps the DB file, forcing a close/reopen). After that, Hilt still holds the old, now-closed instance, and every `@Singleton` repository (`RegionCache`, `ArrivalsRepository`, `ReminderRepository`, …) has already captured a DAO from it, so those throw on their next query.

## Current workaround (kludge)

`BackupUtils.doRestore` restarts the whole process after a Room-format restore (`restartApp` → `Intent.makeRestartActivityTask` + `Runtime.exit(0)`). It works, but restarting the app to swap a file is heavy-handed and hides the real problem: split ownership of a singleton.

## The right fix

Make `AppDatabase` single-owned so no restart is needed. **Recommended:** fold `DatabaseProvider` into the Hilt graph so Hilt is the sole owner, and restore by **merging rows** into the live instance (no file swap, no close) — this means extending the importer to cover all tables (survey + wide-alerts, not just the 11 legacy tables `LegacyDataImporter` handles today) so both backup formats take the in-place path and Room invalidation notifies observers.

## Acceptance criteria

- A backup restore (both Room-format and legacy-format) completes **without** killing/restarting the process.
- No consumer holds a stale/closed `AppDatabase` or DAO after a restore.
- `restartApp` in `BackupUtils` is removed.
- The rollback-on-incompatible-backup behavior in `Backup.restoreRoomBackup` is preserved.

## References
- `backup/Backup.kt` (`restore`, `restoreRoomBackup`)
- `backup/BackupUtils.java` (`restartApp`)
- `database/DatabaseProvider.kt`, `app/di/DatabaseModule.kt`

PR #1646, opened 2026-07-04
Restore backups in place instead of restarting the process (#1617)

bmander:
The Room AppDatabase had two owners that diverged on a backup restore: the
Hilt object graph and the DatabaseProvider object with its own @Volatile
INSTANCE. A Room-format restore swapped the DB file, forcing
DatabaseProvider.closeDatabase()/reopen and a brand-new AppDatabase instance,
leaving Hilt (and every @Singleton repository that captured a DAO from it)
pointing at the old, closed instance. BackupUtils worked around this by
restarting the whole process.

Make the AppDatabase single-owned and restore in place:

- Delete DatabaseProvider; Hilt's DatabaseModule now builds and solely owns
  the AppDatabase singleton (migrations moved here). Bare-Context callers
  (Backup) reach the same instance via DatabaseEntryPoint.appDatabase().
- Extend the importer to cover every table of a full Room-format backup
  (11 legacy tables + studies/surveys + wide-alerts) and merge the backup's
  rows into the live database in a single transaction (LegacyImportDao gains
  RoomBackupData and a replaceAll overload). No file swap, no close, so no
  consumer is left holding a stale/closed AppDatabase or DAO, and Room
  invalidation notifies observers.
- Both backup formats now take the in-place merge path; Backup.restore no
  longer returns a "restart required" flag and restartApp is removed from
  BackupUtils.
- Rollback-on-incompatible-backup is preserved: the merge is one transaction,
  so an unreadable file or a schema/FK mismatch rolls back with the live
  database untouched.

Adds tests covering the full Room-backup merge (survey + alert tables
replaced) and transactional rollback leaving live data intact.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YCFZESQmGVtkDV5bEUZTLn

## Summary by CodeRabbit

* **New Features**
  * Backups and restores now include additional app data: saved studies, surveys, and wide-alerts.
  * Room-format backup restores now update the database in place (no restart required).
* **Bug Fixes**
  * Restores are more reliable: incompatible or corrupt backups no longer overwrite live data.
  * Full restore now replaces the relevant tables cleanly, avoiding stale rows.
* **Tests**
  * Expanded automated coverage for Room backup/restore behavior, including rollback and full-database replace scenarios.


bmander:
Addressed the review feedback in 0a5e2fc3d:

**Legacy path exception wrapping (actionable):** Fixed. Extracted `mergeBackup { ... }` so *both* restore paths — Room-format and legacy ContentProvider-format — wrap merge failures in `IOException`. Previously only the Room branch did this; the legacy branch let unchecked exceptions escape and crash the app instead of letting `BackupUtils.doRestore()` catch it and show the restore-error toast.

**`runBlocking` on the caller's thread (nitpick):** Left as-is intentionally. Restore is a rare, explicitly user-initiated, one-shot action gated behind a confirmation dialog, and the merge is a single transaction whose failure must roll back before we report success/failure — so blocking is the simpler correct behavior here. Moving it off-thread would add async completion plumbing and a progress UI for a payload that is a handful of small tables. Happy to revisit if backups grow large enough to risk an ANR.

PR #1662, opened 2026-07-04
Trip planner: typed endpoints with cancellable pills + Geocoder fallback

bmander:
## Summary

Reworks the trip planner so location endpoints are typed and can't be edited as raw text when they shouldn't be.

- **Sealed `TripEndpoint`** replaces `PlaceItem`: `FreeText` / `Geocoded` / `AddressBook` / `CurrentLocation` / `MapPoint`. Only free text stays an editable autocomplete field; every resolved endpoint renders as a **non-editable, cancellable pill inside the field** (✕ clears it back to the empty field). Picking another input method (contacts / current location / map) overrides the current pill.
- **Android Geocoder fallback**: `DefaultGeocodeRepository` uses Pelias when a key is configured (`BuildFlavorUtils.isPeliasApiKeyDefined()`), else falls back to the on-device `android.location.Geocoder` so key-free builds still geocode. Degraded fallback (no typeahead / transit categories), region-bbox biased, off the main thread.
- **Stale-result fix**: changing or clearing an endpoint now drops a cached plan result (`replanOrClearResult`) so an old route can't remain on screen when the form is no longer submittable.
- **Semantic UI driving**: `testTagsAsResourceId` enabled app-wide and trip-plan fields tagged; fixed the contacts icon's mislabeled content description ("From:" → "Choose from contacts").

## Testing

- `TripPlanViewModelTest` updated for the new model, plus new coverage for endpoint clearing and stale-result invalidation. Passing (13/13).
- Built and exercised on a Pixel 7 Pro emulator: type → pick suggestion → in-field pill; clear a pill after planning → route sheet correctly disappears; Geocoder fallback returns results with no Pelias key.

## Rebase note

Rebased onto current `main`. Conflicts resolved in three places:
- `HomeActivity.kt` — kept upstream's structure (`activityActions`, `RegionPickerHost`) and layered in this PR's app-wide `testTagsAsResourceId` `Box` wrapper.
- `TripPlanScreen.kt` — kept upstream's `if (hasResults) TripResultsMap else …form…` body and added the new `onClearFrom`/`onClearTo` callbacks to the form call.
- `GeocodeRepository.kt` — retargeted the old `io.elements.ObaRegion` reference to the modernized `region.Region` type (`RegionUtils.getRegionSpan` / `LocationUtils.processPeliasGeocoding` already take it).

## Summary by CodeRabbit

* **New Features**
  * Trip planning now uses improved endpoint selection with resolved places shown as removable “pills,” plus support for choosing from contacts, current location, and map points.
  * Added “Clear” and “Choose from contacts” actions/labels for trip endpoints.
* **Bug Fixes**
  * Improved trip planning stability so changing, clearing, or editing endpoints reliably prevents stale routes from staying visible.
* **Style**
  * Enhanced UI automation support by improving Compose test tag exposure across the trip-planning flow.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/HomeActivity.kt
@@ -24,6 +24,12 @@ import android.os.Bundle
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.navigation.compose.rememberNavController
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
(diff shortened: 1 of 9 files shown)


bmander:
Addressed the review feedback:

- **Pill label wrapping** (`TripPlanForm`): the in-field chip `Text` is now `maxLines = 1` + `TextOverflow.Ellipsis` so long geocoder/contact names don't crowd the clear icon.
- **Current location without a fix** (`TripPlanScreen.setCurrentLocation`): now returns early after the toast instead of building a coordinate-less `CurrentLocation(null, null)` that couldn't submit and would clear an existing result.
- **Stale result on text edit** (`TripPlanViewModel`): `onFromQueryChange`/`onToQueryChange` now run the same result-clearing path as the other mutators, so typing over a resolved endpoint drops a now-invalid route.
- **In-flight plan race** (`TripPlanViewModel`): reworked plan submission onto a `mapLatest` pipeline over a conflated channel — a new input cancels the plan in flight, so a slow plan that outlives its edit can't publish a stale `Success`. (Replaced an initial request-token approach with this flow, which is the idiomatic latest-wins mechanism and mirrors the existing suggestion pipelines.)
- **Restore drops transit flag** (`TripPlanScreen.toGeocoded`): the restore mapper now carries `isTransit = isTransitCategory`, matching the geocode mapper.
- **bbox readability** (`GeocodeRepository`): destructured the span array into `centerLat`/`latSpan`/… so the corner math reads as center ± half-span (the arithmetic was already correct).

Added unit tests for the text-edit clear and the in-flight-plan race.

**Not changed — `canSubmit` gating contacts** (`TripPlanFormState`, flagged as the `AddressBook(null, null)` "can't submit" issue): this is a deliberate, tested invariant — `TripPlanFormState.canSubmit` requires resolved coordinates (mirroring `TripRequestBuilder.ready()`), and the test `an endpoint without coordinates does not enable submit` locks it in. Relaxing it to submit raw contact strings for server-side geocoding is a real design change I didn't want to make unilaterally against an explicit test. Happy to take it as a follow-up if maintainers want that behavior.

Issue #1659, opened 2026-07-04
Decompose Application.java: redistribute responsibilities to their modules, then convert to idiomatic Kotlin


bmander:
## Background

`app/Application.java` is the last big pre-modernization god-object. The DI campaign (#1624 → #1626/#1627/#1629, then #1630/#1631) moved app-global **state** out into injected reactive repositories, so `Application` no longer *owns* region/analytics/config/location state. But the file is still a ~490-line Java grab-bag: it holds two service objects on static fields, runs ~8 unrelated startup chores inline in `onCreate`, and exposes a pile of `static` helpers that are really util/module functions wearing an `Application` costume. It's also still Java while the surrounding modernization is Kotlin.

This issue tracks **redistributing those responsibilities to their correct modules** and then **converting the thin residue to idiomatic Kotlin** — the structural finish of the god-object retirement, distinct from the locator-reach sweeps in #1624/#1631.

Note: the dead OTP-URL accessors are already being removed in #1657; this issue is the broader decomposition.

## Responsibility inventory

| # | Responsibility (current form) | Callers | Target home |
|---|---|---|---|
| A | `isIgnoringBatteryOptimizations()` static | **0 anywhere** | **delete (dead)** |
| A | OTP-URL accessors (×4) | 0 | already deleted in #1657 |
| B | `@HiltAndroidApp` class + thin `onCreate`/`onTerminate` | — | **keep** on Application |
| C | `DonationsManager` held on `mDonationsManager` + `getDonationsManager()` static | 14 sites | injectable `@Singleton` (+ `DonationsEntryPoint`, mirroring `AnalyticsEntryPoint`) |
| C | `GtfsAlerts` held on `mGtfsAlerts` + `getGtfsAlerts()` static | 8 sites | injectable `@Singleton` (+ EntryPoint) |
| D | `createNotificationChannels()` + `CHANNEL_*` constants | — | `notifications/NotificationChannels` object/initializer |
| D | `initFirebaseMessaging()` + `getUserPushID()` static | 4 sites | `push/FirebaseMessagingManager` `@Singleton` |
| D | `incrementAppLaunchCount()` (write) | onCreate | fold write into `PreferencesRepository` (it already owns the read, #1642) |
| D | `reportAnalytics()` — initial region label | onCreate | analytics startup task next to `AnalyticsProvider` |
[...]

PR #1663, opened 2026-07-04
Make DonationsManager an injectable @Singleton, drop the Application holder

bmander:
Slice 3 of #1659 (Application.java decomposition) — the largest fan-out.

`DonationsManager` was held on `Application.mDonationsManager`, reached via the static `getDonationsManager()`, and re-exported through a delegating `@Provides` in `AppModule`. This makes it a real Hilt `@Singleton` and routes every consumer to injection.

### Changes
- **`DonationsManager`** → `@Singleton` + `@Inject constructor(@ApplicationContext Context, PreferencesRepository)`. The constructor now only *captures* its collaborators — no dereferencing — with `Resources` and the app-launch count read lazily at their use sites. This keeps the `@Singleton` off the region graph at construction *and* preserves the property that a bare instance can back `DonationViewModelTest` (whose exercised paths never touch the manager).
- **`AppModule`** drops `provideDonationsManager()` (Hilt builds it directly from the `@Inject` constructor) and its now-unused `Application`/`DonationsManager` imports; class doc refreshed.
- **ViewModels inject it**: `DonationViewModel`/`SurveyViewModel` already did (through the old `@Provides`, unchanged); `AdvancedSettingsViewModel` now constructor-injects it instead of the static.
- **Compose nav destinations without a ViewModel** (`SettingsDestinations`, `ExtraDestinations`) reach it through a new **`DonationsEntryPoint`** (mirrors `AnalyticsEntryPoint`). `DonationCard` — which *does* have a `DonationViewModel` — routes through it (new `buildDonationsPageIntent()` passthrough) rather than a locator.
- **`Application`** loses the field, the `onCreate` construction, `getDonationsManager()`, and the import.

No behavior change.

### Verification
- `:onebusaway-android:compileObaGoogleDebugSources` + `compileObaGoogleDebugUnitTestSources` pass (Java + Kotlin + Hilt/KSP).
- `DonationViewModelTest` passes against the new constructor.

## Summary by CodeRabbit

* **New Features**
  * Added a more reliable way to open donation-related pages from multiple parts of the app.

* **Bug Fixes**
  * Improved donation reminder and dismiss state handling in settings.
  * Updated donation screen behavior so actions continue to work consistently across home and settings flows.

* **Refactor**
  * Modernized how donation-related app services are accessed, reducing reliance on shared app-level state.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -24,7 +24,6 @@ import android.util.Log;
 
 import org.onebusaway.android.BuildConfig;
 import org.onebusaway.android.R;
-import org.onebusaway.android.donations.DonationsManager;
 import org.onebusaway.android.notifications.NotificationChannels;
 import org.onebusaway.android.app.di.AnalyticsEntryPoint;
 import org.onebusaway.android.api.ObaApi;
@@ -56,8 +55,6 @@ public class Application extends android.app.Application {
 
     // Region preference (long id)
     private static final String TAG = "Application";
-
-    private DonationsManager mDonationsManager;
 
     private static Application mApp;
 
@@ -97,8 +94,6 @@ public class Application extends android.app.Application {
         incrementAppLaunchCount();
 
         FirebaseMessagingEntryPoint.get(this).fetchAndStoreToken();
-
-        mDonationsManager = new DonationsManager(getApplicationContext(), getResources(), getAppLaunchCount());
     }
 
     /**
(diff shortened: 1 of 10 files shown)


bmander:
Addressed the CodeRabbit comment: updated the `mContext` doc in `DonationsManager` — it was stale after this PR made `mContext` the source for `Resources` (via lazy `getResources()` in `reportUIEvent`/`buildOpenDonationsPageIntent`). The comment now states both uses — (a) reading `Resources` for event labels / the donate URL, and (b) lazily resolving the `AnalyticsProvider` — while keeping the "why lazy" region-init-ordering rationale. Comment-only change.

PR #1665, opened 2026-07-04
Extract notification channels out of Application into NotificationChannels

bmander:
Slice 4 of #1659 (Application.java decomposition).

`Application` owned the three notification-channel id constants, the `createNotificationChannels()` registration, and the `NotificationChannel`/`NotificationManager` imports. This moves all of it into a new `notifications/NotificationChannels` object.

### Changes
- New **`NotificationChannels`** (Kotlin `object`, `@JvmStatic` + `const val` so the Java notification builders keep working): the three ids + `registerAll(context)`.
- IDs move to `NotificationChannels.{TRIP_PLAN_UPDATES,ARRIVAL_REMINDERS,DESTINATION_ALERT}_ID`, dropping the now-redundant `CHANNEL_` prefix. **The persisted id strings (`trip_plan_updates`, etc.) are unchanged**, so existing channels are unaffected.
- `onCreate` calls `NotificationChannels.registerAll(this)` instead of the inlined method.
- The 6 consumers repoint from `Application.CHANNEL_*` to `NotificationChannels.*`:
  - `nav/NavigationService.kt` (×2), `nav/NavigationServiceProvider.java` (×2), `nav/FeedbackReceiver.java`, `directions/realtime/RealtimeService.java`, `tripservice/MyFirebaseMessagingService.kt`, `ui/tripresults/TripResultsScreen.kt`.
- `Application` loses the constants, the method, and the `NotificationChannel`/`NotificationManager`/now-orphaned `Build` imports.

No behavior change. Verified: `:onebusaway-android:compileObaGoogleDebugSources` passes (Java + Kotlin).

## Summary by CodeRabbit

* **New Features**
  * Added centralized notification channel definitions and registration at app startup to ensure channels are consistently created on supported Android versions.

* **Bug Fixes**
  * Updated trip updates, arrival reminders, destination alerts, and feedback-related notifications to use the shared channel IDs instead of app-level constants, improving reliability and consistency across notification types.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -16,18 +16,16 @@
  */
 package org.onebusaway.android.app;
 
-import android.app.NotificationChannel;
-import android.app.NotificationManager;
 import android.content.Context;
 import android.hardware.GeomagneticField;
 import android.location.Location;
-import android.os.Build;
 import android.text.TextUtils;
 import android.util.Log;
 
 import org.onebusaway.android.BuildConfig;
 import org.onebusaway.android.R;
 import org.onebusaway.android.donations.DonationsManager;
+import org.onebusaway.android.notifications.NotificationChannels;
 import org.onebusaway.android.app.di.AnalyticsEntryPoint;
 import org.onebusaway.android.api.ObaApi;
 import org.onebusaway.android.region.Region;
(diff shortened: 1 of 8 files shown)


bmander:
Addressed the CodeRabbit review comment: `registerAll` now guards a null `NotificationManager` (`getSystemService` has a nullable platform return). Rather than a bare `?: return`, it logs a warning before skipping — per the repo's no-silent-failures norm — since there's nothing to register or recover if the service is genuinely absent (a restricted/instrumented-context edge; never null on a real API 26+ device). Compiles clean.

Issue #1642, opened 2026-07-04
Make SurveyUtils' appLaunchCount / DonationsManager reaches injectable (last Application.get() state reaches)


bmander:
Follow-up to #1636 / #1641.

After #1641 threads a `Context` (or injection) into the last context/resource `Application.get()` reaches, the only remaining literal `Application.get()` reaches outside the deliberately-kept `PreferenceUtils.repo()` graph handle are two **Application-lifecycle state** reaches inside a single method:

**`ui/survey/utils/SurveyUtils.kt` — `shouldShowSurveyView(...)`:**
- `Application.get().appLaunchCount` — the every-Nth-launch survey gate.
- `Application.getDonationsManager().shouldShowDonationUI()` — suppress the survey when the donation UI is showing.

These are out of #1636's scope (which was context/resource reaches). They're left as a mixed state: `shouldShowSurveyView` already receives an injected `SurveyPreferences` as a parameter (via #1641), but still statically reaches `Application` for these two.

## Why they weren't done in #1641

- **`appLaunchCount`** is a preference-backed counter (`APP_LAUNCH_COUNT_KEY = "appLaunchCountPreferencesKey"`), read via `Application.getAppLaunchCount()`. It has no injectable source yet — it would need to be exposed through `PreferencesRepository` (or a small `AppLaunchCounter`).
- **`DonationsManager`** is a hand-constructed singleton (`Application.java`: `mDonationsManager = new DonationsManager(...)`, exposed via the static `Application.getDonationsManager()`). It is not in the Hilt graph, so injecting it requires a provider/binding first.

## Approach

1. Expose `appLaunchCount` as an injectable read (e.g. a value on `PreferencesRepository`, or a dedicated `@Inject` counter).
2. Provide `DonationsManager` through Hilt (`@Provides` in an `AppModule`, or an `@Inject` constructor) so it can be injected rather than reached via `Application.getDonationsManager()`.
3. Inject both into whatever owns the survey gate (currently `SurveyViewModel`, which already injects `SurveyPreferences`), and pass them into `SurveyUtils.shouldShowSurveyView` the same way `SurveyPreferences` is passed — keeping `SurveyUtils` a pure, JVM-testable object.

## Acceptance

`shouldShowSurveyView` no longer references `Application.get()`; the only remaining `Application.get()` in the app is the documented `PreferenceUtils.repo()` DI graph handle.

Low priority / diminishing returns, same as #1636.

PR #1668, opened 2026-07-04
Extract the onCreate startup chores out of Application

bmander:
Slice 6 of #1659 (Application.java decomposition) — the diffuse "startup chores" cleanup. Moves the small, unrelated first-run/startup chores off the `Application` god-object to their proper owners, so `onCreate`/`initOba` shrink to a short list of named steps.

### App-launch counter → `PreferencesRepository`
The repo already owned the *read* (`getAppLaunchCount`, #1642); it now also owns the **write** (`incrementAppLaunchCount()`) and the **key** (`APP_LAUNCH_COUNT_KEY` moves to its companion object — the read impl and both test fakes no longer reach back to `Application` for it). `onCreate` calls both through `PreferencesEntryPoint`. `Application` drops the constant and both methods.

### Per-install `app_uid` — seeding stays eager at startup (by design)
`app_uid` has **two independent direct readers** — `ObaEndpointResolver` (sends it as `app_uid`) and `DefaultOpen311Repository.submit()` (reads it as `device_id`) — so it can't be seeded lazily in one reader without risking a null read from the other (per CodeRabbit's review). It's a genuine cross-cutting install-identity invariant, so its unconditional one-time seed stays in `onCreate` (just re-homed out of the vague `initOba()`); the readers stay pure reads. Only `initOba()` as a wrapper is gone.

### First-run defaults → their util owners
- Arrival-info-style default → `BuildFlavorUtils.applyDefaultArrivalInfoStyleIfUnset()` (it already owns the flavor→pref mapping; the 3-case switch collapses to "only STYLE_A → A, else B", preserving behavior).
- Applying the saved theme → `ThemeUtils.applyPersistedTheme()` (it already owns `setAppTheme`).

`onCreate` calls these two directly; `initOba()` / `checkArrivalStylePreferenceDefault()` / `checkDarkMode()` are gone.

### No behavior change
`PreferenceUtils` already delegates to `PreferencesRepository`, so every read/write hits the same DataStore before and after; `app_uid` is still generated once at startup and persisted (invariant per install), and the API interceptor / Open311 `device_id` still read it exactly as before.

### Verification
main + unit + androidTest sources all compile (Java + Kotlin + Hilt/KSP); the `*Preferences*` / `*Survey*` unit tests pass.

## Summary by CodeRabbit

* **New Features**
  * App startup now restores your saved theme automatically.
  * First-time setup now applies a sensible default arrival-info style when none is saved.

* **Bug Fixes**
  * App launch counts are now tracked more consistently across the app.
  * Per-install app ID setup is handled earlier at startup, helping other features read it reliably.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/net/ObaEndpointResolver.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/net/ObaEndpointResolver.kt
@@ -60,7 +60,8 @@ class ObaEndpointResolver @Inject constructor(
     /** The app version code (`app_ver`) — a build constant, so no per-request lookup. */
     val appVersion: Int get() = BuildConfig.VERSION_CODE
 
-    /** The persisted per-install app UID (`app_uid`), generated once at app startup. Invariant per
-     * process, so read once at construction rather than on every request. */
+    /** The persisted per-install app UID (`app_uid`), seeded eagerly in `Application.onCreate` (it has
+     * other direct readers too, e.g. the Open311 report path). Invariant per process, so read once at
+     * construction rather than on every request. */
     val appUid: String? = preferences.getString(ObaApi.APP_UID, null)
 }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/Application.java
@@ -49,6 +49,7 @@ import dagger.hilt.android.EntryPointAccessors;
 import dagger.hilt.android.HiltAndroidApp;
 import org.onebusaway.android.app.di.DatabaseEntryPoint;
 import org.onebusaway.android.app.di.FirebaseMessagingEntryPoint;
+import org.onebusaway.android.app.di.PreferencesEntryPoint;
 
 @HiltAndroidApp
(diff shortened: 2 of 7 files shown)


bmander:
Addressed the CodeRabbit review.

**`app_uid` seeding (the Major finding — valid, thanks).** The finding is correct: `app_uid` has **two** independent direct readers — `ObaEndpointResolver` (sends it as `app_uid`) *and* `DefaultOpen311Repository.submit()` (reads it as `device_id`, `Open311Repository.kt:140`). My original slice moved seeding into the resolver, which is only constructed lazily on first OBA-API use, so an Open311 submit before that could have sent a null `device_id` — a weakened invariant vs. the old unconditional `onCreate` seed. Fixed by **restoring eager seeding at startup** (it's a genuine cross-cutting install-identity invariant with no single reader-owner) and reverting the resolver back to a pure read + the interceptor's `?.let`. So this slice now extracts the launch-count and first-run-defaults chores; `app_uid` seeding stays in `onCreate` by design.

**`= Unit` no-op nit.** Moot — the `incrementAppLaunchCount` override it referred to was already removed in the prior cleanup pass (it's now an interface default method, so the test fake inherits it and has no override).

Verified: main + unit + androidTest compile.

Issue #1625, opened 2026-07-03
Modernize the OpenTripPlanner directions subsystem (last AsyncTask / Jackson / Java island)


bmander:
## Background

After the API modernization (OBA REST → Retrofit + kotlinx.serialization in `api/`), the **OpenTripPlanner trip-planner** subsystem under `directions/` is the one coherent legacy island left untouched. It's the mirror image of the migration that already happened for the OBA endpoints — it just never got done for OTP.

It's worth calling out as one target because it *bundles* several legacy patterns that otherwise no longer exist in the app:

- **The app's last real `AsyncTask`** — `directions/tasks/TripRequest.java` (every other `AsyncTask`/`AsyncTaskLoader` reference in the tree is now just a "replaces the old…" comment).
- **The only remaining Jackson consumer** — the entire `com.fasterxml.jackson.*` dependency exists solely for `directions/util/JacksonConfig.java` + the OTP POJOs. Modernizing this is the only thing blocking dropping Jackson.
- **Manual `HttpURLConnection`** networking (`TripRequest.java`, and `ui/tripplan/TripPlanRepository.kt`), vs. the Retrofit/OkHttp stack used everywhere else.
- **All Java, no Kotlin** — 11 Java files, 0 Kotlin.
- Several `java.util.Date`/`Calendar` sites (pre-`java.time`).

## Files (`directions/`, all Java)

```
directions/model/Direction.java              directions/util/ConversionUtils.java
directions/model/ItineraryDescription.java   directions/util/CustomAddress.java
directions/realtime/RealtimeService.java      directions/util/DirectionsGenerator.java
directions/realtime/RealtimeWakefulReceiver.java  directions/util/JacksonConfig.java
directions/tasks/TripRequest.java             directions/util/OTPConstants.java
directions/util/TripRequestBuilder.java
```

## Proposed work

1. Replace `TripRequest`'s `AsyncTask` + `HttpURLConnection` with a Retrofit `OtpWebService` + a coroutine-based repository (mirroring `api/`).
2. Move OTP JSON off Jackson to kotlinx.serialization (or Retrofit's converter), then **remove the Jackson dependency** and its build.gradle entry.
3. Port the `directions/` files to Kotlin; migrate `Date`/`Calendar` to `java.time`.

**No user-facing behavior change** intended — trip planning should work identically. Related: this removes the last consumer of the Jackson dependency called out in the recent modernization-residue cleanup.

bmander:
[...]

PR #1677, opened 2026-07-05
Modernize OTP directions subsystem to Kotlin + java.time (#1625)

bmander:
## What / why

Finishes the remaining work on #1625 — modernizing the OpenTripPlanner trip-planner subsystem under `directions/`, the last coherent legacy island. The AsyncTask, Jackson, and Retrofit goals from that issue were already done on `main`; this PR completes the **Kotlin port** and the **`java.util.Date`/`Calendar` → `java.time` migration**.

## Changes

- **All 9 `directions/` files ported Java → Kotlin** (the package is now 0 Java files): `OTPConstants`, `CustomAddress`, `Direction`, `ItineraryDescription`, `ConversionUtils`, `DirectionsGenerator`, `TripRequestBuilder`, `RealtimeService`, `RealtimeWakefulReceiver`. ~2,650 lines of Java removed.
- **`java.time` migration:**
  - `ConversionUtils.parseOtpDate` → `Instant?`
  - `ItineraryDescription.endDate` → `Instant`
  - `TripRequestBuilder` date API → `Instant` (drops the `Date`/`Calendar` setters; formats with `DateTimeFormatter` instead of `SimpleDateFormat`)
  - `DirectionsGenerator` transit times → plain epoch-`Long` math
  - `RealtimeService` → `Instant` / `System.currentTimeMillis()`
  - This lets us delete the `Calendar` bridge in `TripPlanRepository.builderFor`.
- **Callers updated:** `TripPlanRepository` (passes `Instant`), `TripPlanScreen` (`dateTime.toEpochMilli()`), `RealtimeServiceTest` (`setDateTime(Instant…)`).

No user-facing behavior change intended.

## Deliberate holdout (reviewer attention)

`ConversionUtils.getTimeWithContext` / `getTimeUpdated` (and their `isToday`/`isTomorrow` helpers) **still use `java.util.Calendar`**. They're GMT-anchored, locale-aware formatters that pair `android.text.format.DateFormat` (which needs `Date`/`Calendar`) with wall-clock rounding and same-day comparisons. They're ported to Kotlin but kept on `Calendar` — rewriting them on `java.time` risks a subtle formatting regression and wants on-device verification. Flagged in a code comment at the call site.

One other behavior-sensitive spot worth a look: `TripRequestBuilder.getFormattedDate` moved from `SimpleDateFormat` to `DateTimeFormatter` for the OTP request date/time strings (`MM-dd-yyyy`, `hh:mma`, US locale, device zone). Verified pattern-equivalent, but calling it out since it feeds the wire request.

## Interop preserved

The remaining Java callers (the `androidTest` and the map-flavor helpers) still compile: `CustomAddress` constructors + `CREATOR`; `RealtimeService` stays `open` with JVM-public `rescheduleRealtimeUpdates`/`getItinerary`/`getSimplifiedBundle`/`onHandleIntent` because `RealtimeServiceTest` subclasses it and calls them.

## Testing

`compileObaGoogleDebugJavaWithJavac` and `compileObaGoogleDebugAndroidTestJavaWithJavac` both build clean. Not yet exercised on-device.

## Summary by CodeRabbit

* **New Features**
  * Enhanced live trip monitoring to more reliably re-check the planned itinerary, alert you to schedule deviations, and notify you when a new plan is available.
  * Improved transit directions (get-on/get-off and intermediate stop details) with more consistent time/distance formatting.

* **Bug Fixes**
  * Improved date/time handling across trip planning and intent restore by using consistent Instant-based timestamps.
  * Increased reliability of realtime check scheduling and notification readiness when required data is unavailable or invalid.


bmander:
Thanks for the review. Addressed 8 of the 9 comments (pushed as an amend to keep the single squashed commit; builds clean for `obaGoogleDebug` main + androidTest):

- **`RealtimeService` `e.printStackTrace()` → `Log.e(TAG, …, e)`** (nitpick) — done.
- **`ItineraryDescription.getDelay` `!!` NPE** — now returns `Long?` (null when either end date is unparseable); the caller treats null as "matched, no computable deviation" and proceeds to the timeout check.
- **`ItineraryDescription.isExpired` device clock** — changed to `fun isExpired(now: Instant)` so the helper no longer reads the clock (per the server-vs-device time guideline); `checkDisableDueToTimeout` passes the "now". There's no server-synced clock available in that background scheduling path, so the device clock stays at the call site rather than inside the model.
- **Unguarded `ItineraryDescription(...)` in the re-plan loop** — wrapped in the same NPE/IndexOutOfBounds guard used by `getSimplifiedBundle` (skip + log the malformed itinerary).
- **`Class.forName(null)` NPE** — guard `NOTIFICATION_TARGET` and return null so the caller takes its no-target path.
- **`RealtimeWakefulReceiver` `intent.extras!!`** — null-safe `intent.extras?.let { … }`.
- **`DirectionsGenerator.itineraryTitle` crash for a lone bicycle/car leg** — instead of force-unwrapping, fall through to the general labeling loop (which already tags a bicycle leg with the bikeshare label); WALK single-legs are unchanged.
- **`TripRequestBuilder.buildRequest` opaque NPE when `dateTime` is unset** — throw a clear `IllegalArgumentException` mirroring the from/to check, and dropped the now-unneeded `!!` in `getFormattedDate`.

**Deferred (1):** the `isToday`/`isTomorrow` device-clock note in `ConversionUtils`. These back the `getTimeWithContext`/`getTimeUpdated` `Calendar` formatters that the PR already flags as a deliberate holdout pending on-device verification — threading `now` through them is part of that eventual java.time rewrite, so I've left them with the documented deferral rather than partially rework the cluster now. (This matches your own note that it's mainly a flag for when that bridge is removed.)

PR #1696, opened 2026-07-05
Fix garbage ETA for closed-stop arrivals with suppressed predictions (#1687)

### LINKED ISSUE #1687: Closed stop shows garbage -29,718,596 min ETAs and arrivals despite closure
## Summary

At a stop with a "stop closed" service alert, the arrivals drawer shows two problems:

1. **Bug 1 (correctness):** Arrivals display an absurd ETA of ~`-29,718,596 min` and/or a "Scheduled" label, instead of a real time.
2. **Bug 2 (UX):** "Load more arrivals" surfaces a normal-looking arrival (e.g. "114 min") into the peek region with no indication the stop is closed.

Reproduced on **stop `1_82673` — "100th Ave NE & NE 140th St" (NB)**, King County Metro, which has situation `1_85852`: *"Stop #82673 … is closed through Fri Aug 28."*

## What the server actually returns

Live `arrivals-and-departures-for-stop/1_82673` (route 230, both entries carry `situationIds:["1_85852"]`):

| field | Entry 1 | Entry 2 |
|---|---|---|
| `predicted` | **true** | true |
| `predictedArrivalTime` | **-1** | 1783122710000 |
| `scheduledArrivalTime` | 1783119110000 | 1783122710000 |

`currentTime` ≈ 1783116081756. So for a closed stop the server keeps `predicted:true` but **suppresses the near-term prediction to a non-positive sentinel** (`-1`; a value of `0` is the other observed sentinel and produces the exact "Scheduled" + `-29,718,596` combination reported). The farther-out trip keeps a valid prediction.

## Root cause — Bug 1

`ArrivalInfo.kt:149-157` selects the predicted time purely off the `data.predicted` **boolean**, never checking the value is positive:

```kotlin
if (data.predicted) {              // true
    eta = predictedMins - nowMins  // predictedMins = -1/60000 = 0  ->  eta = -nowMins ~ -29,718,601
    displayTime = predictedTime    // -1
}
```

The `-1`/`0` sentinel is treated as a real epoch-millis timestamp. `predictedMins` truncates to `0`, so `eta ~ -(now in minutes)` ~ **-29,718,596**. The status label (`computeStatusLabel`, line 230) branches on `predictedTime != 0L`, so a `0` sentinel yields the **"Scheduled"** label while `-1` yields nonsense "N min early" text. This same value is rendered directly in both the list (`ArrivalRows.kt`) and the peek row (`ArrivalsPanel.kt:262`).

The wire->model adapter (`ArrivalAdapters.kt:38-40`) passes `predicted`/`predictedArrivalTime` through unnormalized, and the legacy Java `ArrivalInfo` had the identical logic — so **this is long-standing, not a modernization regression**.

**Fix direction:** guard on a positive predicted time (treat `predictedArrivalTime <= 0` as "no prediction" -> fall back to the scheduled time / "Scheduled" label), ideally normalized at the wire->domain boundary per CLAUDE.md, so both `0` and `-1` sentinels are handled in one place.

## Root cause — Bug 2

Arrivals filtering (`ConvertArrivals.kt`) is only route-filter + `eta >= 0`/`show_negative_arrivals`; **service alerts never affect the arrival rows** — a closure is shown only as a separate banner. So a healthy far-out arrival is presented at a closed stop with no closure cue on the row itself. Needs a product decision: annotate/suppress arrivals for a closed stop, or surface the per-arrival `situationIds`.

## Acceptance
- Closed-stop arrivals with a non-positive predicted time never show a garbage ETA; they render as scheduled or are handled sanely.
- Add a JVM unit test for `ArrivalInfo` covering `predicted=true` with `predictedArrivalTime` of `-1` and `0`.

bmander:
Fixes #1687 (Bug 1).

## Problem

At a stop with a "stop closed" service alert, the server keeps `predicted:true` but suppresses the near-term prediction to a **non-positive sentinel** (observed `-1` and `0`). `ArrivalInfo` selected the ETA purely off the `predicted` boolean, so it subtracted a ~0 epoch timestamp from "now" and rendered a garbage `~ -29,718,596 min` ETA — plus a bogus "N min early" status for `-1`, or a "Scheduled" label paired with a garbage time for `0`.

## Fix

- **Normalize at the wire→domain boundary** (`ArrivalAdapters`): a `predictedOrAbsent` helper collapses any non-positive predicted instant to the single canonical "no prediction" value `0`, so both `-1` and `0` are handled in one place and a `-1` can never be treated as a real epoch-millis timestamp (per CLAUDE.md's "normalize units/shape at the wire boundary" guidance).
- **`ArrivalInfo`** now treats a prediction as usable only when the predicted instant is positive (`data.predicted && predictedTime > 0`). ETA, display time, and the status label all branch on that one flag, so a suppressed arrival cleanly falls back to the scheduled time / "Scheduled" label instead of garbage.
- **Harden `reminderDepartureTime`** against the `-1` sentinel (`> 0`, not `!= 0`) — otherwise it would set a reminder at epoch −1.
- **JVM unit tests** for `ArrivalInfo` covering `predicted=true` with `predictedArrivalTime` of `-1` and `0`, plus a genuine positive prediction and an unpredicted arrival.

## Acceptance criteria

- ✅ Closed-stop arrivals with a non-positive predicted time never show a garbage ETA — they render as scheduled.
- ✅ Added the requested JVM unit test.

## Note: Bug 1 fix widens the Bug 2 surface

Issue #1687 also describes **Bug 2** (a healthy far-out arrival appearing at a closed stop with no closure cue on the row), which it explicitly defers to a product decision. This PR does **not** address Bug 2, and reviewers should be aware it makes the Bug 2 symptom slightly more prevalent:

Previously, a closed-stop entry with a `-1` sentinel had a garbage negative ETA and so was dropped by `convertArrivals`' `eta >= 0` filter (for users who disable "show negative arrivals"). Now that its ETA is a sane positive scheduled value, that row surfaces as a normal-looking "N min"/"Scheduled" arrival — with the closure still shown only as a separate banner. This is the intended "render sanely" outcome for Bug 1, but it means fixing the garbage ETA un-hides these rows. A follow-up (e.g. surfacing per-arrival `situationIds` on the row, or annotating/suppressing arrivals at a closed stop) is the natural home for Bug 2.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved arrival handling so invalid or non-positive predicted times are treated as unavailable, preventing them from showing as real-time.
  * Arrivals, reminder selection, and status/ETA labeling now consistently fall back to scheduled times when predictions are suppressed, and the reported “predicted” flag matches usability.

* **Tests**
  * Added/expanded unit tests for predicted, unpredicted, and sentinel predicted-time cases to verify display time, ETA calculations, and trip report context behavior.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
@@ -26,6 +26,15 @@ import org.onebusaway.android.time.ServerTime
 /** Adapts a modernized [ArrivalDeparture] DTO (arrivals fetch) to the [ArrivalData] model. */
 internal fun ArrivalDeparture.asArrivalData(): ArrivalData = DtoArrivalData(this)
 
+/**
+ * A closed (or otherwise suppressed) stop keeps `predicted:true` but replaces the near-term
+ * prediction with a non-positive sentinel — observed `-1` and `0` for stop `1_82673` in issue #1687.
+ * Collapse any non-positive predicted instant to the single canonical "no prediction" value `0`
+ * here at the wire→domain boundary, so downstream `!= 0` / `> 0` guards all see one sentinel and a
+ * `-1` can never be treated as a real epoch-millis timestamp.
+ */
+private fun predictedOrAbsent(epochMs: Long): Long = if (epochMs > 0L) epochMs else 0L
+
 private class DtoArrivalData(private val d: ArrivalDeparture) : ArrivalData {
     override val routeId get() = d.routeId
     override val tripId get() = d.tripId
@@ -39,9 +48,9 @@ private class DtoArrivalData(private val d: ArrivalDeparture) : ArrivalData {
     override val predicted get() = d.predicted
     // Wire→server mint: these are the server clock, already epoch millis.
     override val scheduledArrivalTime get() = ServerTime(d.scheduledArrivalTime)
-    override val predictedArrivalTime get() = ServerTime(d.predictedArrivalTime)
+    override val predictedArrivalTime get() = ServerTime(predictedOrAbsent(d.predictedArrivalTime))
     override val scheduledDepartureTime get() = ServerTime(d.scheduledDepartureTime)
-    override val predictedDepartureTime get() = ServerTime(d.predictedDepartureTime)
+    override val predictedDepartureTime get() = ServerTime(predictedOrAbsent(d.predictedDepartureTime))
     override val status get() = d.tripStatus?.status?.let { Status.fromString(it) }
     override val frequency
         get() = d.frequency?.let { FrequencyWindow(ServerTime(it.startTime), ServerTime(it.endTime), it.headway) }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/arrivals/ArrivalInfo.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/arrivals/ArrivalInfo.kt
@@ -103,7 +103,7 @@ class ArrivalInfo(
 
     /** The departure time to set a reminder for: the predicted time if known, else the scheduled. */
     val reminderDepartureTime: Long
-        get() = if (data.predictedDepartureTime.epochMs != 0L) data.predictedDepartureTime.epochMs
+        get() = if (data.predictedDepartureTime.epochMs > 0L) data.predictedDepartureTime.epochMs
         else data.scheduledDepartureTime.epochMs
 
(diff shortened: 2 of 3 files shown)


bmander:
Addressed the review feedback: `toTripReportContext()` now passes the normalized `ArrivalInfo.predicted` (raw `data.predicted` **AND** a positive predicted instant) instead of the raw server flag. Without this, a closed stop that keeps `predicted:true` while the instants are normalized to `0` would drive `Open311Repository`'s `if (arrival.predicted)` branch to format `Date(0)` as a 1969 garbage report time. Added two `ArrivalInfoTest` cases covering the suppressed and genuine prediction paths. Squashed into the existing commit.

PR #1716, opened 2026-07-06
Derive predicted arrival/departure from scheduleDeviation (#1688)

### LINKED ISSUE #1688: WSF ferry departures show garbage ~-900 min ETAs from stale predicted time
## Summary

At the Washington State Ferries **"Seattle" terminal stop (`95_7`)**, the arrivals drawer shows a couple of healthy arrivals plus a long list of garbage negative ETAs — e.g. `-906 min`, `-901 min`, `-891 min`. These are the ferry **departures** out of Seattle. The server's `predictedDepartureTime` for them is ~15 hours stale, and the app trusts it.

Same underlying theme as #1687: the app blindly trusts `predicted=true` and uses the absolute predicted timestamp without validating it. (The `predictedTime > 0` guard proposed in #1687 does **not** catch this case — here the bad value is a large positive number.)

## Evidence (live `arrivals-and-departures-for-stop/95_7`)

`currentTime = 1783117313123`. The app selects the departure time for origin-terminal sailings (`stopSequence == 0`) and the arrival time for terminating ones (`stopSequence == 1`):

| # | Trip | seq | app uses | scheduled | predicted (server) | app ETA |
|---|------|-----|----------|-----------|--------------------|---------|
| 1 | Seattle -> Bremerton | 0 | departure | +53 min | **1783062899000 (~15h ago)** | **-907** |
| 2 | Seattle -> Bainbridge | 0 | departure | +28 min | **1783063199000 (~15h ago)** | **-902** |
| 3 | Bainbridge -> Seattle | 1 | arrival | +18 min | 1783118580000 | **+22** (OK) |
| 4 | Bremerton -> Seattle | 1 | arrival | +40 min | 1783119660000 | **+40** (OK) |

The -900-ish values match the reported "-906 / -901 / -891" (they drift with `currentTime`). A wider time window returns more of the day's Seattle departures, each with the same stale prediction, producing the long list.

## Root cause

This is an **OBA-server data defect**, provable from the server's own `scheduleDeviation`. OBA's canonical relationship is `predicted = scheduled + scheduleDeviation`. The server applies it **correctly for the arrival leg** and **incorrectly for the departure leg of the same trip**:

- Entry 3 (arrival): `scheduledArrivalTime 1783117800000 + 780s = 1783118580000` == reported `predictedArrivalTime` — **exact match**.
- Entry 4 (arrival): `scheduledArrivalTime 1783119600000 + 60s = 1783119660000` == reported `predictedArrivalTime` — **exact match**.
- Entry 1 (departure): `scheduledDepartureTime 1783120500000 + 60s = 1783120560000`, but reported `predictedDepartureTime = 1783062899000` — **garbage, pinned near the service-day start (~15h stale)**.
- Entry 2 (departure): `scheduledDepartureTime 1783119000000 + 780s = 1783119780000`, but reported `predictedDepartureTime = 1783063199000` — **garbage**.

Meanwhile the trips are healthily tracked (`tripStatus`: `phase=in_progress`, `status=SCHEDULED`, `lastUpdateTime` ~20s old, small `scheduleDeviation` of 60s / 780s). So the vessel real-time data is fine; only the absolute predicted **departure** timestamp is corrupt.

The client (`ArrivalInfo.kt:149-157`) trusts `data.predicted` and uses the absolute `predictedDepartureTime`:

```kotlin
if (data.predicted) {
    eta = predictedMins - nowMins   // predicted is ~15h in the past -> eta ~ -900
    displayTime = predictedTime
}
```

## Fix direction

**Principled (not a heuristic per CLAUDE.md):** derive the predicted time as `scheduledTime + scheduleDeviation` — the exact formula the server already applies correctly for the arrival leg (proven above) — instead of the unreliable absolute `predictedDepartureTime`. `scheduleDeviation` is carried on `ArrivalData` already (`ArrivalAdapters.kt:49`).

Caveat: this changes ETA computation app-wide, so it needs care not to regress agencies where the absolute predicted time is authoritative and `scheduleDeviation` is absent/zero. The maintainer should decide scope.

This should also be **reported upstream** to the OneBusAway / WSF server — the `predictedDepartureTime` it computes for ferry departures is internally inconsistent with both `scheduleDeviation` and the same trip's `predictedArrivalTime`.

## Acceptance
- WSF Seattle-terminal departures show sane ETAs consistent with `scheduled + scheduleDeviation` (~+54 / +41 min in the example), not ~-900 min.
- JVM unit test for `ArrivalInfo` covering `predicted=true` where the absolute predicted timestamp disagrees with `scheduled + scheduleDeviation`.

Related: #1687 (same "app blindly trusts the predicted timestamp" theme, different sentinel).

bmander:
Fixes #1688.

## Problem

At the Washington State Ferries **"Seattle" terminal** (stop `95_7`), origin-terminal ferry departures showed a long list of garbage negative ETAs — `-906 min`, `-901 min`, etc. The server's absolute `predictedDepartureTime` for these sailings is pinned ~15h stale near the service-day start, and the app trusted it verbatim. (Same theme as #1687, but the `predictedTime > 0` guard doesn't catch this — the bad value is a large *positive* number.)

## Root cause

An OBA-server data defect, provable from the server's own `scheduleDeviation`. OBA's canonical relationship is `predicted = scheduled + scheduleDeviation`. The server applies it **correctly for the arrival leg** and **incorrectly for the departure leg of the same trip**:

- Arrival leg: `scheduledArrivalTime + scheduleDeviation` == reported `predictedArrivalTime` — exact match.
- Departure leg: reported `predictedDepartureTime` is garbage (~15h stale), disagreeing with both `scheduleDeviation` and the same trip's `predictedArrivalTime`.

The vessel real-time data is healthy (`phase=in_progress`, small `scheduleDeviation`, ~20s-old `lastUpdateTime`); only the absolute predicted **departure** timestamp is corrupt.

## Fix

At the wire→domain boundary (`DtoArrivalData`), when a trip is tracked (`predicted` + a `tripStatus`) **and** has a real scheduled anchor, derive the predicted instant as `scheduled + scheduleDeviation` — the exact formula the server already applies correctly for the arrival leg — instead of trusting the absolute `predicted{Arrival,Departure}Time`.

Because the server computes both the absolute predicted times and `tripStatus.scheduleDeviation` from the same deviation, this is a **no-op for healthy agencies** and only repairs the corrupt WSF value. It's the canonical OBA relationship, **not a heuristic** (no magnitude thresholds or magnitude-based unit guessing, per `CLAUDE.md`).

Applied uniformly to arrival and departure, so downstream consumers that also read the corrupt absolute value — the reminder alarm time (`reminderDepartureTime`) and the report context — are corrected too, not just the ETA badge.

The `scheduled > 0` guard falls back to the sentinel-normalized absolute value when no scheduled anchor is present (the wire defaults an absent scheduled time to `0`), so the derivation can't itself synthesize a ~epoch-1970 instant that would pass the downstream `> 0` prediction gate and reproduce the garbage ETA.

## Tests

JVM unit tests (no device needed):
- `ArrivalAdaptersTest` — corrupt departure repaired, healthy arrival no-op, early/negative deviation, no-`tripStatus` and zero-anchor fallbacks, sentinel collapse.
- `ArrivalInfoTest` — end-to-end case routing the exact issue #1688 WSF numbers through the real adapter → sane ~+54 min ETA, never ~−900.

## Acceptance
- ✅ WSF Seattle-terminal departures show sane ETAs consistent with `scheduled + scheduleDeviation`, not ~−900 min.
- ✅ JVM unit test covering `predicted=true` where the absolute predicted timestamp disagrees with `scheduled + scheduleDeviation`.

## Follow-up (out of scope for this repo)
The defect should also be **reported upstream** to the OneBusAway / WSF server — the `predictedDepartureTime` it computes for ferry departures is internally inconsistent with both `scheduleDeviation` and the same trip's `predictedArrivalTime`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved arrival and departure time calculations so predicted times are derived more reliably from scheduled time plus delay information when available.
  * Prevented stale or invalid predicted times from showing as large negative or otherwise incorrect ETAs.
  * Kept unpredicted or missing values safely normalized to zero instead of displaying garbage values.

* **Tests**
  * Added coverage for predicted-time calculation edge cases, including early/late delays, missing prediction data, and stale timestamps.
  * Added an end-to-end check to confirm arrival estimates remain sensible in real-world scenarios.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
@@ -36,6 +36,35 @@ internal fun ArrivalDeparture.asArrivalData(): ArrivalData = DtoArrivalData(this
 private fun predictedOrAbsent(epochMs: Long): Long = if (epochMs > 0L) epochMs else 0L
 
 private class DtoArrivalData(private val d: ArrivalDeparture) : ArrivalData {
+
+    /**
+     * The authoritative predicted instant for a stop, in epoch millis, computed at the wire→domain
+     * boundary from OBA's canonical relationship `predicted = scheduled + scheduleDeviation` rather
+     * than the absolute `predicted{Arrival,Departure}Time` field.
+     *
+     * The server derives BOTH the absolute predicted times and `tripStatus.scheduleDeviation` from
+     * the same deviation, so for healthy data this yields exactly the server's own predicted time (a
+     * no-op). But the two can disagree when the absolute field is corrupt: at the WSF "Seattle"
+     * terminal (stop `95_7`, issue #1688) the `predictedDepartureTime` for origin-terminal sailings
+     * is pinned ~15h stale near the service-day start, while `scheduleDeviation` stays sane (60s /
+     * 780s) and the same trip's `predictedArrivalTime` matches `scheduled + scheduleDeviation`
+     * exactly. Deriving from the deviation repairs that garbage (~ -900 min ETA).
+     *
+     * Only applied when the trip is actually tracked (`predicted` AND a `tripStatus` to read the
+     * deviation from) AND there is a real scheduled anchor to add the deviation to; otherwise there
+     * is no usable real-time source, so fall back to the sentinel-normalized absolute value (issue
+     * #1687). The `scheduledEpochMs > 0` guard matters because the wire defaults an absent scheduled
+     * time to 0: without it, `0 + deviation` would be a ~epoch-1970 instant that passes the downstream
+     * `> 0` prediction gate and reproduces the very garbage ETA this fix removes. [scheduleDeviation]
+     * is seconds (+late/−early); scheduled times are epoch millis.
+     */
+    private fun predictedInstant(scheduledEpochMs: Long, absolutePredictedEpochMs: Long): Long =
+        if (d.predicted && d.tripStatus != null && scheduledEpochMs > 0L) {
+            scheduledEpochMs + d.tripStatus.scheduleDeviation * MS_IN_SECOND
+        } else {
+            predictedOrAbsent(absolutePredictedEpochMs)
+        }
+
     override val routeId get() = d.routeId
     override val tripId get() = d.tripId
     override val stopId get() = d.stopId
@@ -48,9 +77,11 @@ private class DtoArrivalData(private val d: ArrivalDeparture) : ArrivalData {
     override val predicted get() = d.predicted
     // Wire→server mint: these are the server clock, already epoch millis.
     override val scheduledArrivalTime get() = ServerTime(d.scheduledArrivalTime)
-    override val predictedArrivalTime get() = ServerTime(predictedOrAbsent(d.predictedArrivalTime))
+    override val predictedArrivalTime
+        get() = ServerTime(predictedInstant(d.scheduledArrivalTime, d.predictedArrivalTime))
     override val scheduledDepartureTime get() = ServerTime(d.scheduledDepartureTime)
-    override val predictedDepartureTime get() = ServerTime(predictedOrAbsent(d.predictedDepartureTime))
+    override val predictedDepartureTime
+        get() = ServerTime(predictedInstant(d.scheduledDepartureTime, d.predictedDepartureTime))
     override val status get() = d.tripStatus?.status?.let { Status.fromString(it) }
     override val situationIds get() = d.situationIds
     override val frequency
@@ -61,4 +92,8 @@ private class DtoArrivalData(private val d: ArrivalDeparture) : ArrivalData {
     override val scheduleDeviation get() = d.tripStatus?.scheduleDeviation ?: 0L
     override val lastKnownLat get() = d.tripStatus?.lastKnownLocation?.lat
     override val lastKnownLon get() = d.tripStatus?.lastKnownLocation?.lon
+
+    private companion object {
+        private const val MS_IN_SECOND = 1000L
+    }
 }
--- a/onebusaway-android/src/test/java/org/onebusaway/android/api/adapters/ArrivalAdaptersTest.kt
+++ b/onebusaway-android/src/test/java/org/onebusaway/android/api/adapters/ArrivalAdaptersTest.kt
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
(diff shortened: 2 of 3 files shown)


bmander:
Closing without merging. After tracing this end to end (client + OBA server source + live `95_7` captures), the right call is **not** to fix this on the client.

**Root cause — upstream, not us.** The WSF GTFS-realtime feed emits stale `predicted{Arrival,Departure}Time` values at Seattle-terminal origin sailings, and the OBA server passes them through verbatim: `ArrivalAndDepartureServiceImpl.setPredictedTimesFromTimepointPredictionRecords` copies each stop's feed prediction directly, guarded only by `<= 0` — which a stale-but-positive epoch defeats. Live capture at stop `95_7` (currentTime `1783320882740`):

| seq | trip | scheduled | predictedArrival | predictedDeparture | scheduleDeviation |
|---|---|---|---|---|---|
| 0 | Seattle→Bremerton | 1783324200000 | 1783235399000 | 1783235399000 | 240 |
| 0 | Seattle→Bainbridge | 1783323900000 | 1783238099000 | 1783238099000 | 1920 |

Both predicted fields are ~24 h stale, so the displayed departure ETA comes out ≈ −1400 min. `scheduleDeviation` stays sane because it's a single robust scalar at the vehicle's current position.

**Why the proposed fix doesn't fly.** Deriving `predicted = scheduled + scheduleDeviation` is exactly what the server's own fallback (`setPredictedTimesFromScheduleDeviation`) does — but as an *unconditional* override it also replaces healthy per-stop predictions, which legitimately differ from `scheduled + scheduleDeviation` (the deviation is sampled at a different stop than the one being predicted). That's a real, measurable correctness regression on good data: the Tampa/HART fixtures shift by a rounding minute (`UIUtilTest` 5→6).

And there's no clean logical invariant to gate the repair. At the origin terminals where the garbage is actually displayed, `predictedArrival == predictedDeparture` (both stale), so a "departure can't precede arrival" dwell check never fires. The only remaining discriminator is a magnitude threshold on `|predicted − (scheduled + scheduleDeviation)|` — an unsanctioned heuristic.

**Decision.** We will not degrade ETA correctness for every well-behaved agency to paper over one upstream feed defect. **The WSF server is sending garbage; the client will faithfully repeat that garbage rather than harm correctness elsewhere.** The fix belongs upstream — the WSF feed, or an OBA server-side sanity check that reconciles the verbatim predicted fields against `scheduleDeviation` at ingestion.

Issue #1728, opened 2026-07-07
Eliminate remaining Kotlin compiler warnings (post-#1726 batch)


bmander:
## Background

The `test (33)` CI job emits **192 unique** Kotlin compiler warnings (1047 raw, deduped across flavors). PR #1729 cleared **~93** of them in five safe, behavior-preserving slices (`@param:` use-site targets, dead-idiom removal, `MenuAnchorType` rename, AndroidX test-API modernization, and a few trivial context/color deprecations).

This issue tracks the **remaining ~100** warnings. They were deliberately deferred from that PR because each needs its own focused PR and, in most cases, on-device verification or a product decision — they are not blind mechanical sweeps.

To regenerate the current list:
```
gh run view --repo OneBusAway/onebusaway-android --job <latest-main-test-job-id> --log | grep "w: "
```

## Remaining buckets (slice PRs off this issue)

- [x] **MapLibre annotation API** (~50) — `Marker`/`Icon`/`Polyline`/`IconFactory`/`MarkerOptions`/`PolylineOptions`, `addMarker`/`addPolyline`/`removeAnnotation(s)`/`selectMarker`/`setOnMarkerClickListener`/`setInfoWindowAdapter`. All deprecated in favor of the Symbol/Line manager APIs. This is a real refactor of the maplibre-flavor map layer and **must be device-tested**.
- [ ] **`IntentService` / `WakefulBroadcastReceiver`** (~9) — `directions/realtime/RealtimeService` + `RealtimeWakefulReceiver` → WorkManager/coroutines. Runtime behavior change; also covers this cluster's `Bundle.getSerializable`/`getParcelable` → `BundleCompat`/`IntentCompat`. Device-test.
- [ ] **Room `Survey.study_id` FK index** (1) — adding `indices = [Index("study_id")]` is a schema change: bump `AppDatabase` `version` 3→4, add `MIGRATION_3_4`, regenerate the exported `schemas/4.json`. Coordinate with the storage-modernization work.
- [ ] **`NetworkInfo` / `isConnectedOrConnecting`** → `NetworkCapabilities` (`util/ObaRequestErrors.kt`) — connectivity-logic change; device-test.
- [ ] **`Drawable.setColorFilter(int, PorterDuff.Mode)`** → BlendMode compat (`ui/common/Shortcuts.kt`).
[...]

PR #1729, opened 2026-07-07
Lint cleanup: clear ~93 compiler warnings (safe batch)

bmander:
Continues the "Lint cleanup" campaign (#1725, #1726) by eliminating a first, **safe** batch of the Kotlin compiler warnings emitted in CI. Baseline was **192 unique** warnings (1047 raw, deduped across flavors); this clears **~93** of them, in five behavior-preserving slices. The remaining warnings are tracked separately in #1728 (they need their own PRs and on-device verification).

Each commit compiles cleanly (`compileObaGoogleDebug{,UnitTest,AndroidTest}Kotlin` + androidTest Java); no runtime behavior changes.

### Slices
1. **Pin annotation use-site target to `@param:`** (38) — `@ApplicationContext`/`@AppScope`/`@Named`/`@StringRes`/`@DrawableRes`/`@ColorRes` on constructor `val`/`var` params now carry an explicit `@param:` target. This pins today's behavior (param-only) — exactly what Hilt constructor injection reads and what lint already checks — so KSP/Hilt still resolve; no binding changes.
2. **Remove dead null-handling / redundant idioms** (8) — K2's stronger flow analysis flags genuinely-dead constructs: a `!!` under `if (clickable)`, a `state?.` where `projection != null` already implies non-null, a `?.` on a non-null `Bitmap` return, a `?:` on a non-null `stopTimes`, a redundant `.toLong()`, two unreachable `else -> null` in exhaustive enum `when`s, and two dead `ObaStop.id == null` checks (`id` is a non-null type). The `DirectionsGenerator` icon-`when` `else -> {log; -1}` is intentionally **kept** as a caller-facing sentinel.
3. **`MenuAnchorType` rename + test `@OptIn`** (11) — `MenuAnchorType` is a deprecated typealias for `ExposedDropdownMenuAnchorType` (import + 8 call sites); `MapRenderStateFramingTest` opts in to `ExperimentalCoroutinesApi` for `UnconfinedTestDispatcher`/`testScheduler`.
4. **Modernize deprecated AndroidX test APIs** (~35) — `androidx.test.runner.AndroidJUnit4` → `androidx.test.ext.junit.runners.AndroidJUnit4` across 26 test classes (adds the `androidx.test.ext:junit:1.2.1` dependency it lives in); deprecated `androidx.test.InstrumentationRegistry` → `androidx.test.platform.app.InstrumentationRegistry.getInstrumentation().(get)targetContext`, mirroring the form the Room DAO tests already use.
5. **Context-aware `WorkManager.getInstance` + `ContextCompat.getColor`** (3) — pass the available context to the deprecated no-arg `WorkManager.getInstance()`; swap `Resources.getColor(int)` for `ContextCompat.getColor`.

### Deferred (tracked in #1728)
MapLibre annotation API migration (~50), `IntentService`/`WakefulBroadcastReceiver` in `directions/realtime` (~9), the `Survey.study_id` Room FK-index (a schema change), and ~10 runtime-risky main-source deprecations (`NetworkInfo`→`NetworkCapabilities`, `setColorFilter`, Settings constants, `TabRow`→`PrimaryTabRow`, `android.R.string.yes/no`, `CustomAddress` `Locale` ctor).

Squash-merge is fine — the five commits are separated by topic for review.

## Summary by CodeRabbit

* **Tests**
  * Updated Android instrumentation tests to use newer AndroidX JUnit runner/context APIs for improved compatibility and reliability.
* **Bug Fixes**
  * Improved survey display behavior when a stop ID is missing.
  * Adjusted schedule replay extrapolation to handle missing stop-time data more consistently.
* **UI/Behavior**
  * Updated dropdown menu anchoring to the latest Compose API across affected screens.
  * Refined time coloring and thumbnail decoding behavior to be more consistent.

### DIFF
--- a/onebusaway-android/build.gradle
+++ b/onebusaway-android/build.gradle
@@ -294,6 +294,7 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.10.1'
     // Unit tests - seems like this is still necessary w/ Android X even though useLibrary is declared earlier
     androidTestImplementation 'androidx.test:runner:1.6.2'
+    androidTestImplementation 'androidx.test.ext:junit:1.2.1'
     androidTestImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.8.1'
     // WorkManager (Java only)
     implementation 'androidx.work:work-runtime:2.9.1'
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/analytics/test/UmamiAnalyticsTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/analytics/test/UmamiAnalyticsTest.java
@@ -4,7 +4,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-import androidx.test.runner.AndroidJUnit4;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 import org.json.JSONObject;
 import org.junit.Test;
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/api/test/LoaderTestCase.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/api/test/LoaderTestCase.java
@@ -27,7 +27,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 
 import androidx.loader.content.Loader;
 import androidx.loader.content.Loader.OnLoadCompleteListener;
-import androidx.test.runner.AndroidJUnit4;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
 
 /**
  * A convenience class for testing {@link Loader}s. This test case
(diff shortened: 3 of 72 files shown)


bmander:
Thanks for the review. Addressed the readability nits and pushed 00d7d10:

- **`LocationUtilsTest` / `RegionUtilTest` / `UIUtilTest`** — cached the repeated `InstrumentationRegistry.getInstrumentation().getTargetContext()` chain into a local (a `before()`-set field in `RegionUtilTest`, since it's shared across tests) and reused it. `androidTest` sources compile clean.

The two "actionable" findings are **false positives** — applying them would be counterproductive, so I left those lines as-is:

- **`TripExtrapolationBuilder.kt:121` (`state.anchor…`)** — `projection = state?.let { extrapolatedVehicleProjection(it, nowMs) }`, so `projection != null` *does* guarantee `state != null`. K2 tracks this bound smart-cast (that's exactly why it flagged the old `state?.` as an unnecessary safe call, which this PR removed). It compiles and is null-safe; the `test (33)` CI job is green. Re-adding `state?.` would re-introduce the warning this PR exists to clear.

- **`PreferenceItems.kt:101` (`onClick = onClick`)** — `val clickable = onClick != null && enabled`; K2 smart-casts `onClick` to non-null inside `if (clickable)` via the boolean-backed val, so no `!!` is needed. Same story: re-adding `!!` would re-introduce a redundant-`!!` warning.

Both are cases where CodeRabbit's analyzer doesn't model K2's flow analysis.

PR #1735, opened 2026-07-07
Only mark an arrival tracked when a vehicle serves its trip (#1681)

### LINKED ISSUE #1681: Non-tracked trips shouldn't have a radiating "tracked" marker
Presently almost every trip has a radiating "tracked" marker on the arrivals row, whether or not there's a vehicle bring tracked. I think this stems from the ambiguity about whether "tracked" means "has avl data" or "the backend is delivering a prediction" which is a genuinely complicated issue (e.g., in principle the backend could be predicting several trips in the future on a given block which is running behind), but in many cases it's very clear that there's no plausible prediction or AVL data for an arrival (it's hours out, etc) and we shouldn't show that there is.

bmander:
Fixes #1681.

## Problem

Nearly every arrival row showed the radiating "tracked" marker, even for trips hours in the future where no vehicle could plausibly be tracked.

The OBA server sets the realtime `predicted` flag **block-wide**: a GTFS block chains many sequential trips onto one vehicle, and once that vehicle reports anywhere on the block, `ArrivalAndDepartureServiceImpl` applies the same `BlockLocation` (`isPredicted() == true`) to *every* stop on *every* trip in the block. So a stop served by a later run in the block inherits `predicted:true` and gets a marker, even though the vehicle is nowhere near it.

Confirmed against a live Puget Sound stop: 20/20 arrivals out to 113 min were all flagged `predicted:true`, but only ~5 had a vehicle actually on the trip. The tell: for the tracked ones, `tripStatus.activeTripId == tripId`; for the rest, `activeTripId` names a different (earlier) trip in the block.

## Fix

Narrow "predicted" to "genuinely tracked" using the server's own `tripStatus.activeTripId` — the trip the vehicle is currently serving.

- New shared predicate `isVehicleServingTrip(tripId, activeTripId)` (in `models/`), **strict**: the vehicle's status counts only for the trip it names as active. An absent/blank `activeTripId` reads as *not serving*, so realtime is shown only when the server explicitly confirms the active trip.
- `ArrivalData.isTracked` is derived at the wire→model adapter as `predicted && isVehicleServingTrip(...)`. The arrivals projection and the report flow read `isTracked`; the raw block-wide flag is no longer exposed to consumers.
- The **trip-details** view already gated its realtime status on the same `activeTripId == tripId` rule inline — it now routes through the shared predicate (no behavior change).
- Suppressed future arrivals fall back to the scheduled time, which equals their predicted time anyway, so **ETAs are unchanged** — they just lose the false marker.
- Also ties the row's deviation **color** to `isTracked` (via `statusColor()`), replacing `computeColor()` — now removed — whose weaker `predicted != 0` test left block-inherited/suppressed rows showing a blue/green deviation color instead of gray.

Not a heuristic: it keys off an explicit server field, not a magnitude/time-threshold guess. (The tempting `predictedTime == scheduledTime` shortcut was rejected because it would wrongly hide a genuinely-tracked *on-time* vehicle.)

On the live data this turns 20/20 marked rows into the 5 genuinely-tracked ones.

## Testing

- `ArrivalTrackingTest` (new): the strict `isVehicleServingTrip` predicate and the adapter's `isTracked` composition (predicted + on-this-trip → tracked; predicted on another block trip / no status / not predicted → not tracked).
- `ArrivalInfoTest`: untracked arrival → scheduled + gray; tracked running-late → prediction + deviation color; plus the existing #1687 sentinel/ETA cases.
- Built and installed on device; far/untracked rows now render gray with no pulse while near/tracked rows keep both marker and color.

## Note on strict semantics

On modern servers, whenever `predicted` is true there is a trip status with `activeTripId` populated, so strict vs. lenient handling of the absent case is indistinguishable in practice (it was on the live Puget Sound data). The strict choice means: on an older/edge server that sends `predicted:true` without an `activeTripId`, realtime is treated as unconfirmed rather than assumed.

## Root cause

The root cause is server-side (`ArrivalsAndDeparturesBeanServiceImpl` copies the block-wide flag onto every arrival bean). This is a client-side defense using data the client already receives; a server-side per-stop gate would be the complementary upstream fix.

## Summary by CodeRabbit

* **New Features**
  * Arrival times now better distinguish real-time tracked predictions from block-wide “predicted” signals, improving ETA accuracy.
  * Status coloring more consistently reflects whether an arrival has a usable real-time prediction for the currently served trip.

* **Bug Fixes**
  * Fixed cases where an arrival could be treated as tracked based only on the server’s predicted flag, even when the vehicle was on a different trip.
  * Preserved expected behavior when the vehicle’s active trip is missing or blank, preventing incorrect tracking/ETA presentation.

* **Tests**
  * Added and updated tests to cover the tracking gate logic and verify ETA/color fallback behavior.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/util/test/UIUtilTest.java
@@ -294,7 +294,10 @@ public class UIUtilTest extends ObaTestCase {
         validateUatcArrivalInfo(arrivalInfo);
 
         // Arrivals/departures that have already happened
-        assertEquals("Arrived on time", arrivalInfo.get(0).getStatusText());
+        // #1681: arrivals whose vehicle is on another trip in the block (activeTripId != tripId) are
+        // "Scheduled", not tracked — indexes 0, 9, 13, 16, 18, 20, 22, 23, 28, 29, 31 here. The rest
+        // (activeTripId == tripId) keep their real-time deviation labels.
+        assertEquals("Scheduled arrival", arrivalInfo.get(0).getStatusText());
         assertEquals("Arrived at " + formatTime(1438804260000L),
                 arrivalInfo.get(0).getTimeText());
         assertEquals("Departed 2 min late", arrivalInfo.get(1).getStatusText());
@@ -316,7 +319,7 @@ public class UIUtilTest extends ObaTestCase {
         assertEquals("On time", arrivalInfo.get(8).getStatusText());
         assertEquals("Departing at " + formatTime(1438804800000L),
                 arrivalInfo.get(8).getTimeText());
-        assertEquals("On time", arrivalInfo.get(9).getStatusText());
+        assertEquals("Scheduled departure", arrivalInfo.get(9).getStatusText());
         assertEquals("Departing at " + formatTime(1438804800000L),
                 arrivalInfo.get(9).getTimeText());
         assertEquals("10 min delay", arrivalInfo.get(10).getStatusText());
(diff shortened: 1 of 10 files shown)


bmander:
Closing — see #1681. The 'tracked' framing turned out to be fraught (block-inheritance + slack-dependent deviation the server doesn't model consistently). Pivoting to a row-level AVL-availability indicator instead.

PR #1744, opened 2026-07-08
Rearchitect trip-plan monitoring onto a foreground service (#1732)

### LINKED ISSUE #1732: Rearchitect trip-plan monitoring off background polling (push / foreground service / coarsen)
## Summary

The "trip plan change" monitoring feature (notify the user if a planned itinerary is delayed/superseded before they travel) is implemented as a **~60-second background network poll** during a 1-hour window before departure. That polling architecture predates modern Android's background-execution limits and fights the platform — the deprecated `IntentService`/`WakefulBroadcastReceiver` (tracked in #1076) are just symptoms.

This issue tracks **rearchitecting the feature off background polling**. It's a product/architecture decision, intentionally kept separate from the mechanical deprecation cleanup (which is being done in its own PR and only de-deprecates the existing design without changing behavior).

## Why the poll itself is the problem

- **WorkManager** (the sanctioned `IntentService` replacement) has a **15-minute minimum** periodic interval and defers work in Doze — it cannot do 60 s cadence. Any faithful modernization therefore has to *keep* `AlarmManager` firing the poll.
- **AlarmManager** exact alarms require `SCHEDULE_EXACT_ALARM`/`USE_EXACT_ALARM` (API 31+/33+, Play-scrutinized); inexact repeating alarms are throttled/coalesced in Doze. So "check every 60 s with the screen off" is already best-effort, not guaranteed.

## Options to evaluate

1. **Server push (FCM / OneSignal)** — the server watches the itinerary's realtime feed and pushes a notification on change; **zero client polling**. Battery-optimal and reliable, and the correct greenfield answer — but needs **backend support** (OTP/OBA server watching planned trips) that doesn't exist today. Requires coordination with the server side.
2. **Foreground service** — a self-contained `while { check(); delay(60s) }` coroutine loop while monitoring (mirrors the app's `NavigationService`). Honest about battery and reliable while active, but adds a **persistent notification** for a passive feature and needs a `foregroundServiceType` on API 34 (no type cleanly fits trip monitoring).
3. **Coarsen to WorkManager 15-min periodic** — drop `AlarmManager` entirely; fully idiomatic and client-only, but degrades the feature (15-min granularity largely kills "your bus is N min late" alerts).

## Related
- #1076 (IntentService deprecated) — being addressed by the faithful modernization PR; this issue is the deeper follow-up.
- #973, #758 — feature *improvements* (smarter/ richer notifications) that assume the feature stays; a rearchitecture should keep these in mind.
- #791, #1486, #790, #749 — historical fragility (empty bundle / NPEs) in this code path.

bmander:
Implements the rearchitecture tracked in #1732: move the trip-plan-change monitor off background polling onto a **foreground service running a coroutine loop**.

#1733 was the faithful `IntentService`/`WakefulBroadcastReceiver` de-deprecation that *kept* the `AlarmManager.setInexactRepeating(60s)` poll (via `RealtimeChecker` + `RealtimeCheckWorker` + `RealtimeReceiver`). This is the deeper follow-up #1732 asks for — the poll itself is the thing being retired.

## What changed
All in `directions/realtime/`:

- **`TripPlanMonitorService`** — Hilt foreground service (`foregroundServiceType=specialUse`) that re-plans the watched request every 60s via a `while { check(); delay(60s) }` loop (modeled on `NavigationService`), notifies on a delayed/early/superseded itinerary, and stops itself.
- **`TripPlanMonitor`** — entry point + scheduler. Within the 1h pre-departure window it starts the service now; further out it schedules a **single one-shot exact alarm** (not a repeating poll) at `departure−1h` that fires **`TripPlanMonitorAlarmReceiver`** → `startForegroundService`. So no persistent notification appears until monitoring actually begins.
- **`TripMonitorDecider` / `TripMonitorWindow`** — the itinerary-diff decision and the start/stop timing boundaries, extracted as pure functions and **JVM-unit-tested** (11 tests).

### Bounded to the trip
The loop stops **once the itinerary departs** — a delay/change warning is moot after travel begins — with the trip-**end** time as a fallback guard when the departure time can't be parsed. This keeps the foreground notification (and the Doze-exempt process) short-lived.

### Also revives a dead feature
The Compose port only handed the monitor `itineraries + index`, never the trip request, so `getSimplifiedBundle` produced a null bundle and **nothing was ever scheduled** (the #790/#791 empty-bundle fragility). This PR captures the `TripPlanParams` that produced the results (`PlanResult.Success.params`) and threads it `TripPlanScreen → TripResultsSheet → TripPlanMonitor.start`; the monitor re-plans it through the shared `TripPlanParams.toRequestBuilder` seam.

### Removed
`RealtimeChecker`, `RealtimeCheckWorker`, `RealtimeReceiver`, `RealtimeWorkData`, `RealtimeCheckerTest`, the now-unused `TripPlanRepositoryEntryPoint`, and two dead `OTPConstants` intent-action constants.

## Design notes
- **Deferral keeps a single one-shot exact alarm** (`SCHEDULE_EXACT_ALARM` already declared): an exact-alarm-fired broadcast is granted the temporary background FGS-start exemption. Falls back to `setAndAllowWhileIdle` if exact-alarm access is revoked. This is the only remaining `AlarmManager` use — the repeating poll is gone.
- **Battery tradeoff (inherent to a foreground service):** during the active window the FGS keeps the process Doze-exempt and shows an ongoing notification. That's the honest cost of a reliable client-only cadence; the battery-optimal answer remains server push (option 1 in #1732), which needs backend work.

## Testing
- ✅ `compileObaGoogleDebugKotlin -PwarningsAsErrors=true`, JVM unit tests (decider + window, 11 tests), androidTest sources compile, full debug APK assembles (manifest merge validates the `specialUse` FGS + `PROPERTY_SPECIAL_USE_FGS_SUBTYPE`), and `lintObaGoogleDebug -PwarningsAsErrors=true` reports **no new issues**.
- ⚠️ **Not yet device-verified** — the FGS / alarm / notification lifecycle should be exercised on-device (plan an imminent trip with realtime legs, notifications enabled).

Closes #1732.

## Summary by CodeRabbit

* **New Features**
  * Added trip-plan monitoring that watches upcoming trips and notifies you if a plan is delayed, early, or changes.
  * Trip results now retain your original trip request to improve notification re-entry and ongoing updates.
  * Updated trip-plan notification messaging (e.g., “Watching your trip”) to reflect what’s monitored.
* **Bug Fixes**
  * Improved reliability of deferred trip monitoring so it starts at the right time and avoids crashes when monitoring data is missing.
* **Tests**
  * Added coverage for trip-monitor decision logic and monitoring window timing boundaries.


bmander:
Thanks for the review. Addressed in df2cd1b:

**Applied (2):**
- **Stale `params` in the selection collector** (`TripResultsScreen`) — real bug, fixed. The itinerary-selection `LaunchedEffect` is keyed on the hoisted `resultsViewModel` (never restarts), so it pinned `params` from the first composition while `currentItineraries` was already updated — after a re-plan, selecting an option armed the monitor with new itineraries but the stale request. Wrapped `params` in `rememberUpdatedState` too.
- **Uncaught exceptions in the monitor loop** (`TripPlanMonitorService`) — fixed. Wrapped each cycle in an inner `try/catch` that logs + retries next tick, rethrowing `CancellationException` so teardown/cancellation still work. (`planBlocking` itself never throws — it maps failures to an empty list — but `initFromBundleSimple`/`decide` can throw on malformed data, and on `Dispatchers.Main` that would crash the app.)

**Declined (2), with rationale:**
- **`withTimeout` around `planBlocking`** — the premise (unbounded hang) doesn't hold: the OTP client already sets `connectTimeout`/`readTimeout` (`NetworkModule.kt`), so a stalled request throws → `runCatching` → empty list → the loop stops gracefully. Adding `withTimeout` would be redundant and, since `TimeoutCancellationException` is a `CancellationException`, would tangle with the cancellation rethrow above.
- **Typed `ServerTime`/`WallTime` for the timing guards** — kept plain epoch millis and documented why. These are coarse foreground-service *lifecycle* booleans ("has this moment passed?"), not user-facing durations, so clock skew is immaterial; and the sibling `ItineraryDescription.isExpired(now: Instant)` in the same subsystem already compares a server-provided time against a plain device `Instant` the same way. The domain-tagged classes are used in the arrivals/`TripState` path, not directions monitoring.

Verified: `compileObaGoogleDebugKotlin -PwarningsAsErrors=true`, JVM unit tests (11), and `lintObaGoogleDebug -PwarningsAsErrors=true` (no new issues) all pass. Still pending on-device verification of the FGS/alarm/notification lifecycle.

PR #1769, opened 2026-07-10
Port stop-marker icon factories to Kotlin and share their drawing

bmander:
## What

Ports the two flavor stop-marker icon factories from Java to Kotlin and removes the ~200 lines of circle+arrow drawing they had copy-pasted between flavors:

- **`StopIconFactory.java` (790 lines, Google)** → Kotlin `object`
- **`MapLibreStopIcons.java` (431 lines, MapLibre)** → Kotlin `object`

Both callers (`GoogleMapRenderer.kt` / `MapLibreRenderer.kt`) are already Kotlin, so the public APIs are unchanged.

## Shared extractions (new in `src/main/.../map/render/`)

- **`StopDirection`** enum — one source of truth for a stop's compass direction: the wire `key`, the cache index (`ordinal`), the Google anchor-offset signs, and the compass angle. Replaces each flavor's parallel `String[]` + `directionToIndexMap` tables, the index-lookup helpers, and (Google) the two 9-branch `getX/YPercentOffsetForDirection` `if/else` chains.
- **`StopBitmaps.directionalStopMarker(...)`** — the per-direction circle + gradient direction-arrow drawing that was byte-for-byte duplicated between the two factories. Google passes a glyph-scaled size and stamps its route-type glyph via an `onCircle` callback; MapLibre uses the bitmap as-is.

## Notable behavior change

MapLibre's arrow-path and cutout math previously used **integer** division (`mPx / 2`, `mPx / 12`) where Google used float. The shared builder is float, so MapLibre's arrow vertices shift by **< 0.5px** — this fixes what appears to be unintended drift (the surrounding bitmap/gradient/rotation math there was already float). Worth a glance on a MapLibre build.

## Incidental modernization

- Deprecated `Resources.getColor(int)` → `ContextCompat.getColor`
- `Bitmap.createBitmap(...)` → androidx `createBitmap` KTX
- Route glyphs rasterized white via `MarkerRendering.rasterize`'s tint (matching `VehicleBitmaps`/`BikeBitmaps`) instead of a draw-time `PorterDuffColorFilter`
- Theme arrow colors fetched once in `loadIcons` and threaded down (was ~180 `getColor` calls at warm-up)

## Net

1221 lines of Java removed; the shared drawing now lives once.

## Verification

- Both flavors compile clean under `-PwarningsAsErrors=true`
- `lintObaGoogleDebug -PwarningsAsErrors=true` (the CI lint gate) passes
- Installed `obaGoogleDebug` on a Pixel 7 Pro and eyeballed stop markers (direction arrows, star favorites, route-type glyphs, dot band, focus scaling) — render correctly

No functional/API change to the map; pure port + de-duplication.

## Summary by CodeRabbit

* **New Features**
  * Updated bus-stop marker artwork with clearer directional arrows, improved gradient contrast, and refined outline styling.
  * Improved direction support with consistent handling and fallback for unknown directions across map views (including focused, favorite, and far-zoom dot variants).

* **Bug Fixes**
  * Corrected marker anchor/alignment for directional full and focused markers to improve visual placement.

* **Refactor**
  * Reworked stop-marker icon generation and caching for consistent behavior between Google Maps and MapLibre.


bmander:
Thanks @coderabbitai — addressed both in the amended commit:

- **`anchorX`/`anchorY` before load:** they now take a `Context` and call `ensureLoaded(...)` themselves (consistent with `stopIcon`), and `GoogleMapRenderer.stopAnchor` passes it. The first directional marker no longer risks anchoring off the default `0.5f` when the anchor is computed before the icon.
- **Anchor offset re-deriving arrow ratios:** the `/3` and `/10` derivation moved into a shared `StopBitmaps.anchorPercentOffset(baseIconPx)`, next to the arrow geometry it must match, so the Google anchors can't silently desync from `directionalStopMarker`.

Both flavors compile clean under `-PwarningsAsErrors=true` and `lintObaGoogleDebug` passes.

Issue #1751, opened 2026-07-08
Rework starred stops/routes onto a minimal ID-set backend (dual-read deprecation)


bmander:
## Summary

Deprecate the current stop/route "favorite" backend and replace it with a deliberately minimal one: **a set of starred stop IDs and a set of starred route IDs**, nothing more. The current system carries a lot of per-favorite structure and a handful of workarounds; the replacement collapses all of that to plain membership and derives everything the UI needs from it.

To de-risk the change, the two implementations live side by side. We **write only to the new store**, and a legacy star is migrated **lazily, the moment it's displayed** — read of an old-style star saves a new-style star, and all display/sort/delete then run off the new store. This severs the old-style-star → display path (it becomes old → new-save → new-read → display), which is what lets deletes always land on the new store. Once the window closes, the converter and legacy tables are retired.

## Target semantics (new store)

The entire user-facing contract becomes:

- **Starred stop IDs** — a stop is shown starred on the map and appears in the *Starred Stops* folder.
- **Starred route IDs** — a route appears in the *Starred Routes* folder, and **any arriving trip whose route is starred is promoted to the top of the stop's arrivals drawer**.

That's it. Starring a route is wholesale (the route, everywhere) — there is no per-headsign, per-stop, or "this stop vs. all stops" dimension anymore.

## What exists today (the thing we're deprecating)

The current model — now living in the Room `database/oba` package (a faithful port of the legacy `ObaProvider`/`ObaContract` SQLite layer, columns preserved) — is **two-tier**:

1. **A nullable `favorite INTEGER` column** on the `stops` and `routes` rows (`1` = starred, `NULL`/`0` = not). Map stars read `SELECT _id FROM stops WHERE favorite = 1` (`StopDao.favoriteStopIds`); the starred folders read `WHERE favorite = 1` (`StopDao`/`RouteDao.starredBy*`).
2. **A richer `route_headsign_favorites` table** keyed on `(route_id, headsign, stop_id, exclude)` that is the *authoritative* route-favorite state; `routes.favorite` is a **derived mirror** kept in sync on every write.

Extra structure and workarounds bundled into this that the new model intentionally sheds:

[...]

PR #1770, opened 2026-07-10
Arrivals: one row per route+direction with a scrollable ETA-pill strip (#1707)

### LINKED ISSUE #1707: Unify arrivals sheet format and semantics
The arrivals drawer currently has two quirks that need to be ironed out to make for a less surprising user experience:

(1) Two different arrivals formats; one for compressed and one for expanded. Fix: consolidate to a single format that has everything important on it. The short version is missing delay and arrival time. Let's try to squeeze it all into one format smaller-format row.
(2) Different semantics for above-the-peek and below-the-peek arrivals rows. It's not even especially clear what the difference is, but there is one. The semantics should be less surprising - either the peek simply shows a continuation of the same list, or the drawer expansion reveals content of a distinctly different type. This could take a couple forms. The peek could show the next arrival for every line in a very compact view (basically the "starred stops" content) and then the below-the-peek could show a chronological list of arrivals.

bmander:
## What & why

Fixes #1707. The arrivals drawer had **two inconsistent formats** — a compact peek and an expanded list that used different logic (the compact one dropped delay/arrival detail) — plus a separate build-flavor **A/B "style" toggle** (flat Style A vs route-grouped Style B cards). This unifies everything into **one row format: one row per (route, direction)**.

Each row shows the route badge, a "heading toward" headsign, and a **horizontally-scrollable strip of per-trip ETA pills**. The *same* row renders in both the standalone arrivals screen and the home map drawer (the peek shows the promoted rows in that format), so the peek/expanded split #1707 describes is gone — as is the peek→card morph.

## Behavior

- **Ordering:** starred routes first, then the rest, each by the soonest **upcoming** ETA. Recent-past (negative) arrivals no longer count toward ordering and render as compact 75% pills, so a just-departed trip neither pulls its row to the top nor dominates the strip.
- **Interaction:** tap a pill → focus that trip's vehicle on the map; long-press a pill → per-trip menu (trip details / reminder / report); the row overflow ⋮ is route-level (star / show-only-this-route / schedule); tapping the row body frames the whole route.
- **Star semantics** are unchanged (the wholesale `routes.favorite` bit, #1751), applied as a reactive overlay so a toggle re-orders with no re-fetch.

## Retired

The `ARRIVAL_INFO_STYLE` A/B system in full: the "Sort by" dialog, the `BuildFlavorUtils` style plumbing, the per-flavor `buildConfigField`, the `MorphByProgress` peek→card animation, and the now-unused string/array/drawable resources.

## Notes

- Pure grouping + ordering live in `RouteRowGroup.kt`, keyed off a small `RouteDirectionItem` interface so they're JVM-unit-tested (`RouteRowGroupingTest`) without a `Context`.
- The report-flow arrivals picker keeps its per-trip rows (it must pick a single trip) via the retained shared row content.

## Testing

- JVM unit tests: grouping/ordering (incl. negative-ETA and all-past cases) + the arrivals ViewModel.
- Kotlin compile and Android Lint both under `-PwarningsAsErrors=true` (clean).
- `oba` and a former-Style-B flavor (`agencyY`) both assemble.
- Device-verified on a Pixel 7 Pro: peek/expand, pill overflow scroll, long-press menu, star re-ordering, two-directions-of-one-route.

## Summary by CodeRabbit

* **New Features**
  * Arrivals are now grouped by route and direction, with favorite routes shown first.
  * Updated arrivals UI to render route/direction rows with horizontally scrollable ETA pills and clearer route vs. trip actions.

* **Improvements**
  * Removed the legacy “Sort by”/arrival-style flow and the related “arrival info style” preferences across languages.
  * Updated menus (route-level overflow including star/show-only-this-route; trip actions via long-press) and refreshed drawer peek to match the new layout.

* **Tests**
  * Added JVM test coverage for route/direction grouping and favorite-based ordering.


bmander:
Addressed the review feedback (amended into the single commit):

- **Peek height measurement** (`ArrivalsPanel`): dropped the `expandProgress() <= 0f` guard on the per-row `onSizeChanged` write and now record the height unconditionally. Since there's no longer a peek→card morph, a route row is the same height collapsed or expanded, so recording anytime is correct — and it fixes the case where a row first laying out while the sheet is already open (e.g. restored expanded after a rotation) would never record, leaving the peek anchor stale. Constant height means unconditional recording can't move the anchor mid-open, so the original strand bug stays fixed.
- **Scroll-hint chevron** (`ArrivalRows`): now `contentDescription = null` (decorative), so TalkBack skips it.
- **Ordering test coverage** (`RouteRowGroup`): extracted a generic `orderGroupsByFavorite<T>(…, routeIdOf)` that `orderRouteGroupsByFavorite` delegates to, and added JVM tests for the starred-first rule + within-partition stability.

Left the docstring-coverage warning as-is (soft threshold; the missing entries are small private composables where docstrings would be noise).

Issue #1780, opened 2026-07-11
Add OTP 2.x GraphQL trip-planning support (on the new domain model)


bmander:
## Summary

Follow-up to #1712 (add OTP 2.x GraphQL trip-planning support — Sound Transit's OTP server dropped the
OTP 1.x REST `/routers/default` endpoint entirely, ST is now GraphQL-only). This issue carries forward
the schema/design findings from scoping that work, now that #1778 gives it a real target to map onto
(an app-owned domain model, not OTP1's vendored POJOs). Depends on #1778 landing first.

## Confirmed against the live ST OTP2 server

Root (`GET .../prod/otp`) self-reports `2.10.0-SNAPSHOT`, branch `orca-add-monorail-ibi` — matches
#1712. GraphQL schema is at `POST .../prod/otp/gtfs/v1`. Note there is **no plain `plan` field** — the
real entry point is `planConnection`, a Relay-style cursor connection:

```graphql
query Plan($origin: PlanLabeledLocationInput!, $destination: PlanLabeledLocationInput!, $dateTime: PlanDateTimeInput) {
  planConnection(origin: $origin, destination: $destination, dateTime: $dateTime, first: 3) {
    searchDateTime
    routingErrors { code description inputField }
    edges {
      node {
        start end duration numberOfTransfers
        legs {
          mode duration distance
          start { scheduledTime } end { scheduledTime estimated { time delay } }
          from { name lat lon stop { gtfsId code } }
          to   { name lat lon stop { gtfsId code } }
          route { shortName longName gtfsId agency { name timezone } }
          trip  { gtfsId tripHeadsign }
          legGeometry { points }
          steps { relativeDirection absoluteDirection streetName distance stayOn }
        }
      }
    }
  }
}
```
with `variables.origin/destination = {label, location: {coordinate: {latitude, longitude}}}` and
`variables.dateTime = {earliestDeparture: "<ISO-8601 offset datetime>"}`. Verified against a real
Westlake→Bellevue TC query — returned a working WALK→TRAM(2 Line)→WALK itinerary.

## Wire-format deltas from OTP1 (drives the adapter design)

| Aspect | OTP1 REST | OTP2 GraphQL (observed) |
|---|---|---|
| Itinerary/leg duration | ms, numeric | **seconds**, numeric |
| start/end times | epoch-ms as string | **ISO-8601 `OffsetDateTime`** string, already tz-aware |
[...]

Issue #1779, opened 2026-07-11
Migrate BikeRentalStation off the vendored OTP1 POJO library


bmander:
## Summary

`org.opentripplanner.routing.bike_rental.BikeRentalStation` (from the same unpublished
`edu.usf.cutr.opentripplanner.android:opentripplanner-pojos:1.0.0-SNAPSHOT` jar tracked in #1778) is a
separate POJO family from trip-itinerary planning, used for bikeshare station data. It's scoped out of
the itinerary domain-model migration to keep that change reviewable, but it's the dependency's only
remaining consumer once that lands — this issue tracks finishing the job.

## Why this one is bigger than it looks

Unlike `Itinerary`/`Leg` (which mostly stayed behind `TripResultsRepository`/`DirectionsGenerator`),
`BikeRentalStation` is baked directly into shared, cross-flavor UI surface:

- `map/compose/ObaMapCallbacks.kt` / `NoOpObaMapCallbacks.kt` — the shared map-callback *interface*
  (`onBikeClick(station: BikeRentalStation)`, `onBikeInfoWindowClick(station: BikeRentalStation)`),
  implemented by Compose call sites in `ui/home/map/MapFeature.kt` and
  `report/ui/InfrastructureIssueScreen.kt`.
- `map/compose/BikeInfoWindow.kt` — a `@Composable` reads `station.name/.isFloatingBike/.bikesAvailable/
  .spacesAvailable` directly.
- `map/render/MapRenderState.kt` — `BikeMarker.station: BikeRentalStation` is a field on the declarative
  render-state snapshot Compose renderers observe.
- `map/bike/BikeStationsRepository.kt` — its public contract (`getStations(): Result<List<BikeRentalStation>>`)
  returns the raw OTP type, unlike `TripResultsRepository`, which does translate.

## Suggested approach

Same shape as the itinerary migration in #1778: a small app-owned `BikeStation` data class (wire DTOs
already exist in `api/contract/BikeModels.kt`), a `BikeStationAdapters.kt` mapping DTO→domain, then
retarget `ObaMapCallbacks`/`BikeInfoWindow`/`MapRenderState`/`BikeStationsRepository`/`MapNavigation`
onto it. Once this lands, the `opentripplanner-pojos` dependency can be removed from `build.gradle`
entirely.

Tracked under #1778.

Issue #1778, opened 2026-07-11
Retire vendored OTP1 trip-planning POJOs (opentripplanner-pojos) in favor of an app-owned domain model


bmander:
## Summary

The trip-planning stack's internal domain model is currently `org.opentripplanner.api.model.*` /
`org.opentripplanner.api.ws.*` / `org.opentripplanner.routing.core.*` — Java POJOs pulled in via:

```
edu.usf.cutr.opentripplanner.android:opentripplanner-pojos:1.0.0-SNAPSHOT
```

fetched from a personal GitHub raw-file "maven repo" (`CUTR-at-USF/cutr-mvn-repo`), not a real package
registry. The jar itself is an **unpublished, 2017-era SNAPSHOT** — the project has no ability to patch,
version, or evolve it.

## Why this matters

- **It's not just a wire-format detail.** These POJOs are used as the app's actual internal domain
  model: they're constructed by the `/plan` JSON decoder (`api/contract/OtpPlanModels.kt`), and then
  flow — largely unconverted — through `TripPlanRepository`, `TripPlanViewModel`/`TripPlanScreen`,
  `TripResultsViewModel`/`TripResultsScreen`, `DirectionsMapViewModel`/`DirectionsMapController`,
  `DirectionsGenerator`, `ConversionUtils`, `ItineraryDescription`, and the trip-plan-monitor
  (`TripMonitorDecider`/`TripPlanMonitor`/`TripPlanMonitorService`) — 32 files total reference
  `org.opentripplanner.*` today, including several Compose UI/ViewModel public signatures
  (`PlanResult.Success.itineraries: List<Itinerary>`, `TripResultsSheet(itineraries: List<Itinerary>, ...)`,
  etc.), not just a repository-internal detail.
- Two real Android-serialization boundaries currently depend on these POJOs being
  `java.io.Serializable` (a `Bundle.putSerializable(OptimizeType)` in `TripRequestBuilder`, and an
  `Intent.putExtra(ArrayList<Itinerary>)` in `TripPlanMonitorService.notifyChange()` / read back in
  `TripPlanScreen.maybeRestoreFromIntent()`).
- It blocks cleanly adding OTP 2.x GraphQL support (#1712): mapping OTP2 responses onto OTP1's own
  vendored POJOs (as originally suggested in that issue) just extends reliance on a dependency we
  don't control, and doesn't fix the underlying UI-leakage problem.

## Plan

[...]

PR #1782, opened 2026-07-11
Trip planning: replace vendored OTP1 POJOs with an app-owned domain model

bmander:
## Summary

The trip-itinerary path (`TripPlanRepository`, `DirectionsGenerator`, the Compose trip-plan/results UI, the trip-plan monitor) read/wrote `org.opentripplanner.api.model.*` directly — POJOs from an unpublished 2017 snapshot jar (`edu.usf.cutr.opentripplanner.android:opentripplanner-pojos:1.0.0-SNAPSHOT`) that the project doesn't control, leaking as far as ViewModel/Compose signatures.

- Introduces `TripItinerary`/`TripLeg`/`TripPlace`/etc. (`directions/model/TripItinerary.kt`) as the app's own domain model, mapped once at the wire boundary (`api/adapters/TripPlanAdapters.kt`, following the existing `RegionAdapters.kt`/`ArrivalAdapters.kt` convention), and retargets every downstream consumer onto it.
- Server-provided timestamps are typed `ServerTime` (`org.onebusaway.android.time`), not raw `Instant` — matching this codebase's existing clock-domain-tagging pattern.
- `from`/`to`/`startTime`/`endTime` are non-null in the domain model: a well-formed OTP leg always has two endpoints and absolute times, so the adapter asserts that once (throwing on a genuinely malformed response) instead of every consumer repeating a `!!`.
- Fixes two real Android-serialization boundaries that relied on the OTP1 POJOs implementing `java.io.Serializable`: `TripRequestBuilder`'s `OptimizeType` Bundle write (now a plain `Boolean`), and the trip-monitor "trip changed" notification `Intent` (now JSON-encoded via kotlinx.serialization instead of `putExtra(String, Serializable)`).
- Removes several now-dead helpers that only existed to support the old POJOs: `ConversionUtils.fixTimezoneOffsets`/`getDuration`/`parseOtpDate`, `DirectionsGenerator.getTotalTimeTraveled` — all confirmed zero-caller before deletion.

Bikeshare (`BikeRentalStation`) migration and OTP 2.x GraphQL support are scoped out and tracked separately under the umbrella tracking issue.

Tracking: #1778
Follow-ups: #1779 (bikeshare), #1780 (OTP 2.x GraphQL)

## Test plan

- [x] `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin -PwarningsAsErrors=true` — clean
- [x] `./gradlew :onebusaway-android:lintObaGoogleDebug -PwarningsAsErrors=true` — clean (no new baseline entries)
- [x] `./gradlew :onebusaway-android:testObaGoogleDebugUnitTest -PwarningsAsErrors=true` — all pass, including a new test covering the adapter's "missing required field" contract
- [x] On-device: built, installed, and manually verified the trip-planning flow (origin/destination entry → itineraries render → directions list → map route) on a physical device
- [ ] Live trip-plan monitoring (the notification `Intent` JSON round-trip) could not be manually exercised — that UI entry point is currently deprecated/removed from the app; covered by unit tests (`TripMonitorDeciderTest`) only

## Summary by CodeRabbit

* **New Features**
  * Added a unified trip itinerary model supporting transit, walking directions, stops, steps, geometry, timing, delays, and route modes.
  * Trip plans now decode into app-ready itinerary data for displaying results, directions, maps, and trip monitoring.
  * Added reliable itinerary serialization for restoring trip plans from notifications.

* **Bug Fixes**
  * Improved handling of missing required trip data and unknown direction or location values.
  * Preserved accurate trip timing, duration, delay, and geometry information across trip planning features.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/TripPlanAdapters.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/TripPlanAdapters.kt
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.api.adapters
+
+import org.onebusaway.android.api.contract.OtpItineraryDto
+import org.onebusaway.android.api.contract.OtpLegDto
+import org.onebusaway.android.api.contract.OtpLegGeometryDto
+import org.onebusaway.android.api.contract.OtpPlaceDto
+import org.onebusaway.android.api.contract.OtpWalkStepDto
+import org.onebusaway.android.directions.model.TripAbsoluteDirection
+import org.onebusaway.android.directions.model.TripItinerary
+import org.onebusaway.android.directions.model.TripLeg
+import org.onebusaway.android.directions.model.TripLegGeometry
+import org.onebusaway.android.directions.model.TripMode
+import org.onebusaway.android.directions.model.TripPlace
+import org.onebusaway.android.directions.model.TripRelativeDirection
+import org.onebusaway.android.directions.model.TripStep
+import org.onebusaway.android.directions.model.TripVertexType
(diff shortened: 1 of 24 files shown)


bmander:
Addressed all 3 actionable review comments in 493d6d64:

1. **`toTripItineraries()` can crash on corrupted Intent extras** — fixed at the source (`TripItinerary.kt`) with `runCatching`, degrading to an empty list on decode failure.
2. **Uncaught `SerializationException` in `TripPlanScreen.maybeRestoreFromIntent`** — resolved by fix #1; that call site now can't throw since `toTripItineraries()` never propagates the exception.
3. **Unguarded `!!` on endpoint lat/lon in `DirectionsMapController`** — `from`/`to` are guaranteed non-null but their coordinates aren't (a place can lack geographic identity), so a missing coordinate now just skips that pin/start-framing instead of crashing map rendering; leg polylines still draw regardless.

Added `TripItineraryJsonTest` covering the JSON round-trip and the malformed/truncated-input degrade. Compile, lint, and full unit test suite all pass.

Issue #1827, opened 2026-07-13
Adjacency focus: on stop tap, draw the routes serving it, dim other stops, badge routes to tap into


bmander:
## Summary

Introduce a new map state — call it **adjacency focus** — that activates when the user taps a stop. While active, the map:

1. **Draws the route lines** of the trips currently coming up at that stop (the routes shown in the arrivals drawer — *not* every route that has ever served the stop).
2. **Minimizes every stop that isn't on one of those routes** to its zoomed-out "dot" appearance, so the relevant network stands out.
3. **Shows a tappable route badge for each drawn route**, and tapping a badge enters the existing single-route "route mode" for that route.

This is a large, multi-part change spanning the arrivals data layer, the declarative map render model, both platform renderers (Google + MapLibre), and the home-screen state machine. This issue is the plan; it should probably land as a **stack of PRs** (see Phasing).

---

## Precise definitions (need sign-off — see Open Decisions)

- **"Routes currently serving the stop"** — derived from the loaded arrivals, *not* `ObaStop.routeIds` (which is every route that ever serves the stop). **DECIDED: all upcoming, matching the drawer 1:1** — `routeGroups.map { it.routeId }.toSet()` (every route with an upcoming arrival, scheduled or tracked). Not narrowed to the real-time `predicted` subset. (The `predicted` flag — `ArrivalInfo.kt:61,173` — is still available if we later want to style tracked vs schedule-only lines differently.)
- **"Adjacent by one route"** — a stop stays full-size iff it is served by at least one route in the focused set: `stop.routeIds ∩ focusedRouteIds ≠ ∅`. Every other visible stop is forced to its minimized dot. (A route's stops == the stops it serves, so "on the drawn line" and "served by the route" are the same set.)
- **"Minimized appearance"** — the existing zoomed-out `DOT` icon kind (`StopZoomBand.kt`), applied regardless of the current zoom level for non-adjacent stops.

---

## What already exists (reuse map)

The declarative-map rework left most of the primitives in place:

| Need | Already there | Anchor |
|---|---|---|
| Draw **multiple** route lines | `MapRenderSnapshot.routePolylines: List<RoutePolyline>` — both renderers already loop over the list; only the *producer* is single-route | `map/render/MapRenderState.kt:193`; `GoogleMapRenderer.kt:233`, `MapLibreRenderer.kt:176` |
[...]

PR #1832, opened 2026-07-13
Adjacency focus phase 1 (data): focused route-id set + multi-route shape fetch

bmander:
Phase 1 ("Data") of the adjacency-focus PR stack (#1827). **Pure data-layer groundwork — no UI, no user-visible behavior change.** The new repository is unconsumed and `focusedRouteIds` is unread until phase 2, so there is no runtime surface to drive here.

## What's in it

**1. Focused route-id set: arrivals → `HomeViewModel`**
- A pure `focusedRouteIds(routeGroups)` helper (on the generic `routeIdSet`, mirroring the existing `orderGroupsByFavorite`/`orderRouteGroupsByFavorite` pair) derives the drawer's routes **1:1** — every route with an upcoming arrival, scheduled or tracked (the issue's decided definition, not the `predicted` subset).
- It's derived from the **computed** `ArrivalsData.routeGroups` (which reflects the past-arrivals filter), so `ArrivalsLoaded` now carries a prebuilt `@Volatile lastLoadedSnapshot` set in both `getArrivals` fold branches, rather than being re-derived from the raw `StopArrivals` on read.
- `HomeActivity` relays it into `HomeViewModel.onArrivalsLoaded`, which stores it on a plain `focusedRouteIds` property (coordination state like `settledSheet`), updated on every load and reset on focus change/clear.

**2. `AdjacencyRouteShapeRepository` — multi-route shape fetch**
- Fetches whole-route shapes for a **set** of routes at once: bounded concurrency via a `Semaphore` (deliberately not `limitedParallelism` — the Retrofit calls are main-safe suspend calls that hold no dispatcher thread while awaiting, so a limited dispatcher wouldn't actually bound in-flight requests; documented at the call site), `SingleFlight` in-flight de-dup, and a success-only 32-entry LRU with a 10-minute TTL for sequential stop focuses. Partial failures still land in `failedRouteIds` and are never cached, so later focuses retry them.
- The `Location → GeoPoint` conversion runs off the main thread. `@Singleton` because it owns the in-flight de-dup and completed-result cache state.

## Testing
Pure logic (route-id derivation, and the fetch's bounded concurrency / in-flight de-dup / completed-cache hits / LRU eviction / TTL expiry / failure retry / partial-failure / cancellation behavior) is JVM-unit-tested. Focused repository/cache tests pass; compiles clean under `-PwarningsAsErrors=true`.

## Follow-ups (rest of the stack, per #1827)
Lines producer + `MapDirective`, stop minimization, de-focused vehicles, Google badges, then MapLibre badges.

## Summary by CodeRabbit

* **New Features**
  * Improved adjacency map rendering by loading whole-route shapes for multiple routes with bounded concurrent fetches and in-flight de-duplication.
  * Enhanced map focus coordination by tracking the focused stop’s route ID set through arrivals updates.
* **Bug Fixes**
  * Arrivals “last good” behavior now keeps derived focused-route data consistent across normal and fallback loads.
  * Route-shape loading tolerates partial failures by returning successful routes while reporting which routes failed.
* **Tests**
  * Added a new adjacency route-shape repository test suite (concurrency, de-duplication, caching, partial failures) and updated home/route grouping tests for the new route-set behavior.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/app/di/RepositoryModule.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/app/di/RepositoryModule.kt
@@ -41,6 +41,8 @@ import dagger.Module
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import javax.inject.Singleton
+import org.onebusaway.android.map.AdjacencyRouteShapeRepository
+import org.onebusaway.android.map.DefaultAdjacencyRouteShapeRepository
 import org.onebusaway.android.map.DefaultRouteMapRepository
 import org.onebusaway.android.map.RouteMapRepository
 import org.onebusaway.android.map.bike.BikeStationsRepository
@@ -236,4 +238,12 @@ abstract class RepositoryModule {
     abstract fun bindTripObservationFetcher(
         impl: DefaultTripObservationFetcher
     ): TripObservationFetcher
+
+    // Adjacency focus (#1827): @Singleton because it owns the SingleFlight de-dup state that coalesces
+    // concurrent/re-tap route-shape fetches (the trip-observation fetcher precedent above).
+    @Binds
+    @Singleton
+    abstract fun bindAdjacencyRouteShapeRepository(
+        impl: DefaultAdjacencyRouteShapeRepository
+    ): AdjacencyRouteShapeRepository
 }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/BoundedLruCache.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/data/BoundedLruCache.kt
@@ -43,6 +43,10 @@ internal class BoundedLruCache<K : Any, V : Any>(private val maxSize: Int) {
         entries[key] = value
     }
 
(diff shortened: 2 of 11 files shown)


bmander:
## Performance handoff before the next adjacency phase

On-device testing of the phase-3 branch exposed a cost that should be pulled forward into this data-layer PR.

A Pixel 7 Pro reached Android thermal status 3 (`severe`) during repeated stop-focus testing. The request log showed batches of roughly 6–10 `stops-for-route` calls per focused stop, and a route shared by two stops was fetched twice within about 11 seconds. The current `SingleFlight` correctly coalesces concurrent callers, but completed results are discarded, so it cannot help across sequential stop focuses.

Proposed amendment to this PR:

- add a bounded completed-result cache in `AdjacencyRouteShapeRepository`, keyed by route ID;
- keep cache ownership in the repository/data layer rather than a ViewModel or renderer;
- preserve the existing bounded concurrency, partial-failure behavior, and in-flight de-duplication;
- cover cache hits, eviction/expiry, and failed-result behavior in the repository tests.

The phase-3 draft will document this as a known limitation. I recommend treating the cache as a prerequisite before starting the next adjacency phase.

bmander:

PR #1834, opened 2026-07-13
Adjacency focus phase 2: draw route lines

bmander:
Phase 2 ("Lines") of the adjacency-focus PR stack for #1827. Stacked on #1832.

## What changed

- Added an `AdjacencyMapController` that loads the focused stop's upcoming route shapes through the phase-1 repository and publishes the resolved whole-route polylines into the existing shared render model.
- Added Home-to-map directives for starting and clearing adjacency focus, including restored-focus ordering and stale-focus validation.
- Defined lifecycle ownership against single-route mode: route mode wins, entering it tears adjacency down, and a late adjacency clear cannot erase route-mode lines.
- Kept the camera fixed and reused the existing Google and MapLibre `routePolylines` rendering paths.
- Split route-polylines into their own distinct snapshot flow in both flavors, retaining equal native lines across stop-only updates instead of removing and recreating long adjacency shapes.
- Retained stop markers now compare against their previously rendered model and skip unchanged native position writes (plus unchanged Google z-index writes).

Partial route-shape failures draw the routes that resolved. Repeated arrivals refreshes with the same stop and route set do not refetch shapes, and stop-only snapshot updates no longer rebuild the resolved native route lines.

## Scope

This PR intentionally contains lines only. Stop minimization, de-focused multi-route vehicles, and route badges remain later phases of #1827.

## Validation

- `./gradlew :onebusaway-android:testObaGoogleDebugUnitTest :onebusaway-android:testObaMaplibreDebugUnitTest -PwarningsAsErrors=true`
- `./gradlew :onebusaway-android:lintObaGoogleDebug -PwarningsAsErrors=true`

Both passed.

On-device interaction validation on a Pixel 7 Pro (repeated stop focus plus pan/zoom over 60 seconds) exercised 1,489 frames: 40 missed the frame deadline (2.69%), with one 45-frame `Choreographer` stall. App CPU averaged 85.5% (p90 175%), and the already-warm device moved from thermal status 2 to 3. This confirms the amended paths were tested under interaction rather than only at idle, while also documenting remaining thermal pressure outside the eliminated annotation churn.

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/GoogleMapRenderer.kt
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/GoogleMapRenderer.kt
@@ -50,6 +50,7 @@ import org.onebusaway.android.map.render.ContinuationBadgeBitmaps
 import org.onebusaway.android.map.render.CorrectionSmoother
 import org.onebusaway.android.map.render.GeoPoint
 import org.onebusaway.android.map.render.MapPing
+import org.onebusaway.android.map.render.MapRenderSnapshot
 import org.onebusaway.android.map.render.MapRenderState
 import org.onebusaway.android.map.render.MarkerRendering
 import org.onebusaway.android.map.render.PingTarget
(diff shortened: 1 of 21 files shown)


bmander:
## Rendering-performance handoff before the next adjacency phase

On-device phase-3 testing also exposed rendering amplification caused by the whole-route lines introduced here.

While interacting with the map, the app used roughly 90–159% CPU, Android reported repeated 40–60-frame main-thread stalls, and the Pixel reached thermal status 3 (`severe`). Profiling showed the dominant work on `GL-Map`, the app main thread, and Google Maps render threads.

The relevant rendering path currently collects the combined static snapshot and calls `renderStatic()`; that method removes and recreates all route polylines whenever any static field changes, including a nearby-stop list update. The stop reconcile then also writes native marker positions and z-indexes unconditionally for retained markers. With multiple long adjacency lines plus up to 200 stops, viewport loading magnifies both costs.

Proposed amendment to this PR:

- reconcile the route-polyline layer independently and retain unchanged native polylines across stop updates;
- avoid native marker position/z-index writes when the values are unchanged;
- validate with an interaction profile, not only an idle snapshot.

The completed-result route-shape cache is tracked separately on #1832. The phase-3 draft will call both items out as known limitations, and I recommend fixing them before starting the next adjacency phase.

bmander:

PR #1836, opened 2026-07-13
Adjacency focus phase 3: minimize non-adjacent stops

bmander:
Phase 3 ("Stop minimization") of the adjacency-focus PR stack for #1827. Stacked on #1834.

## What changed

- Carries the GTFS directions of the actual upcoming trips from the arrivals snapshot instead of treating an entire route as adjacent.
- Keeps the parent PR's exact trip-pattern geometry for rendering; route geometry is not expanded back to every branch of a route.
- Fetches direction-to-stop membership as an independent, bounded, single-flight, cached pass over route-map data. Geometry can render as soon as exact shapes resolve instead of waiting for stop metadata.
- Keeps the focused stop and stops served by the relevant trip directions emphasized while minimizing unrelated stops, including the opposite side of the road for the same route.
- Falls back conservatively to whole-route membership when trip direction metadata, route direction metadata, or a route response is missing.
- Adds the minimized stop treatment to both Google Maps and MapLibre, including the favorite-stop dot treatment.
- Covers arrival direction aggregation, route response mapping, filter fallback behavior, controller lifecycle, zoom-band icon selection, and Home-to-map coordination.

The parent PR's clipped, fixed-width, upstream/downstream exact trip-pattern lines remain unchanged; this phase only narrows which stop markers stay emphasized.

## Rebase and performance status

This branch is rebased onto the latest #1834 performance work. Exact shape geometry remains cached and renderer transforms remain viewport-clipped. Route stop membership is cached separately from geometry so phase 3 does not retain or render whole-route polylines, and its metadata request cannot delay exact-shape rendering.

The PR remains a draft pending final on-device behavior and performance validation of the rebased stack.

## Validation

- `./gradlew --daemon --build-cache :onebusaway-android:testObaGoogleDebugUnitTest --tests org.onebusaway.android.map.AdjacencyMapControllerTest --tests org.onebusaway.android.map.AdjacencyRouteShapeRepositoryTest --tests org.onebusaway.android.map.StopAdjacencyTest --tests org.onebusaway.android.ui.arrivals.ArrivalsAdjacencyTest --tests org.onebusaway.android.ui.home.HomeViewModelTest -Pkotlin.warningsAsErrors=true`
- `./gradlew --daemon --build-cache :onebusaway-android:testObaMaplibreDebugUnitTest --tests org.onebusaway.android.map.AdjacencyMapControllerTest --tests org.onebusaway.android.map.AdjacencyRouteShapeRepositoryTest --tests org.onebusaway.android.map.StopAdjacencyTest --tests org.onebusaway.android.ui.arrivals.ArrivalsAdjacencyTest --tests org.onebusaway.android.ui.home.HomeViewModelTest -Pkotlin.warningsAsErrors=true`


bmander:

Issue #1957, opened 2026-07-19
Register for OBACloud service-alert push notifications (new push_registrations API)


aaronbrethorst:
## Summary

OBACloud can now send push notifications for service alerts to riders who have opted in to notifications (OneBusAway/obacloud#981). For that to work, the app needs to register the device's FCM registration token with the OBACloud server proactively — on launch and whenever the token rotates — via a new registration API. Today the server only learns about tokens as a side effect of alarm creation, which is not enough (see rationale below).

## Rationale — why the app needs to do anything

When a transit agency publishes a service alert through OBACloud, they can now attach a push notification that fans out to every opted-in device in the region, in the rider's language. The server's audience for that fan-out is its `push_tokens` table. Right now that table is populated only when a rider creates a trip alarm, which has three problems:

1. **Coverage:** riders who allow notifications but never set an alarm are invisible — the agency's "estimated reach" undercounts, and those riders never receive alerts.
2. **Locale:** alarm registrations don't carry the device locale, so the server can't pick the right translation of the alert copy. Devices without a locale get English.
3. **Freshness:** the server prunes tokens not seen for 180 days. Without periodic re-registration, even alarm-derived tokens go stale and reach decays.

The new API fixes all three, but only if the app calls it.

## API contract

Both endpoints live on the OBACloud server (same host and trust model as the v2 alarms API — unauthenticated, ownership is possession of the token, rate-limited to 30 requests/minute/IP).

### Register / refresh

```
POST /api/v2/regions/{region_identifier}/push_registrations
```

Form params:

| param | value |
|---|---|
| `token` | the FCM registration token (from `FirebaseMessaging.getInstance().token` / `onNewToken`) |
| `operating_system` | `"android"` (literal) |
| `locale` | the device's BCP-47 locale tag, e.g. `Locale.getDefault().toLanguageTag()` → `"es-MX"`, `"zh-Hant-TW"` |
| `test_device` | `true`/`false` — see below |

Responses: `204 No Content` on success (create **and** update are the same call — it's an upsert on `(region, token)`); `422` with `{error, messages}` on validation failure; `404` for an unknown region; `429` when throttled.

### Unregister

```
DELETE /api/v2/regions/{region_identifier}/push_registrations
```

with `token` param. `204` on success, `404` if the token wasn't registered (safe to ignore).

## Implementation notes

**When to register:**
[...]

Issue #1908, opened 2026-07-16
Focused-trip data layer: stops-for-route fetched via two paths, shapes double-cached, focusedTrips in three copies


bmander:
From the 2026-07-16 debt review (finding 8). **Severity: medium.**

Three related smells introduced by #1865/#1893:

1. **stops-for-route via two independent repository paths.** `FocusedTripRepository.getStops` uses `RouteStopsDataSource.stopsForRoute` with its own `routeStopCache` (`FocusedTripRepository.kt:124-125,190-194`), while `RouteMapRepository.getRoute` hits the same wire endpoint uncached via `MapDataSource.routeMap` (`MapDataSource.kt:72-73`), into a different model shape (`RouteStopGroup` vs `RouteMapStop`); `RouteInfoRepository` is a third consumer. "Focus a stop, then select one of its routes" fetches stops-for-route twice with no coalescing, and the two projections can come from different fetches. Refactor: one caching stops-for-route repository both project from. (medium)

2. **Shape geometry double-cached under the same shapeId with divergent policies.** `TripObservationRepository`'s `ShapeCache` (`Polyline`, LRU 100, no TTL) vs `FocusedTripRepository.shapeCache` (`List<GeoPoint>`, LRU 32, 10-min TTL — `FocusedTripRepository.kt:71-73,126-127,209-219`). ~2x memory per focused shape, two eviction policies, and the 10-minute TTL on effectively immutable GTFS data is an unexplained freshness threshold (contrast `MAX_CACHED_SHAPES`, which carries a full rationale). Drop or document the TTL. (small)

3. **The focused-trip set has three copies with scattered manual resets.** `ArrivalsRepository.lastGood.loaded.focusedTrips` (`ArrivalsRepository.kt:486-492`) → `HomeViewModel.focusedTrips`, a plain `var` (`HomeViewModel.kt:155`) reset imperatively at four sites (175, 363, 471, 545) → `RouteMapController.stopFocusSession.trips`. Deriving the set from the focus state flow removes the "forgot to clear on transition X" bug class. (medium)

bmander:
All three findings addressed and merged into main:
1. Stops-for-route unified behind one caching repository — #1921
2. Shape double-cache dropped, reusing shared Polyline cache — #1919
3. focusedTrips now derived from the focus state flow — #1922

Closing.

PR #1958, opened 2026-07-19
Register devices for OBACloud service-alert push notifications (#1957)

bmander:
## Summary

Registers devices proactively with OBACloud's **push_registrations** API (#1957) so service-alert push notifications reach every rider who has notifications enabled — not just those who happen to create a trip alarm.

Today OBACloud only learns a device's FCM token as a side effect of creating a trip alarm, which means:
- Riders who enable notifications but never set an alarm are never registered.
- Device locale is never captured, so alerts default to English.
- Tokens age out after 180 days without re-registration.

## What's included

- **`PushRegistrationWebService`** — form `POST` register (`token`, `locale` as-is, `test_device`, `description`, `operating_system=android`) and `DELETE` unregister, on the region's sidecar host via `@Url`, mirroring `ReminderWebService`. Provided through `NetworkModule.plainRetrofit`.
- **`PushRegistrationDecision` (`decidePushRegistration`)** — a pure, Android-free function that reconciles the desired registration against the last-sent one → register / re-register / unregister / no-op, including the 24h keep-alive. JVM-unit-tested.
- **`PushRegistrationManager`** — derives the desired registration from OS notifications-enabled + region + FCM token + locale + the test-device settings; persists the last registration and when it was sent (so opt-out DELETEs the right token, redundant POSTs are skipped given the 30 req/min/IP limit, and the server's `last_seen_at` stays fresh). Driven reactively off the region flow and the preferences it reads, plus a `ProcessLifecycleOwner` `ON_START` resync to catch OS-permission changes made in system settings.
- **"Register as test device" toggle + "Test device name" field** on the Advanced settings screen → OBACloud's "Test users only" preview audience. The name is capped at the server's 255-character limit.

**Opt-in** is the OS notification permission — there is no separate in-app toggle.

## Triggers covered

App launch, app foreground, FCM token rotation, region change, the test-device settings, and notification opt-in/opt-out.

## Two bugs found during device testing

Both were found by exercising the real endpoint against the live sidecar, and both are fixed here.

**1. Test-device registration was rejected outright.** A `test_device=true` POST returns `422 {"error":"Unable to register device","messages":["Description can't be blank"]}` — the server requires a `description` field that the contract table in #1957 never sent, so the toggle could not work at all. The requirement landed server-side after that table was written, which is why it wasn't implemented initially. It is now specified in [OBACloud's published documentation](https://developer.onebusaway.org/projects/obacloud): free text ≤255 chars identifying the device to admins, required only when `test_device=true`, and cleared server-side when a device is demoted.

**2. HTTP error responses were silently swallowed.** `runCatching { … }.onFailure` only fires for *thrown* exceptions, but a non-2xx is a **successful call returning an unsuccessful response** — so the 422 above produced no log output whatsoever. Worse, because a failed call persists nothing, the same doomed request repeated on every single app foreground, invisibly, against a rate-limited endpoint. Failures are now branched on explicitly and logged with the status code and the server's error body (never the token, and never the URL — the unregister URL carries the token as a query param).

## Conformance with the published OBACloud documentation

OBACloud's [push-notifications documentation](https://developer.onebusaway.org/projects/obacloud) was published after this PR was opened. Checking the implementation against it found it already conformant on every point below — notably the dedupe rule, which was derived independently from the iOS client's source and matches the documented behaviour exactly.

| Documented requirement | This PR |
|---|---|
| `POST /api/v2/regions/{id}/push_registrations` | `registrationUrl()` builds exactly this |
| `DELETE …?token={token}` | token as a query param |
| `404` on DELETE = treat as success | explicit `HTTP_NOT_FOUND` check |
| `operating_system` must be `ios`/`android` | `"android"` |
| `locale` BCP-47, **do not normalize** | `Locale.getDefault().toLanguageTag()`, sent as-is |
| `test_device` sent on **every** call (omission resets it to false) | non-null `Boolean`, always sent |
| `description` required iff `test_device=true` | `testDevice = description != null` — holds by construction |
| `description` ≤255 chars | capped at the input field **and** before the wire |
| Dedupe on `(token, region, locale, test_device, timestamp)`; re-POST on change or >24h | `PUSH_REFRESH_INTERVAL = 24.hours`, identical semantics |
| Register on launch/foreground, token rotation, region change | all three wired |

The one gap this surfaced was the **255-character cap**, now fixed. The rider-entered name was unbounded, so a long one would 422 — and since a failed registration persists nothing, that doomed POST would have repeated on every foreground. It is bounded in two places: at the input field, so the limit is felt while typing, and in `buildTarget()`, so a value stored before that bound existed cannot reach the wire.

**Two deliberate divergences**, both worth a maintainer's eye:

- **No API key.** The doc says non-OBAKit clients "should send an API key the same way" OBAKit appends `key`/`app_uid`/`app_ver`/`version=2`. This client sends none, matching the sibling `ReminderWebService` on the same sidecar, which already ships that way. Device testing returns `204` without it and the rate limit is per-IP rather than per-key, so there is no functional impact today — but if OBACloud intends to enforce or attribute by key, both sidecar services should change together.
- **No 429 backoff.** The doc's guidance on 429 is diagnostic ("your dedupe is broken"), iOS implements none, and the 24h dedupe keeps a healthy device at roughly one POST per day.

One documentation ambiguity worth flagging upstream: `token` is described as "APNs device token, lowercase hex", which is iOS-specific — FCM tokens are neither. Since `operating_system` accepts `android` and the live server returns `204`, this is evidently phrasing rather than a constraint.

## Parity with the iOS client

Comparing against `OneBusAway/onebusaway-ios` (whose `docs/push-notifications.md` §5.2 is a conformance checklist for non-OBAKit clients) surfaced two more gaps and improved a third:

- **24h keep-alive.** Previously we re-posted only when an input *changed*, so a device whose token, region and locale never change would go untouched until the server's 180-day `last_seen_at` prune dropped it and alerts silently stopped — defeating the "freshness" goal of #1957 that this PR's own KDoc claimed to address. `decidePushRegistration` now takes elapsed-time and re-registers after `PUSH_REFRESH_INTERVAL` (24h, matching iOS). An unknown timestamp counts as stale, so existing records refresh once on upgrade. No `WorkManager` job: every foreground already runs the reconcile, and a rider who hasn't opened the app in 180 days has no reach to preserve.
- **An unresolved region is no longer mistaken for an opt-out.** `RegionRepository` seeds its region flow to null and fills it asynchronously, so a cold-start reconcile could read "no region" as "the rider opted out", `DELETE` the registration, and re-`POST` it seconds later. Notifications being off is now the only definitive opt-out; unresolved inputs bail and wait for the collector to re-fire. (The underlying `RegionState` limitation affects other consumers too — filed as #1969.)
- **The test-device description is now rider-entered** rather than derived from `Build.MODEL`, which could not distinguish two identical handsets in the admin UI. An unnamed device registers as an *ordinary* device rather than sending a request the server is guaranteed to reject — the same gating iOS uses.
- **Server rejections are reported to Crashlytics.** Registrations are the server's only audience source for service alerts, so a systematic failure must not be visible only in one developer's logcat — which is exactly how bug 1 above stayed hidden. Transport failures stay local; reporting every offline moment would drown the signal.

## Device verification

On a Pixel 7 Pro (API 36) against the live `sidecar.onebusaway.org`, region 1:

| Scenario | Result |
|---|---|
| Launch with notifications denied | No traffic — correct no-op |
| Opt-in via system settings, then foreground | `POST` → **204** |
| Opt-in via the in-app permission dialog | `POST` → **204**, fired from the result callback |
| Repeat foregrounds | **Zero** extra POSTs — no-op path holds |
| Opt-out (permission revoked) | `DELETE …?token=…` → **204**, record cleared |
| Test device with a description | `POST` → **204** |
| Upgrade from a build without the timestamp | One refresh `POST` → **204**, then quiet |

Two details worth recording. The in-app dialog case is the one the shared permission helper exists for: the system dialog launches as a *transparent* activity in the app's own task, so the host activity is only paused and `ProcessLifecycleOwner`'s `ON_START` never fires — logs confirm no stop/start cycle, so without the result-callback resync the device would have stayed unregistered until the next cold launch. And the DELETE contract question raised in an earlier revision of this description is **settled**: the token as a query param is both accepted (204) and now the documented form.

**Not yet device-exercised** (unit-tested only): the `Reregister` path via a region change, the unnamed-test-device downgrade, renaming a test device, and the 255-character truncation.

## Notes

- **`main` has been merged in.** The notification-permission conflict resolved toward this branch's Compose `rememberNotificationPermissionRequest()` helper, which requests `POST_NOTIFICATIONS` behind the same API-33 guard as the request-code flow it replaces *and* resyncs the registration from the result callback. The now-unreferenced `PermissionUtils.NOTIFICATION_PERMISSION_REQUEST` constant is removed. The branch's files are also reformatted for the Spotless gate (#1970–#1972) that landed on `main` while this was open.
- No heuristics remain in this change. The `description` semantics are documented upstream and the value is rider-supplied rather than inferred.
- The repo's `CLAUDE.md` documents a OneSignal push setup that doesn't exist in code (the app is FCM-only). Left untouched here; worth a separate cleanup.

## Testing

- `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin -PwarningsAsErrors=true` — clean.
- `./gradlew spotlessCheck` — clean.
- `./gradlew :onebusaway-android:testObaGoogleDebugUnitTest` — 29 push-registration tests pass (11 decision + 18 manager), covering the decision matrix, the persist/retry semantics across all four re-register failure quadrants, the keep-alive window, the region guard, the test-device gating, the description truncation, and the failure-reporting split.

## Summary by CodeRabbit

* **New Features**
  * Added automatic push registration synchronization tied to app start/foreground, region availability, notification enablement, and FCM token changes.
  * Added Advanced Settings for configuring test-device service alerts, including a capped optional device name.
  * Added Android 13+ notification permission prompting that triggers an immediate push re-registration after approval.
* **Bug Fixes**
  * Improved push registration reconciliation for safer idempotency, refresh timing, and resilient handling of transient/server failures.
* **Tests**
  * Added/expanded unit and integration tests covering push registration decisioning, sync/persistence, and test-device name normalization.


bmander:
# 🤖 Code review by Claude (Fable 5)

*Automated review generated with [Claude Code](https://claude.com/claude-code).*

## Overview

Adds proactive device registration against OBACloud's new `push_registrations` API: a Retrofit contract (`PushRegistrationWebService`), a pure reconciliation function (`decide()` in `PushRegistrationDecision.kt`), a reactive `PushRegistrationManager` singleton kicked from `Application.onCreate`, and a hidden "Register as test device" toggle in Advanced settings. +514/−0 across 15 files.

I verified the integration points against the codebase: `arrivals_reminders_api_endpoint` is `/api/v2/regions/`, so the URL build exactly mirrors the alarms path in `TripInfoRepository.requestAlarm`; `PreferencesRepository` has every overload used (`observeString`/`observeBoolean`/`getLong`, res-id keyed); `region.sidecarBaseUrl` and `FirebaseMessagingManager.userPushId()` exist as used. The `plainRetrofit` provider mirrors the five existing sidecar clients.

## Strengths

- **The decision extraction is exactly the house style** — pure, Android-free, exhaustively JVM-tested (all 4 action types, 8 transitions including locale/test-flag refresh vs. token/region re-register). This matches the project's "extract branchy logic and unit-test it" standard.
- **The commit-sentinel persistence design is sound.** I checked `DefaultPreferencesRepository.put`: cache updates are synchronous in call order, disk persists async per-key. Torn writes (crash between `persist`'s writes) read back as "no record" because the token is written last, and re-registration is idempotent. The design comment matches the actual prefs semantics.
- **Rate-limit awareness**: `decide()`'s no-op path + persisted last-registration correctly avoid redundant POSTs against the 30 req/min/IP limit; DELETE treating 404 as success is the right call.
- **Coroutine hygiene**: `runCatchingCancellable` used in both network paths, per the #1908 rule. Mutex-serialized `sync()` prevents interleaved reconciliations.
- The pre-existing-heuristic and DELETE-shape ambiguities are honestly flagged rather than silently guessed — consistent with the "no unsanctioned heuristics" policy.

## Issues

**1. In-app notification-permission grant doesn't trigger registration** (`PushRegistrationManager.start`). The app requests `POST_NOTIFICATIONS` at runtime in the trip-details flow (`DestinationReminder.kt:177`, `TripDestinations.kt:179`). The system permission dialog pauses but does not *stop* the activity, so no process-lifecycle `ON_STOP`→`ON_START` cycle occurs and the grant isn't picked up until the next app foreground/launch. That's the highest-value opt-in moment of the ones this feature exists for. Fix is small: call `resync()` from the permission-result callbacks (the manager is injectable, or via the entry point).

**2. `Reregister` can orphan the previous registration** (`PushRegistrationManager.sync`). In the `Reregister` arm, `unregister(previous)`'s result is discarded; if both the DELETE and the subsequent POST fail (e.g. offline during a region switch), `clearLast()` forgets `previous`. The next sync sees `last == null` → plain `Register`, and the old region's registration is never DELETEd — it keeps pushing that region's alerts to this still-valid token until the 180-day age-out. Suggested shape:

```kotlin
is PushRegistrationAction.Reregister -> {
    val cleaned = unregister(action.previous)
    if (register(action.target)) persist(action.target)
    else if (cleaned) clearLast()   // else keep `previous` so the next sync retries the full Reregister
}
```

**3. DELETE token conveyance** — already flagged in the PR body and at the call site; agreeing this is the one hard pre-merge blocker. While verifying against the OBACloud spec, also confirm **the boolean form encoding**: `@Field("test_device") testDevice: Boolean` serializes as the literal strings `"true"`/`"false"` — fine for a Rails-style parser, but worth confirming the server doesn't expect `1`/`0`. Same verification session, zero extra cost.

## Minor / nits

- `decide` is a very generic name for a package-level public function; call sites outside the file lose context. `decidePushRegistration(...)`, or hanging it off a `PushRegistrationDecision` object, would read better — but it's package-local in practice, so take or leave.
- No retry/backoff on transient failure: a failed `register` waits for the next trigger (foreground/pref change). Given the triggers fire on every app foreground, that's an acceptable eventual-consistency story — just noting it's deliberate-looking and fine.
- `loadLast()` defaults `regionId` to `-1` and `locale` to `""` for missing slots; both are unreachable given the sentinel ordering, and harmlessly self-heal (inequality → refresh) if somehow hit. OK as is.
- `sync()` is `suspend` + public while `resync()` is the fire-and-forget wrapper — the split is clean, but nothing outside the class calls `sync()` directly; it could be `private` with `resync()` as the only public trigger. (`start()` uses it from the collector, so `private` works.)

## Test coverage

The pure decision core is fully covered, and `SettingsUiStateTest` was updated for the new snapshot field. `PushRegistrationManager` itself is untested — the blocker is `NotificationManagerCompat.from(context)` buried in `currentTarget()`; every other dependency is already constructor-injected. If you want `sync()`'s apply/persist semantics under JVM test (issue 2 above is exactly the kind of path that would have been caught), injecting a `notificationsEnabled: () -> Boolean` supplier would get you there cheaply. Not a merge blocker given the extraction that was done.

## Security

Nothing concerning client-side: the token travels over TLS to the region's sidecar, and "possession of token = ownership" on an unauthenticated endpoint is the server's documented model. The token was already persisted in prefs pre-PR. The `push_reg_*` bookkeeping adds no new sensitive data beyond the token slot that already existed.

## Verdict

Well-built and convention-faithful; the decision/manager split is the right architecture. Before merge: confirm the DELETE shape (author-flagged), and I'd recommend fixing **#1** (in-app permission grant is a missed registration trigger for the feature's primary audience) and **#2** (orphaned registration on double failure) — both are small, contained changes. On-device verification of the actual request shapes, as the PR body notes, remains pending.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
```diff
@@ -0,0 +1,225 @@
...
+        drainPendingDelete()
+
+        // Notifications being off is the ONLY definitive "this device should not be registered" signal.
+        // A missing region or FCM token just means we can't tell yet — both resolve asynchronously, and
+        // the region flow is explicitly "briefly null at cold start" (see RegionRepository) — so treating
+        // that as an opt-out would fire a spurious DELETE on launch and re-POST seconds later, burning
+        // two requests against a rate-limited endpoint and leaving a window with no registration at all.
+        val target = if (notificationsEnabled()) {
+            // Inputs still settling — bail rather than decide; a later trigger re-runs us.
+            buildTarget() ?: return@withLock
```
aaronbrethorst: **🔴 Confirmed correctness bug — stale registration leaks when the new region has no sidecar host**

When notifications are enabled but `buildTarget()` returns `null` because the current region's `sidecarBaseUrl` is blank, `sync()` bails via `?: return@withLock` here — *before* `decidePushRegistration` runs. A `null` target on this path is treated as "inputs not resolved yet", not opt-out, so a registration stored for a *different* region is never `Unregister`ed.

Scenario: rider registers in region A (has a sidecar), then switches to region B whose `Region.sidecarBaseUrl` is blank (the field defaults to `""`). On next foreground, `notificationsEnabled()` is true, `buildTarget()` returns null at the `takeIf { it.isNotBlank() }` guard (~line 210), and `sync()` returns. The region-A registration is never deleted — the device keeps getting region A's alerts and no region-B alerts until the server's 180-day prune.

Consider distinguishing "no sidecar for this region" (which should still reconcile away a stale registration) from "inputs genuinely not ready yet."

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationStore.kt
```diff
@@ -0,0 +1,157 @@
...
+        // The commit sentinel, written last — mirroring [record].
+        prefs.setString(KEY_PENDING_DEL_TOKEN, registration.token)
+    }
+
+    /** The registration awaiting a retried DELETE, or null if none is queued. */
+    fun pendingDelete(): PushRegistration? {
+        val base = prefs.getString(KEY_PENDING_DEL_BASE, null) ?: return null
+        val token = prefs.getString(KEY_PENDING_DEL_TOKEN, null) ?: return null
+        return PushRegistration(
+            regionId = prefs.getLong(KEY_PENDING_DEL_REGION_ID, -1L),
```
aaronbrethorst: **🔴 Confirmed correctness bug — torn pending-delete write DELETEs against region `-1` and treats the 404 as success**

`pendingDelete()` reconstructs `regionId` with a default of `-1L`. `queuePendingDelete` issues three *independent* async DataStore writes, with the token written last as the presence sentinel — but the sentinel guards record *presence*, not `regionId` correctness. If the process is killed after `KEY_PENDING_DEL_TOKEN` flushes but before `KEY_PENDING_DEL_REGION_ID`, the next launch reads `regionId = -1L`, `drainPendingDelete` issues `DELETE …/regions/-1/push_registrations?token=…`, the server returns 404, the client counts 404 as success, and `clearPendingDelete()` runs.

Net effect: the genuinely-stale registration is never deleted and keeps pushing to the still-valid token until the 180-day prune. Consider persisting the pending-delete record atomically (single serialized value) rather than as three separate keys, and/or refusing to drain when `regionId < 0`.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationClient.kt
```diff
@@ -0,0 +1,144 @@
...
+        val response = service.register(
+            url = registrationUrl(target),
+            token = target.token,
+            locale = target.locale,
+            testDevice = target.testDevice,
+            // Required by the server for a test device, omitted otherwise (see the service KDoc).
+            description = target.description
+        )
+        val registered = response.isSuccessful
+        if (!registered) reportHttpFailure("register", target, response)
```
aaronbrethorst, line 82: **🟠 Confirmed correctness concern — Crashlytics retry storm on rate-limited (429) responses**

`reportHttpFailure` records *every* non-2xx `register` response to Crashlytics, gating only on `response.isSuccessful` with no status-code carve-out — contrast `unregister()` (line ~90), which does carve out 404. The unauthenticated endpoint is rate-limited (30 req/min/IP). On a shared-NAT network (public wifi, carrier CGNAT) a 429 gets reported like a systematic 422. Because a failed `register` persists nothing, the next foreground retries and reports again, flooding the error channel with transient throttling noise and drowning the genuine systematic-rejection signal the reporting was meant to surface.

Consider treating 429 (and other transient statuses) as non-reportable, mirroring the 404 carve-out already in `unregister()`.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
```diff
@@ -0,0 +1,225 @@
...
+        } else {
+            null
+        }
+
+        when (val action = decidePushRegistration(target, store.last(), store.sinceLastSent())) {
+            PushRegistrationAction.NoOp -> Unit
+            is PushRegistrationAction.Register ->
+                if (client.register(action.target)) store.record(action.target)
+            is PushRegistrationAction.Unregister ->
```
aaronbrethorst, line 154: **🟠 Correctness / contract divergence — Android DELETEs on OS-level opt-out; iOS never does**

When notifications are disabled at the OS level, `target` becomes `null`, `decidePushRegistration` returns `Unregister(last)`, and the client DELETEs `…/push_registrations?token={token}` (plus `store.clear()`). Issue #1957 explicitly says an OS-level disable *doesn't* need a DELETE — "FCM bounces feed back to the server and it cleans up unregistered tokens itself" — and the iOS `PushRegistrationManager` only ever POSTs, never DELETEs.

Beyond the divergence: if the server couples trip-alarm delivery to the `push_registration` row for that token, this Android-only DELETE can silently drop the delivery target of an already-scheduled alarm when a rider toggles notifications off. Recommend matching the iOS contract — don't DELETE on opt-out; let FCM-bounce/server prune handle it.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
```diff
@@ -0,0 +1,225 @@
...
+     * cheap guard on the wire boundary itself, so no path can put an over-long `description` on a
+     * request the server would reject outright.
+     */
+    private fun buildTarget(): PushRegistration? {
+        val region = regionRepository.region.value ?: return null
+        val base = region.sidecarBaseUrl?.takeIf { it.isNotBlank() } ?: return null
+        val token = firebaseMessagingManager.userPushId().takeIf { it.isNotEmpty() } ?: return null
+        val description = prefs.getString(R.string.preference_key_push_test_device_name, null)
+            ?.trim()
+            ?.take(PUSH_DESCRIPTION_MAX_LENGTH)
```
aaronbrethorst: **🟡 Plausible correctness — `take(255)` counts UTF-16 code units and can split a surrogate pair**

The test-device name cap uses `String.take(PUSH_DESCRIPTION_MAX_LENGTH)`, which counts UTF-16 code units. A description near the boundary containing emoji / astral characters can be truncated mid-surrogate-pair, leaving a lone surrogate (mangling the last char), and the code-unit count may not match the server's char/byte limit.

If the server then 422-rejects, nothing is persisted, so `buildTarget` rebuilds the identical doomed request and re-POSTs on every foreground — the exact repeat loop the cap was meant to prevent, plus a Crashlytics report each time. Consider capping on Unicode code points / grapheme boundaries, and confirming whether the server limit is chars or bytes.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationClient.kt
```diff
@@ -0,0 +1,144 @@
...
+     */
+    suspend fun unregister(previous: PushRegistration): Boolean = runCatchingCancellable {
+        val response = service.unregister(url = registrationUrl(previous), token = previous.token)
+        val removed = response.isSuccessful || response.code() == HttpURLConnection.HTTP_NOT_FOUND
+        if (!removed) reportHttpFailure("unregister", previous, response)
+        removed
+    }.onFailure { logTransportFailure("unregister", previous, it) }.getOrDefault(false)
+
+    /** `{sidecarBaseUrl}/api/v2/regions/{regionId}/push_registrations` — mirrors the alarms URL build. */
+    private fun registrationUrl(registration: PushRegistration): String = registration.sidecarBaseUrl +
```
aaronbrethorst: **🔵 Cleanup (DRY) — sidecar registration-URL assembly duplicates `TripInfoRepository.requestAlarm`**

`registrationUrl()` hand-concatenates `sidecarBaseUrl + arrivals_reminders_api_endpoint + regionId + "/push_registrations"`, which is the same sidecar-path shape `TripInfoRepository.requestAlarm` (~line 263) builds for `"/alarms"`. Both independently re-derive the `/api/v2/regions/{id}` path, so a future change to the sidecar path scheme must be found and edited in two places. Consider extracting one helper (e.g. `Region.sidecarUrl(endpoint, suffix)`) as the single source of truth for the sidecar URL shape.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
```diff
@@ -0,0 +1,225 @@
...
+     * calls can fail independently, and all four quadrants have to leave the device recoverable:
+     *
+     * - DELETE ok   + POST ok   → record(target): old gone, new on record.
+     * - DELETE ok   + POST fail → clear(): the server holds nothing, so the next sync re-Registers.
+     * - DELETE fail + POST fail → keep `previous`: the next sync retries the whole Reregister.
+     * - DELETE fail + POST ok   → record(target) AND queue `previous` for a retried DELETE, rather than
+     *   dropping it — otherwise the old region keeps pushing to a still-valid token until the server's
+     *   180-day prune.
+     */
+    private suspend fun applyReregister(action: PushRegistrationAction.Reregister) {
```
aaronbrethorst, line 167: **🔵 Cleanup / over-engineering vs. the ticket + iOS — the whole Reregister + persisted pending-delete machinery**

Issue #1957 marks deleting the old region's registration as *optional* ("optionally DELETE from the old one"), and iOS `PushRegistrationManager` never DELETEs at all — it re-POSTs and relies on the server's 180-day prune + FCM-bounce cleanup. This branch instead adds a full `Reregister` action, `applyReregister`'s 4-quadrant handling, three persisted pref slots (`KEY_PENDING_DEL_*`), the `queuePendingDelete`/`pendingDelete`/`clearPendingDelete`/`drainPendingDelete` cycle drained on every `sync`, and a large slice of the manager test — all to robustly recover the one DELETE-fail/POST-ok quadrant of a behavior the sister platform chose not to implement.

That's permanent maintenance + test surface for an optional cleanup the server already performs, and it silently diverges from iOS. Worth confirming with a human whether the DELETE-on-region-change path is wanted at all, or whether matching iOS (POST-only) is the simpler, intended contract.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
bmander: Fixed in 20a051bf, and refined in 57c5d395. The null-target bail is gone: `deriveDesiredRegistration` now returns a sealed `DesiredRegistration` in which "the region hasn't loaded yet" (`Unresolved` — freezes reconciliation) and "this region has no sidecar" (`NoSidecar` — a resolved fact) are distinct types, so a blank `sidecarBaseUrl` can no longer masquerade as not-ready. `decidePushRegistration` only accepts resolved states, and `NoSidecar` with a record on file yields `Unregister` — your exact scenario (region A → sidecar-less region B) is pinned by the `switching to a region without a sidecar unregisters the stale registration` test, which asserts the DELETE targets region A's host.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationStore.kt
bmander: Fixed at the root in 8a8f67a1 by implementing #1978's option 1: `PreferencesRepository` gained an atomic batch `edit {}` (one cache swap + one `dataStore.edit`, so a multi-slot record persists whole or not at all across process death), and every multi-slot store write goes through it. A record whose presence marker exists but whose `regionId` didn't flush can no longer exist, so no `regionId < 0` drain guard was needed — nothing shipped the old layout, so no legacy torn state exists in the wild. A bonus fix fell out: `record()` folds the same-endpoint pending-delete drop into the same commit, closing the sibling crash window where a fresh registration and a stale queued DELETE for the same endpoint could both survive and delete the row just POSTed.

(Since then, the pending-delete record itself was removed entirely per the over-engineering thread below — but the atomicity fix stands for the six-slot live record, and closes #1978.)

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationClient.kt
bmander: Fixed in 2812ca26. `reportHttpFailure` still logs every non-2xx locally, but a new `isTransient()` (429, plus the 5xx range — a server outage has the same fleet-wide flood shape) skips the Crashlytics report, mirroring the transport-failure split that already existed. Contract-shaped 4xx like the 422 still report. The boolean return is untouched, so the retry-on-next-foreground behavior is identical — tests pin that a 429 logs without reporting and succeeds on the next sync once the throttle lifts.

One deliberate trade worth your eyes: a *chronic* 5xx (server-side register handler broken indefinitely) is now invisible to Crashlytics too. We judged that acceptable since a fleet-wide 5xx shows up in server-side monitoring, but if you'd rather catch it client-side, a report-after-N-consecutive-failures counter would cover it.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
bmander: Fixed in 57c5d395, verified against all three sources: #1957's "OS-level disable doesn't need a DELETE", the iOS manager (no DELETE call exists at all), and the iOS docs ("OS-level permission revocation does *not* require a DELETE"). One nuance: the old `DesiredRegistration.None` conflated two cases with opposite correct behaviors, so it's split rather than removed — `OptedOut` (notifications off) is now a pure `NoOp` that keeps both the server row and the local record (so re-enabling inside the refresh window reads as "unchanged", with no spurious re-POST), while `NoSidecar` still `Unregister`s, since your docs sanction exactly that old-row DELETE on a region change. The trip-alarm delivery-target risk you raised is recorded in `OptedOut`'s KDoc as part of the rationale.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
bmander: Fixed in 6381afa5. Both cap sites (the settings persistence boundary and the wire-boundary clamp) now share `truncatedToDescriptionCap()`, which keeps the 255-UTF-16-unit budget but drops a straddling pair whole — a lone surrogate can no longer be stored or sent. Tests cover the straddling and exact-fit boundary cases.

On your chars-vs-bytes question: we couldn't settle it at the source — the OBACloud repo is private — so the cap deliberately stays at 255 UTF-16 units as the *strictest* plausible reading of the documented "≤255 chars" (≤N units implies ≤N code points, so it's legal under either counting; only a byte-count limit could differ, and nothing in the docs suggests bytes). Since you have OBACloud access: if the Rails validation counts characters/code points, this could safely loosen to 255 code points later — purely a permissiveness change. Happy to file that as a follow-up if you can confirm.

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationClient.kt
bmander: Fixed in 9b6c9733 (polished in 7efe9f5b): both sites now call `sidecarRegionUrl()` in `api/contract`, next to the two sidecar Retrofit interfaces it serves. Output is byte-identical (the test pinning the exact registration URL passes untouched). Two deviations from the `Region.sidecarUrl()` sketch, both forced: it's not a `Region` method because the push client addresses DELETEs from its *recorded* registration, whose region may no longer be current; and the `/api/v2/regions/` segment stays a parameter because it lives in the `arrivals_reminders_api_endpoint` string resource. Open question for you: no flavor overrides that resource today — if the override affordance isn't wanted, the segment could be hardcoded in the helper and the resource deleted, which would also let the client drop its constructor param. (Also noted in the helper's KDoc: the v1 sidecar endpoints — weather, surveys — use a separate `regionID`-placeholder idiom; this helper owns only the v2 shape.)

onebusaway-android/src/main/java/org/onebusaway/android/push/PushRegistrationManager.kt
bmander: Confirmed with a human as requested, and resolved down the middle in 6d8e8e0f (+7efe9f5b): the **Reregister DELETE stays, but it's now best-effort** — the persisted pending-delete queue (three pref slots, queue/drain on every sync, and its test surface, net −110 lines) is gone. The reasoning: on a *region change* the old row holds a live token, so nothing bounces and the rider who moved keeps getting the old region's alerts until the 180-day prune — your own iOS docs flag that consequence and sanction the DELETE, so we kept it. The queue, though, only insured the DELETE-fails/POST-lands corner, and its failure mode without the queue (old row lingers until prune) is precisely iOS's baseline for *every* region change — disproportionate machinery for that corner, exactly as you argued. `applyReregister` is now the plain two-step composition of the Unregister/Register rules, with the best-effort choice documented on the `Reregister` action. If you'd rather go full POST-only (drop the region-change DELETE too), that's now a five-line change — but we'd argue the wrong-region-alerts window makes the best-effort DELETE worth keeping.

PR #1964, opened 2026-07-20
Stop route focus: project the banner headsign through the shared row resolver

bmander:
## Why

This is the "make the bug impossible" follow-up to **#1962**, which fixed the missing-headsign symptom but left the anti-pattern in place: the focus banner still resolved the headsign on its own, separately from every other surface.

The bug was possible because two entry points into a focused stop's route sub-focus each independently decided what display data to carry, and the banner rendered whatever was carried — so the display quality depended on which tap you used.

## The key realization

The shared identity I'd have had to *invent* already exists:

- `routeRowKey(routeId, directionId, headsign)` is the stable row key used by grouping, the LazyColumn, **and the drawer's selection highlight**.
- `resolveSelectedRouteGroupKey` is the one resolver, and it already handles the map-badge case (a single-row-by-route fallback for when map and arrivals direction labels differ).

The drawer highlight *already* resolved a map route-label selection correctly through this. The focus banner was the **only** surface that bypassed the resolver and read the raw stored `originHeadsign` — which the map-label path never set. That was the whole bug.

## The change

- Add `resolveSelectedRouteGroup` (returns the resolved `RouteRowGroup`, wrapping the existing key resolver).
- The banner reads its headsign from `resolveSelectedRouteGroup(routeGroups, selection.selectedArrivalRowKey(), originLeg.routeId)` — the exact identity the drawer already feeds the resolver.
- `originHeadsign` stays, but purely as an **identity-key component** (disambiguating the legacy null-`directionId` row), never rendered. Its doc now says so.
- Removes the per-entry-point `subordinateRouteHeadsign` helper #1962 added.

Now every stop-focus surface projects display from **one** identity→row resolution. The two entry points supply only row *identity* (`StopRouteSelection`); the shown headsign is always read back from the resolved arrivals row. No entry point carries a display headsign, so they can't disagree — the class of bug is gone by construction, and it generalizes to any future surface (they resolve through the same function or they don't render the row at all).

## Stacking

Branched on top of **#1962** (`fix/stop-route-focus-headsign`). Review/merge that first, or merge this alone as the complete version — its diff net-removes the helper #1962 introduced. Net −28 lines.

## Testing

- Compiles clean under the CI gate (`-PwarningsAsErrors=true`) on both `obaGoogle` and `obaMaplibre`.
- `RouteRowGroupingTest` extended: the resolved group supplies the row headsign for a map-badge `(route, direction)` key (no headsign string), and resolves to null when the route is ambiguous.
- Not yet re-exercised on-device.


bmander:
Folded into #1962 — that PR now contains the single 'done right' commit (the intermediate per-entry-point helper this PR removed never appears in the squashed history). Deleting this branch.

Issue #986, opened 2019-06-12
"Your item has been successfully reported" dialog message translation


cbergy:
**Summary:**

Trying to modify/translate the message shown in a dialog when successfully sending a problem report, the text shown seems not to be translatable or editable.

**Steps to reproduce:** 

Send a problem report about stop or arrival time problem. 

If report is successfully sent, a dialog is shown with the message "Your item has been successfully reported" and an "Ok" button.

**Expected behavior:** 

The message should be editable and/or translatable

**Observed behavior:** 

Looking for the string in the code in order to edit or translate it, I was not able to find it.

**Device and Android version:** 

Possibly any android versions and different devices.

**Screenshots:** 

![OBA_issue_reported](https://user-images.githubusercontent.com/39773271/59325373-15a35000-8ce3-11e9-8b54-e227fadb5b04.jpg)

barbeau:
Thanks @cbergy!  What region are you located in?

I believe this string comes from the Open311 library we're using:
https://github.com/CUTR-at-USF/open311-client

cbergy:
@barbeau: Many thanks. I checked the library and i can confirm the message is hardcoded in Open311 library. Actually I was able to change it modifying the Open311 library and loading the new library in Onebusaway-Android. Obviously it's still hardcoded.

I'm translating the app in Spanish.

PR #1994, opened 2026-07-23
Let a stop be starred before its arrivals load (#684)

### LINKED ISSUE #684: Can't star a stop until first arrival info has been loaded
**Summary:** 

You can't star a stop while things are initializing

**Steps to reproduce:** 

I've seen this previously, although now I can't reproduce it.  IIRC, the steps were something like:

1. Use a slow Internet connection
1. Go to "Nearby" mode and tap on a stop
1. Immediately try to star a stop in the sliding panel header (tap the star) before the arrival info loads

**Expected behavior:** 

Allow me to star the stop

**Observed behavior:** 

When I tap the star, nothing happens.

I've seen this previously on stock LG G4 w/ Android 6.0.  User with Galaxy J1 (j1xlteatt) w/ Android 6.0 also reported this via Google Play review.

bmander:
Closes #684.

## Problem

On the map focus banner, the stop **star was disabled until the first arrivals response loaded** (`favoriteEnabled = arrivalsContent != null` in `HomeScreen.kt`). On a slow connection this left the star unusable for a while after tapping a stop.

Root cause: favoriting a stop does `UPDATE stops SET favorite WHERE _id = …` (`StopDao.setFavorite`), which is a no-op until that row exists. A focused stop lives only in the map’s `cached_stops` cache — its user-state `stops` row is created when arrivals load. So the star was gated on arrivals purely to guarantee the row existed.

## Fix

Mirrors the existing route-favorite path (`RouteFavoritesRepository`), which already "ensures the row" before flipping the flag:

- **`StopDao.setFavoriteEnsuringRow(identity, favorite)`** — a `@Transaction` that inserts the stop’s identity row (id/name/code/coords + current region) when absent, otherwise just flips the flag. An existing row’s `user_name` / `use_count` / `access_time` are never clobbered, and a later arrivals load merges onto the row via `markStopUsed()` (which leaves `favorite` untouched), so an early star sticks.
- **`StopFavoritesRepository`** (new) — import-gated `favoriteStopIds()` plus a `setFavorite(...)` that ensures the row then flips, parallel to `RouteFavoritesRepository`.
- **`FocusBannerViewModel`** — now exposes `favoriteStopIds` and `toggleStopFavorite(stop)`, reusing the same optimistic-overlay pattern the route star already uses (instant fill on tap).
- **`HomeScreen`** — the banner star reads `isFavorite` from the favorites set and is **always enabled**; the toggle routes to `focusBannerViewModel.toggleStopFavorite(...)`.

The standalone full-screen arrivals star is unchanged (still `ArrivalsViewModel.toggleFavorite`); both paths write `stops.favorite`, so display stays consistent.

## Verification

- `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin -PwarningsAsErrors=true` → BUILD SUCCESSFUL (strict CI gate).
- Not yet exercised on-device; happy to add an instrumented `StopDao` test for `setFavoriteEnsuringRow` (row-absent insert vs row-present flip, favorite-preservation on merge) if desired.

## Summary by CodeRabbit

* **New Features**
  * Added starring/unstarring for the focused stop view, with immediate UI updates.
  * Stop favorites now persist reliably, including when imports run, without overwriting existing stop details.
  * Favorite controls stay disabled until stop favorites are ready to prevent inconsistent toggles.
* **Bug Fixes**
  * Prevented later arrivals/import data from undoing user favorite changes.
* **Tests**
  * Added instrumented coverage for favorite insert/update and early-star behavior.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/database/oba/StopDaoMergeTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/database/oba/StopDaoMergeTest.kt
@@ -68,6 +68,91 @@ class StopDaoMergeTest {
     }
 
     @Test
+    fun setFavoriteEnsuringRow_insertsIdentityRowWhenAbsent() = runBlocking {
+        // A stop focused on the map (only in cached_stops) has no user-state row yet — starring it
+        // must create the row with the flag already set (#684).
+        assertEquals(null, dao.getStop("s1"))
+
+        dao.setFavoriteEnsuringRow(
+            StopRecord(
+                id = "s1",
+                code = "100",
+                name = "Main St",
+                direction = "",
+                useCount = 0,
+                latitude = 47.0,
+                longitude = -122.0,
+                regionId = 1L
+            ),
+            favorite = 1
+        )
+
+        val s = dao.getStop("s1")!!
+        assertEquals(1, s.favorite)
+        assertEquals("Main St", s.name)
+        assertEquals("100", s.code)
+        assertEquals(1L, s.regionId)
+    }
+
+    @Test
+    fun setFavoriteEnsuringRow_onlyFlipsFlagWhenRowExists() = runBlocking {
+        // Existing row carries user state that a re-star must not clobber.
+        dao.markStopUsed("s1", "100", "Main St", "N", 47.0, -122.0, regionId = 1L, now = 100)
+        dao.upsert(dao.getStop("s1")!!.copy(userName = "Home"))
+
+        // Ensure-row is handed the focused-stop identity (blank direction, no user state), but the row
+        // exists, so only the flag flips — user_name / use_count / access_time survive.
+        dao.setFavoriteEnsuringRow(
+            StopRecord(
+                id = "s1",
+                code = "100",
+                name = "Main St",
+                direction = "",
+                useCount = 0,
+                latitude = 47.0,
+                longitude = -122.0,
+                regionId = 1L
+            ),
+            favorite = 1
+        )
+
+        val s = dao.getStop("s1")!!
+        assertEquals(1, s.favorite)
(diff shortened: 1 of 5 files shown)


bmander:
Rebased onto `main` and addressed the review feedback. The branch is now a single, strictly-scoped commit.

**Merge conflict** (`HomeScreen.kt`): reconciled with #1987 — kept the new `stopCode = arrivalsContent?.stopCode ?: currentFocus.stop.code` alongside the #684 star wiring.

**CodeRabbit — import-gate readiness** (`FocusBannerViewModel`/`HomeScreen`): the banner star no longer treats the pre-import `emptySet()` as authoritative. `FocusBannerViewModel` now tracks the persisted set as `Set<String>?` (null until the store's first emission) and exposes `stopFavoritesReady`. The star's `favoriteEnabled` is gated on that readiness (not on arrivals), and `toggleStopFavorite` ignores taps until ready — so a legacy-starred stop is never shown unstarred (and thus un-unstarrable) during the one-time legacy-import window.

**CodeRabbit — reconcile optimistic state on persistence failure** (`FocusBannerViewModel.toggleStopFavorite`): the write is now wrapped in `runCatchingCancellable` (rethrows `CancellationException` per the coroutine guideline); on failure the optimistic override is rolled back, but only if it still matches this request's value, so a newer tap isn't clobbered.

**Out-of-scope `InfrastructureIssueScreen`/`strings.xml` changes**: these turned out to be the #986 localization that already landed on `main` via #1991. Rebasing onto current `main` drops them entirely — the branch now touches only the four stop-favorite files.

Verified `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin -PwarningsAsErrors=true` → BUILD SUCCESSFUL (strict CI gate).

Issue #2020, opened 2026-07-25
Crash: NPE in VehicleBitmaps.vehicleType when a vehicle's activeTripId is missing from trips-for-route references


bmander:
Tapping a route on the map crashes the app instantly (`FATAL EXCEPTION: main`) when the vehicle layer draws a vehicle whose `activeTripId` isn't in the trips-for-route response's `references` pool.

Captured on-device (Pixel 7 Pro, Android 16, `v153 (26.1.0)`, `obaGoogleDebug`, foreground, screen on) from `dumpsys dropbox --print data_app_crash`.

## Stack trace

```
2026-07-24 20:45:49 data_app_crash
Process: com.joulespersecond.seattlebusbot   PID: 5877   Foreground: Yes
Build: google/cheetah/cheetah:16/CP1A.260405.005/15001963:user/release-keys

java.lang.NullPointerException
	at org.onebusaway.android.map.render.VehicleBitmaps.vehicleType(VehicleBitmaps.kt:122)
	at org.onebusaway.android.map.render.VehicleBitmaps.iconKey(VehicleBitmaps.kt:112)
	at org.onebusaway.android.map.googlemapsv2.GoogleMapRenderer.vehicleIcon(GoogleMapRenderer.kt:666)
	at org.onebusaway.android.map.googlemapsv2.GoogleMapRenderer.reconcileVehicleMarkers(GoogleMapRenderer.kt:646)
	at org.onebusaway.android.map.googlemapsv2.GoogleMapRenderer.reconcileVehicles(GoogleMapRenderer.kt:528)
	at org.onebusaway.android.map.googlemapsv2.compose.GoogleComposeAdapter$Content$5$1$1.emit(GoogleComposeAdapter.kt:218)
	at kotlinx.coroutines.flow.StateFlowImpl.collect(StateFlow.kt:401)
	at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:79)
	...
```

## Cause

`VehicleBitmaps.vehicleType` resolves the vehicle's route type through two `!!`s on one line:

```kotlin
// VehicleBitmaps.kt:119-124
private fun vehicleType(vehicle: VehicleMarker, response: RouteTrips): Int {
    val status = vehicle.status
    // Non-null chain matches the former Java (it NPE'd on a missing trip/route too).
    val type = response.route(response.trip(status.activeTripId)!!.routeId)!!.type
[...]

PR #2022, opened 2026-07-25
Give a fixed-region flavor with no Open311 endpoint an empty array, not null

bmander:
`getRegionFromBuildFlavor` handed `Region` a **null** Open311 array whenever the flavor configured no `FIXED_REGION_OPEN311_BASE_URL`:

```java
} else {
  open311Array = null;
}
```

`Region.open311Servers` is a non-null Kotlin property (defaulting to `emptyArray()`), so that crashes the constructor and kills the app at launch. Having no Open311 endpoint is ordinary configuration — "this agency doesn't run one" — and should yield a region with no Open311 servers, not no region at all.

## Latent, but a trap laid directly under rebranders

Not currently reachable on a shipped build: `agencyY` is the only flavor with `USE_FIXED_REGION=true`, and it happens to set an Open311 URL. But `kiedybus` already has a **null** one alongside a non-null name, so flipping its flag would hit this on the first launch — and it's precisely the path `docs/REBRANDING.md` sends a new white-label brand down. Every `FIXED_REGION_*` field is optional-looking in the flavor file; only two of them are load-bearing for `Region`'s non-null half.

## The diagnostics are the other half

R8 rewrites Kotlin's non-null parameter check into `Object.getClass()`, so a release build (`isMinifyEnabled = true`) reports:

```
java.lang.NullPointerException: Attempt to invoke virtual method
'java.lang.Class java.lang.Object.getClass()' on a null object reference
    at org.onebusaway.android.region.Region.<init>(Unknown Source:6)
    at org.onebusaway.android.util.RegionUtils.getRegionFromBuildFlavor(...)
```

— naming **no field**. You cannot tell from that which `buildConfigField` you forgot. The region name is the other argument that is non-null in `Region` while being literally `null` in the flavors that don't use a fixed region, so it now gets an explicit check that says which one is missing:

```java
Objects.requireNonNull(
    BuildConfig.FIXED_REGION_NAME,
    "FIXED_REGION_NAME must be set in the build flavor when USE_FIXED_REGION is true");
```

Failing loudly at the source, rather than letting a missing field surface later as an anonymous NPE.

## Tests

Extracting `open311ServersFrom` is what makes the broken branch testable at all — it takes its three values as parameters instead of reading `BuildConfig`, which is a per-variant constant a unit test can't vary. New `RegionUtilsOpen311Test` (JVM, no Robolectric) covers the empty case, the one-server case, and the null-jurisdiction/key case that `agencyY` actually ships. `tests="3" failures="0" errors="0"`.

To be straight about coverage: the empty-array test can't be run against the pre-fix code, because the helper it calls didn't exist then — the old `null` was inline. What it does pin is the property that matters, asserting the result actually constructs a `Region`, which is what used to blow up.

Also ran `assembleAgencyYGoogleDebug`, since `agencyY` is the only flavor that executes this code. `spotlessCheck` and `testObaGoogleDebugUnitTest` green under `-PwarningsAsErrors=true`.

## Scope

Deliberately left alone: `FIXED_REGION_OBA_BASE_URL` is nullable on `Region`, so a fixed region without one doesn't crash — it just quietly won't work. Validating that too is defensible but it's a behavior change for existing flavors rather than a crash fix, so it isn't in here.

Found while investigating an unexplained crash report; unrelated to #2020.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Open311 handling so missing or blank base URLs result in no servers instead of null values.
  - Tightened fixed-region configuration to require a non-null region name, avoiding unclear startup issues.
  - Preserved Open311 jurisdiction and API key details when configured.

- **Tests**
  - Expanded regression tests to cover `null`, empty, and whitespace-only Open311 base URLs, plus jurisdiction/key carry-through behavior.

- **Documentation**
  - Updated fixed vs. multi-region configuration guidance with clearer warnings for single-region flavors.

### DIFF
--- a/docs/REBRANDING.md
+++ b/docs/REBRANDING.md
@@ -294,6 +294,16 @@ Users can change the sorting style using the "Sort by" button regardless of the 
 * `false` - The app works across various regions defined in the Regions API (recommended for most brands)
 * `true` - The app is fixed to the region information provided in the flavor configuration
 
+**When `USE_FIXED_REGION` is `true`, `FIXED_REGION_NAME` is required** - it is the one
+`FIXED_REGION_*` field the app cannot start without, and leaving it `null` (the value every
+multi-region flavor uses) crashes on the first launch. Every other `FIXED_REGION_*` field is
+optional in the sense that it will not crash: `FIXED_REGION_OBA_BASE_URL` in particular is
+*silently* accepted as `null`, which builds and installs fine and then simply cannot reach a
+server — so set it too, and set the bounds, or the fixed region will not work.
+
+When `USE_FIXED_REGION` is `false`, the `FIXED_REGION_*` fields are unread; leave them `null` as
+`agencyX.gradle` does.
+
 ### Geocoding Provider
 
 `USE_PELIAS_GEOCODING` - Controls which service is used for searching origins and destinations in trip planning:
--- a/onebusaway-android/flavors/README.md
+++ b/onebusaway-android/flavors/README.md
@@ -131,10 +131,19 @@ buildConfigField "boolean", "USE_FIXED_REGION", "false"
 
 ```groovy
 buildConfigField "boolean", "USE_FIXED_REGION", "true"
+// Required: the app cannot start without a region name.
 buildConfigField "String", "FIXED_REGION_NAME", "\"Your Region\""
+// Not required to launch, but the fixed region can't reach a server without it.
 buildConfigField "String", "FIXED_REGION_OBA_BASE_URL", "\"https://api.example.com/api\""
 // ... configure all region bounds and features
 ```
+
+`FIXED_REGION_NAME` is the only one of these the app will refuse to launch without — it's non-null
+on the region model, so leaving it `null` throws from `RegionUtils.getRegionFromBuildFlavor()` on
+first launch. The rest degrade quietly instead of crashing, which is worse to debug: a null
+`FIXED_REGION_OBA_BASE_URL` installs cleanly and then can't load anything. `FIXED_REGION_OPEN311_*`
+is genuinely optional — a brand with no Open311 endpoint leaves the base URL `null` and gets a
+region with no Open311 servers.
 
 ## Benefits of This Architecture
 
--- a/onebusaway-android/src/main/java/org/onebusaway/android/util/RegionUtils.java
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/util/RegionUtils.java
@@ -22,10 +22,12 @@ import android.location.Location;
 import android.util.Log;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import org.onebusaway.android.BuildConfig;
 import org.onebusaway.android.R;
 import org.onebusaway.android.api.bridge.RegionsClient;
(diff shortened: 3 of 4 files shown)


bmander:
## Review

**Verdict: correct, well-scoped, ready to merge.** Two optional follow-ups below; nothing blocking.

Verified the claims against the tree and ran the suite in an isolated worktree at the PR head (`bb36447b`): `testObaGoogleDebugUnitTest --tests '*RegionUtilsOpen311Test*' -PwarningsAsErrors=true` → `BUILD SUCCESSFUL`, `tests="3" skipped="0" failures="0" errors="0"`. Main + test Kotlin compiled clean under the CI warnings gate.

### Correctness

- **The fix is right and behavior-preserving on the happy path.** A non-null base URL still yields exactly one server with the same three fields.
- **Reachability claims check out.** `agencyY` is the only flavor with `USE_FIXED_REGION=true` (`flavors/*.gradle:15`) and it does set an Open311 URL, so this is latent today. `kiedybus` has a non-null `FIXED_REGION_NAME` with a null `FIXED_REGION_OPEN311_BASE_URL` — flipping its flag lands on the crash immediately.
- **`requireNonNull` is safe.** The only caller is `RegionRepository.kt:222-223`, guarded by `if (BuildConfig.USE_FIXED_REGION)`. The new throw can't fire on `oba`/`agencyX`/`kiedybus`.
- **`name` really is the only remaining non-null gap.** Sweeping `Region`'s constructor: `id`/`active`/booleans are primitives, `bounds` is built locally, every other `String` is `String?`. Nothing else can NPE here.
- **No sibling instances of the bug.** `RegionUtils` is the only Java construction site for `Region`; `RegionMapper` and `LegacyDataImporter` build the array from Kotlin lists via `.toTypedArray()`, which can't be null. The fix is complete in scope.
- **`@VisibleForTesting` won't trip lint.** It defaults to `otherwise = PRIVATE`, and the production caller is inside `RegionUtils` itself, so `VisibleForTests` stays quiet. Same-package Kotlin test access to the package-private static works (confirmed by the green run).

### Test coverage

The extraction is the right call — `BuildConfig` is a per-variant constant a unit test can't vary, so parameterizing is the only way to reach the branch.

One nit on how the coverage is described. The PR body says the second assertion "asserts the result actually constructs a `Region`, which is what used to blow up" — but the crash was Kotlin's null-check on a **Java** caller, and a Kotlin test can't pass `null` to a non-null parameter at all. So `Region(open311Servers = servers)` is unconditionally constructible and that assertion proves nothing beyond the line above it. The regression is genuinely guarded, just by the *first* assertion: `assertArrayEquals` fails with "actual array was null" on a null actual. Consider dropping the second assertion, or keeping it as documentation with an honest comment.

Minor: existing `RegionUtils` tests live in `androidTest/.../util/test/RegionUtilTest.java`, while this one is `test/.../util/RegionUtilsOpen311Test.kt`. The divergence is forced (package-private access + JVM test) and matches the other `src/test` unit tests like `SituationUtilsTest.kt`. Fine as is.

### Suggestions

**1. A build-time check would beat a launch-time one.** `Objects.requireNonNull` turns an anonymous crash into a labeled crash — real progress, but the rebrander still has to build, install, and launch to discover a typo. Since this PR's whole framing is "a trap laid under rebranders," a `check()` in `flavors/load-flavors.gradle` asserting that `USE_FIXED_REGION == true` implies a non-null `FIXED_REGION_NAME` would fail at configuration time instead. Worth a follow-up; the runtime check is a good backstop either way.

**2. The docs still don't say which fields are load-bearing.** `docs/REBRANDING.md:292` ("Fixed vs. Multi-region") and `flavors/README.md:132-137` both describe `USE_FIXED_REGION` without stating that `FIXED_REGION_NAME` is mandatory when it's true. Two lines there is the cheapest part of this fix and it's the part a rebrander actually reads first.

**3. Consider treating a blank base URL as absent.** If a flavor writes `""` rather than `null`, `open311ServersFrom` builds a server with an empty `baseUrl` and `Open311Subsystem.applyRegion` registers a bogus endpoint (`Open311Subsystem.kt:44-50`). Not a regression — pre-PR behavior was identical — but that same function already applies exactly this normalization one line down (`jurisdictionId?.takeIf { it.isNotEmpty() }`), and `Region.usesOtp2` uses `isNullOrBlank()` for the same "configured or not" question. Explicit normalization with in-repo precedent, not a magnitude guess, so it doesn't run into the no-unsanctioned-heuristics rule.

### Style / conventions

- Follows the codebase's verbose-explanatory comment norm (cf. `Region.kt`'s field comments). The R8 rationale is stated three times though — helper javadoc, caller javadoc, test kdoc. Trimming the caller's version to a sentence would lose nothing.
- Java formatting, `@NonNull`/`@Nullable` annotations, and the 2026 Apache header on the new file all match project convention.
- No time-domain, coroutine, or heuristic concerns — this path touches none of those subsystems. No security surface: no new I/O, no credential handling (the `agencyY` Open311 API key was already in the flavor file).

### Risks

Low. The only behavior change on shipping variants is nil (`oba` never calls this). For `agencyY`, output is byte-identical to before. The new `requireNonNull` can only throw on a flavor that was already going to crash a few lines later.

PR #2025, opened 2026-07-26
Split the mode picker into vehicle and street, and add the rider's own bike

bmander:
> **Stacked on #2024.** Built on `feature/otp2-street-preferences`, so until #2024 merges the diff here also contains its commit. GitHub collapses this to the single commit `Split the mode picker into vehicle and street, and add the rider's own bike` once #2024 lands. **Review/merge #2024 first.**

## The gap

The trip planner could only put a bike in an itinerary by **renting** one. Both bike modes map to `BICYCLE_RENTAL`, so a rider who owns a bike had no way to ask for a plan that carries it onto transit. `PlanAccessMode.BICYCLE` has been in the schema the whole time and the app has never used it.

## Why this restructures the picker instead of adding a sixth entry

Adding "Transit & my bike" to the old **Travel by** list would have made that list worse. Its six entries were already combinations of two independent things — which vehicles the rider boards, and how they cover the street at either end — so most combinations simply didn't exist. There was no way to ask for rail plus your own bike, bus plus bikeshare, or a plain walking route, and every new street option would multiply the list again.

```
Ride:            Any transit | Bus only | Rail only | No transit
Getting around:  Walk | Walk & bikeshare | My own bike
```

`buildModes` **composes** the two rather than tabulating twelve cases. `VehicleMode` decides whether this is a transit search and which transit modes are usable; `StreetMode` decides the access/egress/transfer shape. Only the per-mode rules are stated, each once.

## The rule that makes own-bike not a copy of the rental branch

The two bike modes have **opposite** requirements:

- `BICYCLE_RENTAL` **must** be accompanied by `WALK` — OTP rejects a bare rental leg, since a hired bike must be walked to and from.
- `BICYCLE` must **not** be. The schema: *"access can use cycling only if the mode used for transfers and egress is also `BICYCLE`"*. Your own bike is with you the whole way, so there is no phase without it.

Getting the second wrong doesn't error — the plan silently degrades to walking. A test pins all three phases for that reason, and another pins that own-bike is **not** gated on bikeshare availability (it needs no rental network, only a bike-routable graph) — the mistake you'd make by copying the neighbouring branch.

## `direct` now matches the street mode

OTP returns a direct suggestion alongside transit results and falls back to it when no transit connection exists. Left unset it defaults to `WALK`, so a cyclist whose requested vehicle had no service in range was offered a multi-hour **walk** — measured: rail-only from an origin with no Link station in range returned a 9.8 km / 128 min walk. It is deliberately *not* `directOnly` except under **No transit**, which is what makes that a street route rather than a transit search that tolerates walking.

## Gating and persistence

Availability is per street mode: bikeshare needs a rental network, own bike needs OTP2. The picker lists only what the region can serve, and `getModeSelection` degrades a stored mode the current region can't serve to walking — both halves persist across a region change.

OTP1 takes one flat mode list, so the halves concatenate, reproducing the exact strings sent before the split. That includes the direct bikeshare trip's historical `BICYCLE_RENT` without `WALK`: adding it is arguably more correct but is an untested change to every OTP1 region's bikeshare request, and this isn't the place for it.

The selection now lives in the request bundle by enum name, so it survives the trip-plan monitor's `copyIntoBundleSimple`/`initFromBundleSimple` round trip — the flat mode id it replaced was a plain field that did not. A rider's saved `travel_by` id migrates to the equivalent pair on read, with no migration write.

## Verified against a live server

Combinations the old list could not express now return real itineraries:

```
rail + own bike       ->  BICYCLE + TRAM(1 Line) + BICYCLE
bus  + bikeshare      ->  WALK + BICYCLE + BUS + WALK
no transit + own bike ->  a single BICYCLE leg
any transit + walk    ->  no `modes` sent at all (byte-identical to before)
```

## What I could not verify

**That OTP excludes bike-hostile trips.** Every transit leg returned for the own-bike mode reported `bikesAllowed: ALLOWED` — but that feed marks *every* trip `ALLOWED`, so nothing contradicts the expected filtering and nothing demonstrates it.

**Bike + rail is degraded on the tested deployment** for reasons outside this PR — see #2026. A bike search often returns bus-only itineraries where the walking equivalent uses rail, because of the server's access/egress stop cap. Not fixable client-side.

## Deliberately not included

**Bike-and-park** (`BICYCLE_PARKING` access, walk at the far end) — the natural sibling and a common commute pattern. The probe returned `NO_STOPS_IN_RANGE` and fell back to a 17 km walk, so that deployment has no bike parking in its graph. Offering a mode that silently returns nothing would be worse than not offering it.

## Tests

`Otp2PlanRequestBuilderTest` 29, `TripPlanViewModelTest` 24, both `failures="0" errors="0"`. `testObaGoogleDebugUnitTest`, `compileObaMaplibreDebugKotlin`, `spotlessCheck` green under `-PwarningsAsErrors=true`. Device-verified on an OTP2 region.

## Summary by CodeRabbit

* **New Features**
  * Added separate vehicle and street mode selection for trip planning, including transit, walking, bikeshare, and bicycle options.
  * Added walking, cycling, and biking preference controls for supported routing services.
  * Preserved advanced trip preferences across sessions and region changes.
  * Added support for transit combined with riding a personal bicycle.

* **Bug Fixes**
  * Improved handling when selected travel modes are unavailable, automatically falling back to walking.
  * Improved routing behavior for bikeshare and bicycle access, transfers, and direct trips.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/directions/util/Otp2PlanRequestBuilder.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/directions/util/Otp2PlanRequestBuilder.kt
@@ -34,6 +34,7 @@ import org.onebusaway.android.api.graphql.type.PlanLocationInput
 import org.onebusaway.android.api.graphql.type.PlanModesInput
 import org.onebusaway.android.api.graphql.type.PlanPreferencesInput
 import org.onebusaway.android.api.graphql.type.PlanStreetPreferencesInput
+import org.onebusaway.android.api.graphql.type.PlanTransferMode
 import org.onebusaway.android.api.graphql.type.PlanTransitModePreferenceInput
 import org.onebusaway.android.api.graphql.type.PlanTransitModesInput
 import org.onebusaway.android.api.graphql.type.TransferPreferencesInput
@@ -43,7 +44,9 @@ import org.onebusaway.android.api.graphql.type.WalkPreferencesInput
 import org.onebusaway.android.api.graphql.type.WheelchairPreferencesInput
 import org.onebusaway.android.ui.tripplan.BikePreference
 import org.onebusaway.android.ui.tripplan.CyclingPreference
-import org.onebusaway.android.ui.tripplan.TripModes
+import org.onebusaway.android.ui.tripplan.StreetMode
+import org.onebusaway.android.ui.tripplan.TripModeSelection
+import org.onebusaway.android.ui.tripplan.VehicleMode
 import org.onebusaway.android.ui.tripplan.WalkPreference
 import org.onebusaway.android.util.BikeshareAvailability
 
@@ -231,7 +234,7 @@ object Otp2PlanRequestBuilder {
                     bikePreference = builder.getBikePreference()
                 )
             ),
-            modes = buildModes(builder.getModeSetId(), BikeshareAvailability.isTripPlanningEnabled(context)),
+            modes = buildModes(builder.getModeSelection(), BikeshareAvailability.isTripPlanningEnabled(context)),
             numItineraries = NUM_ITINERARIES,
             alternativeLegs = ALTERNATIVE_LEGS
         )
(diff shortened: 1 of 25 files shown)


aaronbrethorst:
nice!

PR #2029, opened 2026-07-26
Log OTP2 GraphQL errors and stop advising a futile retry

### LINKED ISSUE #2023: OTP2 GraphQL errors surface as "something went wrong" and are never logged
When an OTP2 `planConnection` request comes back with a GraphQL `errors` array, the trip planner shows:

> Sorry, something went wrong.  Please try again.

…and throws away the server's own explanation, which is usually precise enough to name the exact problem. For a malformed request that advice is also wrong: retrying sends the identical query and fails identically, forever.

## How it happens

`Otp2Planner.plan` unwraps the response with `dataOrThrow()`. A GraphQL-level error makes that throw an `ApolloGraphQLException` (an `ApolloException` subclass carrying the `errors`), which lands in the broad catch:

```kotlin
} catch (e: ApolloException) {
    // A non-network ApolloException (HTTP status, parse failure, GraphQL-protocol error) isn't
    // a timeout — surface it as an unclassified request failure, not a connectivity one.
    throw TripPlanException(TripPlanError.Unknown, e)
}
```

`TripPlanError.Unknown` → `R.string.tripplanner_error_not_defined` → the generic message above.

The exception *is* attached as the `cause`, but nothing ever reads it: `TripPlanException` extends `IOException(cause)`, and there is no `Log` call anywhere on that path. So the server's message is discarded rather than merely un-surfaced — it never reaches logcat, a bug report, or Crashlytics.

Note this is distinct from `routingErrors`, which `resolveOtp2Plan`/`otp2ErrorFor` already classify well. This is the *transport/validation* tier above that, and it's the one with no classification at all.

## Worked example

Sending a `walk.reluctance` of `0.08` (below OTP's floor) returns HTTP 200 with a 207-byte body:

```json
{"errors":[{"message":"Validation error (WrongType@[planConnection]) : argument 'preferences.street.walk.reluctance' with value 'FloatValue{value=0.08}' is not a valid 'Reluctance' - Reluctance needs to be between 0.1 and 100000.0","locations":[{"line":1,"column":193}],"extensions":{"classification":"ValidationError"}}]}
```

That message names the offending argument, the rejected value, and the valid range. The rider saw "Sorry, something went wrong. Please try again." Nothing was logged. Diagnosing it required noticing an anomalous response size in an OkHttp `BASIC` log line and then replaying the query against the live server by hand — when the answer had already been returned, in full, to the app.

(The `0.08` that produced this is fixed; the point is the class of failure, not that instance. Any future preference/mode/date combination the server rejects will behave the same way.)

## Suggested direction

Roughly in order of value per unit of effort:

1. **Log the GraphQL errors.** Even without any UI change, `Log.e(TAG, …)` of the `errors` on this path turns a silent dead end into a one-line diagnosis, and gets the text into bug reports. This alone would have saved the whole investigation above.
2. **Classify `ApolloGraphQLException` separately from the rest of `ApolloException`.** They mean different things: a GraphQL error is a *deterministic* rejection of this request, whereas an HTTP-status or parse failure may well be transient. Only the latter is worth "try again".
3. **Stop advising a retry for the deterministic case.** A distinct string — something like "This trip request could not be completed. Try changing your trip preferences." — is honest about the fact that repeating the identical request cannot help, and points at the settings, which is where the cause almost always is.

Worth deciding as part of this: whether `extensions.classification == "ValidationError"` should be treated as its own category. It reliably means "the client built an invalid query", which is arguably a bug to report rather than something to ask the rider to fix — but the practical trigger today is a user-facing preference value, so the rider-facing wording still matters.

Deliberately **not** suggested: surfacing the raw GraphQL message in the UI. It is developer-facing English, untranslated, and leaks schema internals.

## Testing note

`otp2ErrorFor` and `resolveOtp2Plan` are already `internal` and JVM-unit-testable without Apollo, so a classifier for this tier can follow the same pattern and be tested from a constructed exception.

Found while adding walk/bike reluctance preferences to the trip planner.

bmander:
Fixes #2023.

## The problem

When an OTP2 `planConnection` response came back with a GraphQL `errors` array, the rider saw *"Sorry, something went wrong.  Please try again."* and the server's own explanation was thrown away — the one that names the offending argument, the rejected value, and the accepted range. For a request the server refuses deterministically, "try again" is also wrong advice: the identical query fails identically, forever.

**The loss is worse than the issue describes.** Apollo 5's `dataOrThrow()` is just `data ?: throw NoDataException(cause = exception)`, and a GraphQL-errors-only response has `data == null` *and* `exception == null`. So the old code threw a `NoDataException` with a **null cause** into the broad `catch (ApolloException)` — the errors weren't merely un-surfaced, they were never attached to anything, and nothing on that path logged.

Restructuring to read the response rather than unwrap it also fixes **a second latent bug**: Apollo 4+ reports transport failures on `ApolloResponse.exception` instead of throwing them, so the `catch (ApolloNetworkException)` around `execute()` could never fire — OTP2 timeouts were classified as unknown request failures, not connectivity ones.

## The change

- **Log the errors** (message, `extensions.classification`, response path) where they arrive — including when the response *also* carried usable data, since GraphQL permits partial results and a renderable plan is no reason to drop the explanation of what was left out of it.
- **Classify the transport tier off `response.exception`**, so a network failure reaches the connectivity result it was always meant to.
- **Classify the GraphQL tier** in a new `otp2GraphQlErrorFor` — the tier above `routingErrors`, which had no classification at all.
- **Attach the errors as the exception's cause**, so the server's wording survives into a crash report even when logcat wasn't captured.

Per the issue, the server's message is deliberately **not** shown to the rider: it's untranslated, developer-facing English that leaks schema internals.

## Decision: is `ValidationError` its own category?

The issue asked this to be settled. **Not a separate user-facing category, but a separate user-facing *result*.** The split that matters isn't validation-vs-everything, it's *deterministic vs possibly-transient*:

| `extensions.classification` | Result |
|---|---|
| `InvalidSyntax`, `ValidationError` | `REQUEST` + `tripplanner_error_bogus_parameter` — "Something went wrong.  Change your trip options and try again." |
| everything else (incl. absent/unrecognized) | `TripPlanError.Unknown` — the generic try-again |

The first two are decided by the parser and the validator against the schema, so the request cannot succeed as sent; the rest happen during execution and may not recur, where "try again" stays honest.

The deterministic case reuses the **existing, already-localized** `tripplanner_error_bogus_parameter` rather than adding a new untranslated string. That's the string OTP1's `BOGUS_PARAMETER` maps to, and it means the same thing — the server refused the parameters we sent — so the two planner paths stay symmetric. In practice the trigger is a trip-preference value the rider can actually change, which is where it points them.

## On the heuristics rule

The classification match is an **exact** match on an explicitly-stated field with a documented, closed value set — not an inference from message text or magnitude. OTP2 runs on graphql-java, whose `graphql.ErrorType` enumerates exactly `InvalidSyntax`, `ValidationError`, `DataFetchingException`, `NullValueInNonNullableField`, `OperationNotSupported`, `ExecutionAborted`, and emits the constant's name verbatim under `extensions.classification`. Verified against both the enum source and the live OTP2 response in the issue. An absent or unrecognized value falls back to today's behaviour, so a server that stops sending the key degrades to the status quo rather than misbehaving.

## Testing

Both new classifiers are top-level and `internal` like `otp2ErrorFor`, and the tier resolution is factored into `resolveOtp2Response`, which takes the three fields an `ApolloResponse` reports a plan through rather than the response itself — so every case runs on the JVM from constructed `com.apollographql.apollo.api.Error` values, with no Apollo client and no HTTP.

- `TripPlanErrorMappingTest` gains the classifier cases: the motivating `walk.reluctance` `ValidationError`, `InvalidSyntax`, an execution-tier failure, a missing classification, an unrecognized one, a non-string one, a mixed list, the empty list, and both transport branches.
- `Otp2PlanResolveTest` gains the tier-precedence cases: transport over GraphQL, itineraries over GraphQL, `routingErrors` over GraphQL, a `planConnection` nulled by a field error classifying from the errors rather than as "no route", an empty `errors` array reading as no errors at all, and the exception message falling back to its cause.

Ran locally: `compileObaGoogleDebugKotlin -PwarningsAsErrors=true`, the full `testObaGoogleDebugUnitTest` suite, and `spotlessCheck` — all green. Lint left to CI per CLAUDE.md.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trip-planning error handling for network failures, server errors, rejected requests, and unknown failures.
  - Preserves server-provided details when displaying trip-planning errors.
  - Correctly identifies invalid trip options and suggests request changes where applicable.
  - Aligns timeout and invalid-parameter errors with their appropriate user-facing categories.

- **Tests**
  - Added coverage for OTP1 and OTP2 error classifications, fallback behavior, and mixed server responses.


bmander:
Addressed in d68f73a9.

The finding is valid. `Otp2PlanRequestBuilder` sends the trip's origin and destination as `PlanCoordinateInput`, and graphql-java renders a rejected variable's value straight into the message — the #2023 sample (`with value 'FloatValue{value=0.08}'`) is exactly that shape — so a validation error on a coordinate argument would have put the rider's location into logcat and onto the exception.

**What changed.** `otp2ErrorDiagnostic` now renders the error in two halves:

- **Every build** keeps the structural half — `classification`, response `path`, document `locations`. The server *composes* these rather than quoting them back from the request, so they cannot contain rider data, and they are what the classification tiers act on.
- **Debug builds only** keep the message text. That is where it is actually read — a developer reproducing against a region, which is the #2023 workflow — so the gate costs nothing operationally.
- Extensions keys other than `classification` are dropped rather than rendered. That is the one key whose meaning this app has verified against graphql-java's `ErrorType`; an unexamined key could quote request values just as the message does.

**On the suggested remedy.** Scrubbing argument values out of the message text was rejected deliberately: recognizing "the coordinate part" of a free-text English sentence is the kind of fuzzy inference this repo's `CLAUDE.md` no-heuristics rule forbids, and it would fail silently the first time OTP reworded an error. Dropping the whole untrusted field is the version that can't rot.

**Scope note**, since it bounds the original exposure: logcat never leaves the device in this app (no `READ_LOGS`, and the feedback uploader ships the navigation trip-log *file*, not logcat), and `TripPlanException` is swallowed by both repository call paths, so it doesn't reach Crashlytics today either. The redaction is belt-and-braces against both of those changing.

Seven new cases in `Otp2PlanResolveTest` pin the behaviour: text dropped and kept per build type, `path`/`locations` surviving the redaction, unexamined extensions dropped, the unclassified rendering, multi-error rendering, and the no-errors null. Verified non-vacuous by reverting the gate — the redaction tests fail without it.

PR #2031, opened 2026-07-26
Build release notes from Changelog commit trailers

bmander:
Release notes are currently written by hand at release time, and `docs/BUILD.md` tells the maintainer to "copy the release notes from a previous release and edit accordingly." This derives them from the commit messages instead.

## The convention

Every PR lands on `main` as a single squashed commit, so one commit is one change — which makes the commit message a usable place to record what that change means to a rider:

```
Offer bikeshare trips where the OTP2 server supports them (#2017)

<body>

Changelog: Bikeshare trips now show up in the trip planner.
```

**The trailer is optional and most commits should not have one.** Our commit subjects are already release-note quality — "Don't crash when a vehicle's trip is missing from the route references (#2021)" needs no help, and every subject reaches the GitHub release body regardless. The trailer covers the gap a subject can't: "Fold report/ into ui/report/ and wire its ViewModels through Hilt" is the right subject and the wrong sentence to show someone waiting for a bus. Expect a handful per release.

## The tooling

`tools/release-notes.sh` reads the trailers since the last `v*` tag (`--since <ref>` to override):

| Command | Output |
|---|---|
| `play` | Play Store "What's new" text — trailers as bullets |
| `play --write` | ...written to `src/obaGoogleRelease/play/release-notes/en-US/default.txt`, where GPP already looks |
| `github` | A `## Highlights` block to prepend to a generated release body |

`github` emits *only* the highlights: `gh release create --generate-notes` already builds the PR list, authors and compare link, and prepends whatever `--notes` passes. No point re-implementing GitHub's release-notes API.

Play's 500-character-per-locale cap is enforced as a **hard failure, never a truncation** — Play cuts from the end, which would sever an entry mid-word:

```
Error: release notes are 775 characters, 275 over the Play Store limit of 500.
       Shorten or drop a 'Changelog:' trailer — do not let this be truncated,
       because Play truncates from the end and would cut an entry mid-word.
```

## Does squash-merge break the trailers?

This was the main risk, so I checked rather than assumed. Git only recognizes a trailer in the block at the very *end* of a message, and GitHub's merge box concatenates a multi-commit branch's messages — which could strand a trailer mid-body.

Measured on the last 50 commits of `main`: **31 of 31 `Co-authored-by:` lines parse correctly as trailers, zero stranded** — including commits carrying 2–9 concatenated `* ` bullets. The trailing trailer block survives concatenation in practice.

The script still detects the case and names the offending commits, as a backstop:

```
Warning: these commits have a 'Changelog:' line that git did not read as a
         trailer, so it is missing below. ...
  018d4720a Offer bikeshare trips where the OTP2 server supports them (#2017)
```

**Recommended (not included here, needs a maintainer):** setting the repo's squash-merge default to "pull request title and description" makes the final commit message predictable and editable before merge, and rules the case out entirely. Documented in `BUILD.md`; no repository settings are changed by this PR.

## The in-app "What's New?" dialog is deliberately not automated

`main_help_whatsnew` is translated into es/fi/it/pl. Generating it would mean four locales describing the *previous* release every time — worse than today, where the text is generic and correctly translated everywhere. `BUILD.md` now records that as intentional so it doesn't get "fixed" later. Two sinks generated, one left alone.

## No CHANGELOG.md

Deliberate. For a consumer app the GitHub release and the Play listing *are* the changelog; a third in-repo copy is one more thing to keep honest. Easy to add later if something wants to read it.

## Files

- `tools/release-notes.sh` — the generator (bash, matching the existing `tools/` scripts)
- `.gitmessage` — opt-in commit template (`git config commit.template .gitmessage`)
- `.github/CONTRIBUTING.md` — "Changelog entries" section; owns the convention
- `.github/PULL_REQUEST_TEMPLATE.md` — one checkbox
- `docs/BUILD.md` — "Release notes" section, wired into the existing release steps

No app code, no build changes.

## Deliberately not included

- A tag-triggered workflow that auto-creates GitHub releases — that changes the release process rather than supporting it. `BUILD.md` documents the `gh release create` one-liner instead.
- A Gradle task the publish step depends on. It would make stale notes impossible by construction, and is the natural follow-up, but it couples this to the build.
- Locales beyond `en-US`, documented as a known limitation.

## Testing

Verified against real history and a scratch repo: happy path, no-trailers, stranded trailer, duplicate trailer, over-limit, `--write`, bad arguments, and invocation from three different working directories.

## Summary by CodeRabbit

- **New Features**
  - Added automated release-note generation for Google Play “What’s new” and GitHub releases.
  - Release notes now support concise, rider-facing highlights derived from commit `Changelog:` trailers.
  - Google Play output can be written directly for the publishing workflow, with character-limit validation.

- **Documentation**
  - Expanded release workflow docs, including how to generate and paste “What’s new” text and how localization is handled.
  - Updated contribution and PR template guidance for `Changelog:` trailer formatting and placement.
  - Added a structured commit-message template to standardize release-note entries.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -33,6 +33,44 @@ first. The one-time bulk reformat is listed in `.git-blame-ignore-revs`; enable 
 `git config blame.ignoreRevsFile .git-blame-ignore-revs` so `git blame` skips it (GitHub does this
 automatically).
 
+## Changelog entries
+
+Pull requests are squashed when they merge, so your PR becomes one commit on `main`, and that commit
+message is what the release notes are built from. Two things follow from that:
+
+**Write the subject as a changelog line.** Imperative, plain language, issue number at the end — it
+gets read by people scanning a release, not just by people reading `git log`:
+
+```
+Offer bikeshare trips where the OTP2 server supports them (#2017)
+```
+
+**If a rider would notice the change, add a `Changelog:` trailer** at the very end of the message,
+describing the effect in their words rather than the code's:
+
+```
+Changelog: Bikeshare trips now show up in the trip planner.
+```
+
+Most PRs don't need one — refactors, dependency bumps, test changes and internal cleanups are covered
+by their subject line alone. The trailer is for the few changes per release that a rider would notice,
+because "Fold report/ into ui/report/ and wire its ViewModels through Hilt" is the right subject and
+the wrong sentence to show someone waiting for a bus. Keep it to one sentence: a release's trailers
+share a 500-character Google Play budget.
+
+The trailer must be in the trailer block at the end of the message or git won't read it as one. If
+your branch has several commits, check the squashed message in the merge box before merging —
+GitHub concatenates the individual commit messages, which can leave the trailer stranded mid-body.
+`tools/release-notes.sh` warns when that happens.
+
+To have the format in front of you when you commit, opt into the message template:
+
+```bash
+git config commit.template .gitmessage
+```
+
+See [BUILD.md](../docs/BUILD.md#release-notes) for how the notes are generated at release time.
+
 ## Closing policy for issues and pull requests
 
 OneBusAway for Android is a popular project and the capacity to deal with issues and pull requests is limited. Out of respect for our volunteers, issues and pull requests not in line with the guidelines listed in this document may be closed without notice.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,3 +3,5 @@ Please make sure these boxes are checked before submitting your pull request - t
 - [ ] Format your changes with `./gradlew spotlessApply`.
 
 - [ ] Run the unit tests with `gradlew connectedObaGoogleDebugAndroidTest` to make sure you didn't break anything
+
+- [ ] If a rider would notice this change, end your commit message with a `Changelog:` trailer describing the effect in their words — see [Changelog entries](CONTRIBUTING.md#changelog-entries). Most PRs don't need one.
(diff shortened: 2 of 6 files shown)


bmander:
Thanks — three of the four were real. Addressed in 21a3c567e.

**Empty range when HEAD is already tagged** (`docs/BUILD.md:186`) — real, and the most serious. Reproduced: the tag is cut before the notes are generated, so the default range resolves to `vx.y.z..HEAD` and the body came out with no highlights *and no error*, since "no trailers in this range" is legitimate for a maintenance release.

Fixed in the script rather than the documented command: if HEAD is itself tagged, measure from the release before it. The suggested `--since "$(git describe --tags --abbrev=0 vx.y.z^)"` fixes the one line in BUILD.md but leaves the trap armed for anyone running the script by hand after tagging — and `play`'s documented step only escapes it today because it happens to run before the tag exists. The `gh release create` line in BUILD.md is unchanged and now works as written.

**Trailing newline against Play's cap** (`tools/release-notes.sh:125`) — real. `notes` loses its newline to command substitution and both writes add it back. Boundary is now exact and tested: 500 characters written passes, 501 fails.

**"Already uploaded" claim too broad** (`docs/BUILD.md:198`) — fair. Rather than split the sentence by publish task, I dropped the claim: whether GPP uploaded the notes depends on which task ran, and that's not worth asserting inside a step about the console UI. It now says to check what the console already shows before pasting.

**`gradlew` → `./gradlew`** (`.github/PULL_REQUEST_TEMPLATE.md:5`) — applied, though it's pre-existing and unrelated to this PR. Taking it because it's one character and line 3 already had the `./`.

**Not taking:** the MD040 warning on the new fenced block in `docs/BUILD.md`. All 10 fences in that file are untagged; tagging only the new one would make it the odd one out. Worth a separate pass over the file if the project wants the lint clean.

Issue #2000, opened 2026-07-23
Self-interline in directions directs to exit bus and reboard it


bmander:
Some routes involve an interline of a route onto itself. For example a trip plan route from 19th and thomas  to 3rd and pine in Seattle is satisfied by taking the 10 or the 12. At a particular point inbound the trip silently transitions to an outbound trip. For route 12 for example the trip plan advises to get off at Pine and 9th and re-board the same 12 vehicle at Pine and 9th. For this particular trip the directions shouldn't mention the seam between trips at all. Hopefully this is deducible in a principled way from the OTP2 trip structured data.

Related: investigate whether stay-aboard interlining between two routes is properly handled in directions.

PR #2045, opened 2026-07-27
Bring schedule-deviation colors into parity with OBA iOS (#2043)

### LINKED ISSUE #2043: Bring schedule-deviation colors into parity with OBA iOS
The app uses color to convey schedule deviation in six different places, and they disagree with each other and with OBA iOS on almost every axis: which hue means what, what counts as "on time", and whether the map should encode punctuality at all. This issue brings the whole surface into parity with the iOS app.

### 1. There is no on-time band — "on time" is nearly unreachable

`ArrivalInfoUtils.computeColorFromDeviation` buckets on strict integer comparison, and its input has already been truncated to whole minutes-past-epoch by the caller:

```java
// ArrivalInfoUtils.java:145
public static int computeColorFromDeviation(final long delay) {
    if (delay > 0)      return R.color.stop_info_delayed;
    else if (delay < 0) return R.color.stop_info_early;
    else                return R.color.stop_info_ontime;
}
```

```kotlin
// ArrivalInfo.kt:181
val scheduledMins = scheduled.epochMs / MS_IN_MINS
val predictedMins = predictedTime?.let { it.epochMs / MS_IN_MINS } ?: 0L
...
color = ArrivalInfoUtils.computeColor(scheduledMins, predictedMins)
```

So "on time" means the predicted and scheduled *minute-since-epoch are exactly equal*. A bus 20 seconds late renders in the "late" color. On live data the green on-time state is nearly never shown — which is the single most user-visible defect here.

iOS instead uses a **±1.5 minute band** (`ArrivalDeparture.swift:305`):

```swift
public var scheduleStatus: ScheduleStatus {
    guard predicted else { return .unknown }
    let minutesDiff = rawDeviationFromScheduleInMinutes
    if minutesDiff < -1.5 { return .early }
    else if minutesDiff < 1.5 { return .onTime }
    else { return .delayed }
}
```

Note that the precision is destroyed *before* it reaches the color function, so this isn't a one-line threshold change — `computeColor`/`statusColor` take whole minutes and need to take a sub-minute value.

### 2. No dark-mode variants

`values-night/colors.xml` contains **zero** `stop_info_*` overrides. The light-mode swatches are used verbatim on a dark surface. iOS explicitly does not do this — `Theme.swift:127` carries a comment that they hand-picked a darker green for light mode because `systemGreen` was illegible as small text, and kept `systemGreen` for dark, citing user reports in their issues #506/#508. We have the same small-colored-text problem and none of the mitigation.

### 3. On-time is tied to `brand_color`, which collides for white-label brands

`stop_info_ontime` resolves `@color/theme_primary_brand` → `theme_primary` → `brand_color`. For the `agencyY` flavor `brand_color` is `#2196F3` blue, so **on-time and late are both blue** there. `agencyX` and `kiedybus` paper over this by hardcoding `stop_info_ontime` back to `#4CAF50`; `agencyY` does not, and is currently broken. Deviation status is semantic, not brand identity — it should not be brand-tinted at all.

### 4. The map colors vehicle markers by deviation; iOS deliberately doesn't

`VehicleBitmaps.colorResource` (`VehicleBitmaps.kt:152`) tints the vehicle disc by schedule deviation. On iOS, grepping `scheduleStatus|colorForScheduleStatus|departure(Early|Late|OnTime)` across `OBAKit/Mapping/**` returns **zero hits** — the map never encodes punctuality. Vehicle markers there are tinted by **route color** with a brand fallback (`TripViewController.swift:419`), and the only status the marker carries is realtime-vs-not: live vehicles pulse in route color, non-realtime go gray with the pulse disabled (`PulsingVehicleAnnotationView.swift:43`).

At map zoom a colored disc reads as identity, not punctuality, and we already have `ObaRoute.color: Int?` available to say something more useful.

### 5. The trip planner's polarity is inverted relative to arrivals

Two independent palettes with contradictory meanings:

| | early | on time | late |
|---|---|---|---|
| arrivals / map (`stop_info_*`) | `#af504c` red | brand (`#4CAF50`) | `#504caf` **blue** |
| trip planner (`trip_realtime_*`) | `#1F5E8C` **blue** | `#136B39` green | `#7E560E` amber |

Blue means *late* in the arrivals drawer and *early* in the trip results screen, in the same app. The trip planner also has its own, different band — `TripLogBuilder.kt:225` uses `roundToLong()` on seconds, giving it a true ±30s window that arrivals lacks.

### A. Adopt the iOS palette verbatim, with light/dark variants

Replace the `stop_info_*` deviation colors with the iOS values, decoupled from `brand_color`:

| status | light | dark | iOS source |
|---|---|---|---|
| early | `#FF3B30` | `#FF453A` | `.systemRed` |
| on time | `#129900` | `#30D158` | hand-picked / `.systemGreen` |
| late | `#007AFF` | `#0A84FF` | `.systemBlue` |
| no realtime | `#8E8E93` | `#8E8E93` | `.systemGray` |

Note the polarity is the counterintuitive one on both platforms — **red = early, blue = late** — because a vehicle running ahead of schedule is a risk to the rider (they can miss it), not a reassurance. That reasoning is already written down at `TripResultsScreen.kt:1173`; this just makes the whole app agree with it.

Add the `values-night` variants that don't exist today. Remove the now-redundant `stop_info_ontime` overrides from the `agencyX` and `kiedybus` flavors, and confirm `agencyY`'s blue-on-blue collision is gone.

### B. Adopt the ±1.5 minute on-time band, app-wide

- Widen `ArrivalInfoUtils.computeColor` / `statusColor` to take a sub-minute deviation (a `kotlin.time.Duration`, consistent with the typed-time work in #1620) instead of whole minutes, and stop truncating at `ArrivalInfo.kt:181` before the color is computed.
- Apply the same band to `TripLogBuilder.realtimeState()`, replacing its `roundToLong()` ±30s window, so the two paths stop disagreeing.
- Unit-test the boundaries — this is exactly the branchy pure logic that belongs in a JVM test: −91s → early, −89s → on time, +89s → on time, +91s → late, `predicted == false` → scheduled regardless of deviation.

**Sign-off note:** ±1.5 min is a magic threshold, which the "No unsanctioned heuristics" rule in `CLAUDE.md` gates on explicit human approval. It is sanctioned here by a named upstream source (the iOS app's shipped behavior, `ArrivalDeparture.swift:305`) rather than invented, and the parity rationale should be recorded at the call site.

### C. Stop coloring map vehicles by deviation

In `VehicleBitmaps`, replace the deviation tint with:
- **realtime** → `ObaRoute.color`, falling back to brand color when the route carries none (`ObaRoute.color` is nullable — many feeds omit it);
- **not realtime** → gray.

Marker color then means route identity + liveness, never punctuality. `VehicleInfoWindow.kt:77` keeps the deviation color for its *text*, which is the iOS treatment (text carries status, marker carries identity) — the info window is where punctuality belongs.

Watch the bitmap cache: `createBitmapCacheKey` (`VehicleBitmaps.kt:189`) keys on a color **resource id**, and a route color is a raw ARGB int with no resource id. The key needs to become the resolved color value.

### D. Fold in the dependent palettes

- **Trip-planner chips** (`trip_realtime_*`): retire in favor of the unified `stop_info_*` set, fixing the inverted polarity.
- **Starred-stop badges** (`badge_*`, `MyListRepository.kt:399`): these are deliberately darkened for WCAG AA white-on-color, so they can't just alias the new palette — re-derive them from it and re-verify contrast. iOS has the same hazard unaddressed (its badge paints white text on `.systemGray` and on `#129900`); we should not copy that part.
- **Trip details** (`TripDetailsRepository.kt:231`): drop the on-time → `theme_primary` remap, which exists only because on-time was brand-tinted.
- **Directions** (`ConversionUtils.kt:281`): follows automatically, verify.
- **Legend** (`HelpDialogs.kt:185`): must be updated to describe the new palette *and* the new band, or it will document behavior the app no longer has.

## Out of scope

Adding non-color affordances (arrows, +/− signs, "Differentiate Without Color" support). Worth noting iOS has none of these either — it relies on redundant text ("arrives 3 min late", "Scheduled/not real-time") next to the colored value, which we also already do. Tracking that separately keeps this issue about parity rather than exceeding it.

## Acceptance criteria

- [ ] A vehicle 60s late renders on-time green in the arrivals drawer; a vehicle 2 min late renders late blue.
- [ ] Arrivals and trip-planner results agree on hue for the same deviation.
- [ ] Deviation colors have distinct, legible light and dark variants; small colored text is checked on both surfaces.
- [ ] `obaGoogle`, `agencyX`, `agencyY`, and `kiedybus` all render four visually distinct deviation states — in particular `agencyY` no longer shows on-time and late as the same blue.
- [ ] Map vehicle discs are route-colored (brand fallback) when live and gray when not, with no deviation input; info-window text still shows deviation color.
- [ ] Boundary unit tests cover ±89s/±91s and the not-realtime case for both the arrivals path and the trip-planner path.
- [ ] The help-dialog legend matches shipped behavior.

## Files

- `util/ArrivalInfoUtils.java:121-157` — band + signature
- `ui/arrivals/ArrivalInfo.kt:181-203` — stop truncating before the color call
- `ui/arrivals/components/EtaStrip.kt` — ETA pills consume the same colors
- `map/render/VehicleBitmaps.kt:152, 189` — marker tint + cache key
- `map/compose/VehicleInfoWindow.kt:77, 92` — keeps deviation color
- `ui/tripresults/TripLogBuilder.kt:223-231`, `ui/tripresults/TripResultsScreen.kt:1167-1177` — chips
- `ui/mylists/MyListRepository.kt:399-405` — badges, re-derive for contrast
- `ui/tripdetails/TripDetailsRepository.kt:231-232` — drop the on-time remap
- `directions/util/ConversionUtils.kt:281`
- `ui/home/help/HelpDialogs.kt:185-187` — legend
- `res/values/colors.xml:56-59, 65-68, 97-99`, `res/values-night/colors.xml` (new entries), `src/agencyX/res/values/colors.xml:24`, `src/kiedybus/res/values/colors.xml:31`

## iOS reference

- `OBAKitCore/Models/REST/ArrivalDeparture.swift:300-320` — `ScheduleStatus` + the ±1.5 band
- `OBAKitCore/Theme/Theme.swift:125-172` — the palette, hard-coded rather than in an asset catalog
- `OBAKitCore/Utilities/Formatters.swift:137-155, 443-462` — status → color, and the half-colored explanation string
- `OBAKit/Mapping/PulsingVehicleAnnotationView.swift:43-52`, `OBAKit/Trip/TripViewController.swift:419-423` — route-colored markers, no deviation

bmander:
Closes #2043.

Color conveyed schedule deviation in six places that disagreed with each other and with OBA iOS on almost every axis. This brings the whole surface to parity.

## The on-time band

`computeColorFromDeviation` bucketed on a strict sign test, and its input had already been truncated to whole minutes-past-epoch by the caller — so "on time" meant the predicted and scheduled minute-past-epoch were *exactly equal*. A bus 20 s late rendered late, and on live data the green on-time state was nearly never shown.

The color oracle moves out of `ArrivalInfoUtils.java` into Kotlin as `util/ScheduleDeviation` (every caller was already Kotlin), takes a full-precision `kotlin.time.Duration`, and applies iOS's ±1.5 minute band. `ArrivalInfo` stops flooring before the color call; `TripLogBuilder` drops its own `roundToLong` ±30 s window. The five different deviation→minutes rules that existed across the app become one:

| site | was | now |
|---|---|---|
| `ArrivalInfo` | floor each epoch, then subtract — zero-width band | `ScheduleDeviation` |
| `TripDetailsRepository` | `inWholeMinutes` — ±59 s | `ScheduleDeviation` |
| `VehicleBitmaps` / `VehicleInfoWindow` | `inWholeMinutes` — ±59 s | `ScheduleDeviation` |
| `TripLogBuilder` | `roundToLong` — ±30 s | `ScheduleDeviation` |
| `ConversionUtils` | **milliseconds passed to a minutes parameter** | `ScheduleDeviation` |

## The palette

`stop_info_*` adopts the iOS values and is **decoupled from `brand_color`** — deviation status is semantic, not brand identity. agencyY previously rendered on-time and late as the same blue; agencyX and kiedybus hardcoded on-time back to green to paper over it, and those overrides are gone. The `values-night` variants that never existed are added.

The palette is **two tiers**, which is the one place this goes beyond the issue. `EtaPill`, the starred-stop badges, the trip-details status pill and the map info-window chip all paint `Color.White` *on* the deviation color, where the iOS hexes fail AA:

| | white-on-fill |
|---|---|
| `#FF3B30` early | 3.55 |
| `#129900` on time | 3.76 |
| `#007AFF` late | 4.02 |
| `#8E8E93` scheduled | 3.26 |

So the iOS hexes are the **foreground** tier (text/glyph), and a parallel `stop_info_*_fill` tier — darkened to 4.8:1 white-on — backs every filled surface. This generalizes the reasoning the old `badge_*` colors already encoded to the pill, which had identical mechanics. iOS has this hazard unaddressed and we deliberately don't copy that part.

**Known tradeoff, flagged for review:** as *text on a light surface* the foreground hexes are 2.95–4.02:1, below AA for small text. That is inherent to matching the iOS values and is what the app ships today for the arrivals ETA (which is 30sp, comfortably over the 3:1 large-text threshold). The remaining small-text foreground uses are the trip-planner chip label and the directions time span. Happy to darken the light-mode foreground tier if you'd rather have AA than literal parity.

## The map

Vehicle discs are colored by `ObaRoute.color` (brand fallback, gray when not real-time) with no deviation input — at map zoom a colored disc reads as identity, not punctuality, which is also what iOS does. Deviation keeps its home in the info window.

The bitmap cache key becomes the **resolved ARGB** rather than a color resource id, since a route color is a raw ARGB int off the wire with no resource id. That also closes a latent staleness bug: the same resource id resolving to a different color across a light/dark switch was invisible to the old key.

Note `iconKey` gains a `Context` parameter to resolve the fallback.

## Dependent palettes

- `trip_realtime_*` retired — its blue meant "early" where the arrivals drawer's blue meant "late", in the same app.
- `badge_*` re-derived from the new palette and re-verified for contrast (they can't just alias it).
- The trip-details on-time → `theme_primary` remap dropped; it existed only because on-time was brand-tinted.
- The help legend reads its swatches from `ScheduleDeviation.Status` so it can't document a palette the app no longer paints, and its labels now state the band.
- `MyListRepository.badgeColor` reverse-mapped the foreground color resource id back to a badge color and silently fell through to "scheduled" for anything unrecognized — the badge now takes the fill tier off the arrival directly.

## Sign-off needed: the ±1.5 minute threshold

±1.5 min is a magic threshold, which CLAUDE.md's "No unsanctioned heuristics" rule gates on explicit human approval rather than an implementer's judgment. It is sanctioned by a **named upstream source** — the iOS app's shipped behavior, `ArrivalDeparture.swift:305` — rather than invented, and the assumption, its provenance and its failure mode are documented at the constant. Calling it out here per that rule.

## Testing

- `ScheduleDeviationTest` (new, 8 tests) — band boundaries including the half-open edges at exactly ±90 s, the not-real-time case, and that all four states have distinct colors in both tiers (the agencyY collision, as a test).
- `ArrivalInfoTest` (+7) — the arrivals path end to end at ±89/±91 s, the headline "60 s late is green / 120 s late is blue" case, a 1-second deviation straddling a minute boundary (the exact shape the old flooring got wrong), and `predicted == false`.
- `TripLogBuilderTest` (+5) — the same boundaries through the trip-planner path, plus that a Late/Early chip can never read "0 min".
- `VehicleIconAllocationTest` — the early-vs-late key assertion is replaced by a liveness one, plus a new test pinning that deviation must **not** affect the icon.
- `./gradlew testObaGoogleDebugUnitTest` green; `compileObaGoogleDebugKotlin` + `compileObaMaplibreDebugKotlin` + `compileObaGoogleDebugAndroidTestKotlin` clean under `-PwarningsAsErrors=true`; `spotlessCheck` clean; resources process for all four brands.

Not run locally: `lintObaGoogleDebug` (per CLAUDE.md, left to CI) and the instrumented suite.

## Not done

On-device visual verification of the light/dark swatches and the route-colored markers — the test phone is shared, so I left it alone. Worth a look before merge, particularly the ETA strip in dark mode and a route whose feed omits `route.color` (should fall back to brand).

## Summary by CodeRabbit

* **New Features**
  * Centralized schedule-deviation styling with separate foreground and filled “chip” tiers across arrivals, trip details, real-time chips, and help legends.
  * Vehicle marker icons now use resolved route-color contrast for clearer realtime vs scheduled visuals.
* **Bug Fixes**
  * Improved vehicle icon caching so icons update correctly across renderer contexts, and ensured schedule-deviation changes don’t incorrectly alter icons.
* **Documentation**
  * Updated rebranding and help legend guidance to discourage overriding `stop_info_*` timing colors and clarified on-time/early/late thresholds.
* **Tests**
  * Added/expanded schedule-deviation boundary/bucketing and icon allocation/cache tests, including prediction-off cases.

### DIFF
--- a/docs/REBRANDING.md
+++ b/docs/REBRANDING.md
@@ -122,10 +122,12 @@ All other branded strings use `%1$s` placeholders that automatically substitute 
     <!-- Primary color with a "df" alpha prefix (~87% opaque) -->
     <color name="tutorial_background">#dfYOUR_PRIMARY_COLOR</color>
     <color name="ic_launcher_background">#YOUR_PRIMARY_COLOR</color>
-    <!-- Keep on-time color green even if your theme isn't green -->
-    <color name="stop_info_ontime">#4CAF50</color>
 </resources>
 ```
+
+Do **not** override the `stop_info_*` colors. They encode schedule deviation (early / on time /
+late / no real-time), which is semantic rather than brand identity, and they are intentionally
+independent of `brand_color` so a brand hue can never collide two states into the same color.
 
 To override material 3 colors, please checkout the [theme builder](https://material-foundation.github.io/material-theme-builder).
 
--- a/onebusaway-android/flavors/README.md
+++ b/onebusaway-android/flavors/README.md
@@ -80,10 +80,12 @@ That's it! All other branded strings use `%1$s` placeholders that automatically 
     <!-- Primary color with a "df" alpha prefix (~87% opaque) -->
     <color name="tutorial_background">#dfYOUR_PRIMARY_COLOR</color>
     <color name="ic_launcher_background">#YOUR_PRIMARY_COLOR</color>
-    <!-- Keep on-time arrivals green even if your theme isn't green -->
-    <color name="stop_info_ontime">#4CAF50</color>
 </resources>
 ```
+
+Do **not** override the `stop_info_*` colors. They encode schedule deviation (early / on time /
+late / no real-time), which is semantic rather than brand identity, and they are intentionally
+independent of `brand_color` so a brand hue can never collide two states into the same color.
 To override material 3 colors, please checkout the [theme builder](https://material-foundation.github.io/material-theme-builder).
 
 #### do_not_translate.xml
--- a/onebusaway-android/src/agencyX/res/values/colors.xml
+++ b/onebusaway-android/src/agencyX/res/values/colors.xml
@@ -21,7 +21,8 @@
     <color name="theme_muted">#FFCAD2</color>
     <color name="theme_accent">#E53935</color>
 
-    <color name="stop_info_ontime">#4CAF50</color>
+    <!-- No stop_info_* override: the schedule-deviation palette is semantic, not brand identity,
+         and is no longer derived from brand_color (#2043). -->
 
     <!-- Tutorials -->
     <color name="tutorial_background">#dfF44336</color>
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
@@ -35,6 +35,7 @@ import org.onebusaway.android.api.contract.TripDetailsEntry
 import org.onebusaway.android.api.data.asRouteTrips
 import org.onebusaway.android.extrapolation.extrapolatedVehicles
 import org.onebusaway.android.map.googlemapsv2.BitmapDescriptorCache
+import org.onebusaway.android.map.render.DEFAULT_ROUTE_LINE_COLOR
 import org.onebusaway.android.map.render.VehicleBitmaps
 import org.onebusaway.android.map.render.VehicleMarker
 import org.onebusaway.android.mock.Resources
@@ -321,93 +322,62 @@ class VehicleIconAllocationTest {
     // --- The route display color on the disc (#2043) ---------------------------------------------
 
     /**
-     * The disc takes the route's **display** color, not its raw GTFS value: it goes through
-     * `routeLineColors`, the shared agency-color policy, which keeps the hue but caps chroma and
-     * re-tones for the active theme. Asserted as "not the raw int" because the whole point is that an
-     * agency's literal choice never reaches the map unmediated — a full-chroma near-black would
-     * otherwise render a disc that vanishes into the basemap.
+     * The disc is the color the map is currently *drawing that route's line* with, carried on the
+     * marker by the controller — not something re-derived here from the route's GTFS color. In
+     * stop-focus view `adjacencyRouteColors` gives each shown route a distinct synthesized hue so the
+     * lines can be told apart, and a vehicle wearing its agency's nominal color there would point at
+     * the wrong line.
      */
     @Test
-    fun discUsesTheRouteDisplayColorNotTheRawGtfsColor() {
-        val raw = 0xFF1B6EF3.toInt() // a vivid, fully-saturated agency blue
-        val response = fakeResponse(routeColor = raw)
-        val colors = VehicleBitmaps.markerColors(context, liveVehicle(), response)
+    fun discUsesTheCarriedRouteDisplayColor() {
+        val assigned = 0xFF1B6EF3.toInt()
+        val colors = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = assigned))
 
-        assertNotEquals("the raw GTFS int must not reach the disc", raw, colors.disc)
-        assertNotEquals("the glyph must contrast with the disc", colors.disc, colors.onDisc)
+        assertEquals("the disc is the route's drawn color, unmodified", assigned, colors.disc)
     }
 
-    /** Two different agency colors must still produce two different discs — toning is not flattening. */
+    /** A vehicle whose route has no color falls back exactly as an uncolored polyline does. */
+    @Test
+    fun absentRouteColorFallsBackLikeAnUncoloredLine() {
+        val colors = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = null))
+
+        assertEquals(DEFAULT_ROUTE_LINE_COLOR, colors.disc)
+    }
+
+    /** Two routes drawn in different colors must render two different discs. */
     @Test
     fun differentRouteColorsStayDistinct() {
-        val blue = VehicleBitmaps.markerColors(context, liveVehicle(), fakeResponse(0xFF1B6EF3.toInt()))
-        val orange = VehicleBitmaps.markerColors(context, liveVehicle(), fakeResponse(0xFFF37B1B.toInt()))
+        val blue = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = 0xFF1B6EF3.toInt()))
+        val orange = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = 0xFFF37B1B.toInt()))
 
         assertNotEquals(blue.disc, orange.disc)
     }
 
     /**
-     * An achromatic source (grey/black/white) has no hue to keep, and an absent one has nothing at
-     * all; both take the brand fallback rather than producing a grey disc that would read as the
-     * "no real-time" state.
+     * The glyph and arrow flip by the disc's luminance rather than by the theme, so a route drawn in a
+     * pale yellow doesn't get a white bus glyph washed out on it.
      */
     @Test
-    fun achromaticOrAbsentRouteColorTakesTheBrandFallback() {
-        val brand = VehicleBitmaps.markerColors(context, liveVehicle(), fakeResponse(routeColor = null))
-        val grey = VehicleBitmaps.markerColors(context, liveVehicle(), fakeResponse(0xFF888888.toInt()))
+    fun glyphFlipsToStayLegibleOnLightAndDarkDiscs() {
+        val onDark = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = 0xFF0B1F55.toInt()))
+        val onLight = VehicleBitmaps.markerColors(context, liveVehicle().copy(routeColor = 0xFFFFEB3B.toInt()))
 
-        assertEquals("an achromatic route color falls back like an absent one", brand.disc, grey.disc)
-        assertNotEquals(
-            "the fallback must not be the not-real-time grey",
-            VehicleBitmaps.markerColors(context, staleVehicle(), fakeResponse(null)).disc,
-            brand.disc
-        )
+        assertEquals("white glyph on a dark disc", android.graphics.Color.WHITE, onDark.onDisc)
+        assertEquals("black glyph on a light disc", android.graphics.Color.BLACK, onLight.onDisc)
     }
 
-    /** A vehicle without real-time is grey regardless of how colorful its route is. */
+    /** A vehicle without real-time is grey regardless of how its route is drawn. */
     @Test
     fun withoutRealtimeTheDiscIsGrey() {
-        val stale = VehicleBitmaps.markerColors(context, staleVehicle(), fakeResponse(0xFF1B6EF3.toInt()))
-        val live = VehicleBitmaps.markerColors(context, liveVehicle(), fakeResponse(0xFF1B6EF3.toInt()))
+        val route = 0xFF1B6EF3.toInt()
+        val stale = VehicleBitmaps.markerColors(context, staleVehicle().copy(routeColor = route))
 
-        assertNotEquals(live.disc, stale.disc)
+        assertNotEquals("liveness must still change the disc", route, stale.disc)
     }
 
     private fun liveVehicle(): VehicleMarker = withLiveness(vehicles(response()).first(), isRealtime = true)
 
     private fun staleVehicle(): VehicleMarker = withLiveness(vehicles(response()).first(), isRealtime = false)
-
-    /** A references pool whose single route carries [routeColor], for the display-color assertions. */
-    private fun fakeResponse(routeColor: Int?): RouteTrips {
-        val vehicleTripId = vehicles(response()).first().activeTripId
-        return object : RouteTrips {
-            override val trips: List<ObaTripDetails> = emptyList()
-            override fun trip(tripId: String?): ObaTrip? = FakeTrip(vehicleTripId, "routeA").takeIf { !tripId.isNullOrEmpty() }
-            override fun route(routeId: String): ObaRoute = FakeRoute(routeColor)
-            override val currentTimeMs: Long = 0L
-        }
-    }
-
-    private class FakeTrip(override val id: String, override val routeId: String) : ObaTrip {
-        override val shortName: String? = null
-        override val shapeId: String? = null
-        override val directionId: Int = 0
-        override val serviceId: String? = null
-        override val headsign: String? = null
-        override val timezone: String? = null
-        override val blockId: String? = null
-    }
-
-    private class FakeRoute(override val color: Int?) : ObaRoute {
-        override val id: String = "routeA"
-        override val type: Int = ObaRoute.TYPE_BUS
-        override val shortName: String? = null
-        override val longName: String? = null
-        override val description: String? = null
-        override val url: String? = null
-        override val textColor: Int? = null
-        override val agencyId: String = "agency"
-    }
 
     companion object {
         // Sweep the bearing through all 8 octants within a few frames, then revisit them (no new icons) —
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
@@ -357,7 +357,7 @@ class VehicleIconAllocationTest {
         val stale = staleVehicle().copy(routeColor = 0xFF1B6EF3.toInt())
 
         assertEquals(
-            context.getColor(ScheduleDeviation.Status.SCHEDULED.colorRes),
+            context.getColor(ScheduleDeviation.Status.SCHEDULED.displayColorRes),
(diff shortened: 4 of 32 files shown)


bmander:
Follow-up pushed (009e712a): the vehicle disc now uses the route's **display** color rather than its raw GTFS value.

It resolves through `routeLineColors` — the shared agency-color policy in `LineBadge.kt`, whose KDoc names "a spine, stroke or filled marker" as its target — so the hue is kept but chroma is capped and the color re-toned to a fixed accent tone for the active theme. An agency's literal choice never reaches the map unmediated, which matters because a near-black or near-white route color previously rendered a disc that vanished into the basemap.

Two consequences worth calling out in review:

- **The glyph and heading arrow are no longer hardcoded white.** The dark-theme tone is a *light* disc, so white would have disappeared on it. Both now take `RouteLineColors.onLine` (black in dark, white in light) — which is why that helper returns the pair together. The bitmap cache key gains the glyph color, since the theme flips it independently of the hue.
- **The disc no longer matches the route polyline.** Polylines, route badges and continuation arrows still draw from `RoutePolyline.resolvedColor` (raw GTFS / adjacency hue / `DEFAULT_ROUTE_LINE_COLOR`). So on the route map the vehicle and the line it rides on can differ in tone. Making the whole map adopt the toned policy is the consistent end state, but it changes every route line on the map and belongs in its own issue rather than riding along with #2043.

An achromatic (grey/black/white) or absent route color has no hue to keep and takes the brand fallback — deliberately *not* the "no real-time" grey, so a colorless route can't be mistaken for a vehicle with no prediction.

Verified on a physical Pixel 7 Pro: `VehicleIconAllocationTest` 9/9, including four new tests covering the raw-int-never-reaches-the-disc contract, distinct routes staying distinct through the toning, the achromatic/absent fallback, and the not-real-time grey. Unit suite, both platform flavors under `-PwarningsAsErrors=true`, and `spotlessCheck` all clean.

bmander:
Addressed the CodeRabbit findings across all three review rounds in `21a637ba`.

## Fixed

**Interline vehicles resolved their color from the wrong poll** (round 2, `RouteMapController:431`) — the real bug in the set, and one I introduced. `sampleVehicles` merges each extra route's own response for a cross-route interline (#2000), but `toMarker` got the *leader's*, so an extra route's vehicle couldn't resolve its trip and fell back to default blue — precisely the case route-coloring exists to serve. Each vehicle is now paired with its originating poll. The color memo had to key off the extra-poll map too: alternating between two responses would otherwise clear it on every call and silently undo the memoization.

**`legibleOn` didn't do what its own doc claimed** (round 3, `MarkerRendering:44-58`) — it promised "whichever contrasts better" but cut over at a hand-picked gamma-space luminance, which keeps white on a wide band of mid-tones black reads better on. Now compares real WCAG ratios via `ColorUtils.calculateContrast`. Note this also changes continuation-badge text on mid-tone route colors, which now flips to black where it stayed white — a pre-existing rendering change riding along, called out here so it isn't a surprise.

**Label/color disagreement** (round 1, `ArrivalInfo:217-224` + `VehicleInfoWindow:95-98`) — a vehicle 65 s late read "1 min late" in on-time green. I'd previously deferred this as out of scope, but the review is right that this PR *introduced* it: before, text and color shared one rule. Both now come from a single `ScheduleDeviation.status` call, magnitude worded by a shared `roundedMinutes` — rounding not truncation, so anything outside the band words as at least "2 min" and can't read as inside it. `TripLogBuilder` had its own copy of that rounding and now shares it; the epoch-minute floors in `ArrivalInfo` are gone entirely.

**Stale descriptor-cache rationale** (round 1, `GoogleMapRenderer:822`) and the legend's "late" copy, which said "more than a minute and a half" where the band is half-open at exactly 1.5.

## Skipped, with reasons

**Foreground contrast against `md_theme_surface`** (round 1, `colors.xml:65-69`). Correct measurement — 2.95–4.02:1 — but this is the deliberate tradeoff in adopting the iOS hexes verbatim, chosen by the maintainer over an AA-compliant variant, and already documented in the PR description. The fix as proposed would abandon literal parity, which is the point of the issue. Worth revisiting as a follow-up if parity turns out to matter less than AA.

**Exact ±90 s boundary assertions in `ArrivalInfoTest` and `TripLogBuilderTest`** (rounds 1 and 3). `ScheduleDeviationTest.the band edges match the iOS half-open comparison` already pins both edges, and both downstream paths delegate their bucketing to that one call — they have no comparison of their own to regress. Re-adding the edges there was duplication I'd deliberately removed in the cleanup pass. Instead I added a test for the genuinely new risk: that the *worded* magnitude can never contradict the bucket it came from.

**Legend on-time wording** ("up to 1½ early / less than 1½ late"). Accurate, but the half-open asymmetry is below what a user-facing legend can usefully convey; "within a minute and a half, either way" is truer to how it reads. I did take the "late" boundary fix.

**Docstring coverage 65.96%** — the gap is Compose `@Preview` functions and small private helpers whose names carry their meaning; the substantive additions are documented.

Verified: unit suite green, both platform flavors and androidTest clean under `-PwarningsAsErrors=true`, `spotlessCheck` clean, `VehicleIconAllocationTest` 9/9 on a physical Pixel 7 Pro.

bmander (commented):
Reviewed the current head (`21a637ba`). The latest commit addresses the interline route-color resolution, arrival/info-window label agreement, and marker foreground selection findings from my initial review. One blocking issue remains:

**[P1] Keep small status-chip text at accessible contrast**

`RealtimeChip` uses the shared foreground colors for `labelMedium` text over a 14%-tinted version of the same color. In the light theme, the resulting text/background contrast is approximately 2.72:1–3.20:1 across the four status colors, below the 4.5:1 requirement for small text and a regression from the retired contrast-safe `trip_realtime_*` palette.

Please retain a separate contrast-safe text tier for this chip (and the directions time span), or darken the light-theme foreground resources while preserving the iOS display colors separately for large text.

Relevant code:
- `onebusaway-android/src/main/java/org/onebusaway/android/ui/tripresults/TripResultsScreen.kt:1166-1194`
- `onebusaway-android/src/main/res/values/colors.xml:66-69`

All reported GitHub checks currently pass.

bmander:
Fixed in `628c28c7`. Your measurement reproduced exactly (2.72–3.20:1), as did the regression claim — the retired `trip_realtime_*` values came out at 5.08–5.35:1 on that same tint, so they had indeed been picked for it.

I took the second of your two suggested routes: keep the iOS values for large text, add a contrast-safe tier for small text. The palette now has **three** tiers, because the same state genuinely gets drawn against three grounds with three different floors:

| tier | ground | floor | light | dark |
|---|---|---|---|---|
| `stop_info_*` (display) | large text / graphics — the 30sp ETA | 3:1 | 3.11–3.83 | 5.10–9.20 |
| `stop_info_*_text` (new) | small text on the surface — this chip, the directions span | 4.5:1 | 4.62–4.72 on the chip's tint, 5.69–5.79 bare | 4.64–7.20 / 5.45–9.20 |
| `stop_info_*_fill` | filled surface under white text — pills, badges | 4.5:1 | 4.79–4.83 white-on | (mode-independent) |

I first tried to avoid a third tier by reusing the fill tier as the chip's text color, since it's already darkened — but it measures 3.75–3.86:1 on the chip's tint in light and 3.39–3.48:1 in dark. Different requirement (contrast against *white* vs against the *surface*), so it can't double up. Hence a real third tier rather than a rename.

**One case your review didn't cover:** measuring the dark theme too, the iOS dark blue `#0A84FF` lands at **4.39:1** on the chip's tint — also below 4.5. So `delayed` is nudged to `#0A8BFF` for the dark text tier. The other three dark values already clear it and simply alias the display tier, so `values-night` states only the one real delta.

Both `ScheduleDeviation.Status` and the tests now carry all three tiers, with a test asserting each state keeps its three resources distinct — so a surface can't silently pick up a color that was never checked against its ground.

The large-text ETA keeps the literal iOS hexes, which is where the parity in #2043 is actually visible.

Verified: unit suite green, both flavors and androidTest clean under `-PwarningsAsErrors=true`, `spotlessCheck` clean, `VehicleIconAllocationTest` 9/9 on a physical Pixel 7 Pro.

PR #2061, opened 2026-07-28
Eliminate remaining Java sources

bmander:
## Summary

- convert every remaining production, flavor, unit-test, and instrumentation-test Java source in `:onebusaway-android` to Kotlin
- remove migration-only JVM interop annotations and every forced non-null operator now that project callers are Kotlin
- remove Java Spotless configuration and add a `verifyNoJavaSources` gate to `check`

## Stack

This PR is stacked on #2060 and should be reviewed and merged after it.

## Validation

- `./gradlew spotlessCheck test check -PwarningsAsErrors=true`
- `./gradlew assembleObaGoogleDebug assembleObaMaplibreDebug -PwarningsAsErrors=true`
- `ANDROID_SERIAL=36141FDH3000J2 ./gradlew connectedObaGoogleDebugAndroidTest -PwarningsAsErrors=true` (157 tests)
- `verifyNoJavaSources` passes with zero `*.java` files under `onebusaway-android/src`
- zero forced non-null operators and zero migration-only JVM interop annotations

The sole remaining `@JvmField` is Android's required static `Parcelable.CREATOR` contract, not a compatibility bridge for Java callers in this project.

## Summary by CodeRabbit

* **New Features**
  * Improved donation reminders and donation-page navigation.
  * Added Kotlin-based backup and restore flows for app preferences, including file picker support.
* **Bug Fixes**
  * Improved location handling so missing permissions no longer interrupt location polling.
  * Improved test reliability for unavailable device locations.
* **Chores**
  * Updated build verification to prevent Java source files from being introduced.


bmander:
Addressed the review in d55b7d03. Triage below.

### Fixed

- **`LoaderTestCase` unbounded wait / swallowed interrupt** — `queue.take()` → `poll(30s)` with an explicit `AssertionError` on timeout, and `Thread.currentThread().interrupt()` before rethrowing.
- **`VehicleIconAllocationTest`** — collapsed the `assertTrue(… != null)` + `requireNotNull` pairing into a single `requireNotNull` with a message. Note the *compile-error* claim was incorrect: `kotlin.requireNotNull` carries a `returns() implies (value != null)` contract, so `vehicle` (a stable `val`) smart-casts on the following line. `compileObaGoogleDebugAndroidTestKotlin` passed both before and after. The cleanup is still worth it for readability.
- **`BackupUtils` `"backup.bin"`** — now uses `Backup.FILE_NAME`. Pre-existing in the Java original, and cosmetic in practice (`EXTRA_TITLE` is ignored for `ACTION_OPEN_DOCUMENT`), but there's no reason to keep the two out of sync.
- **`ServiceUtils` heuristics** — documented the keyword matching and the two-keyword stop-ID rule, including why no exact source exists (Open311 catalogs are city-operated and carry no "this is a transit stop issue" field) and what happens when they misfire.
- **`GtfsAlertsHelper` seconds→millis** — kept the conversion in place but documented it. These alerts come off a **raw GTFS-realtime feed**, where `active_period.start` is POSIX seconds by spec — a fixed unit, not a magnitude guess. That's a different case from the polymorphic OBA `situation` field that `situationEpochToMillis` exists for, and there is no wire→domain adapter on this path to move it to. Also named the constant.

### Not changed

- **`MockRegion` shared `id = 0`** — the observation about `Region.equals` being id-only is right, but the Java original also passed `0` for every synthetic region, and the only two live callers (`isRegionUsable`) never compare regions for equality. The suggested negative-ID convention would be actively wrong here: these fixtures set `custom = false` and stand in for directory regions, not rider-added ones. Picking an ID scheme with no consumer needing one is speculative.
- **`UIUtilTest` doesn't restore global state** — legitimate, but pre-existing; the Java had no `@After` either. Out of scope for a conversion PR.
- **`DonationsManager` nullable Hilt deps** — the premise is incomplete. `DonationViewModelTest.kt:45` deliberately constructs `DonationsManager(null, null)`, and the JVM unit-test classpath has only `junit` + `kotlinx-coroutines-test` — no mocking library. The nullability is load-bearing today; removing it means adding a mocking dependency or reworking the ViewModel seam, which belongs in its own change.

### Validation

- `./gradlew :onebusaway-android:compileObaGoogleDebugKotlin -PwarningsAsErrors=true`
- `./gradlew :onebusaway-android:compileObaGoogleDebugAndroidTestKotlin :onebusaway-android:compileObaGoogleDebugUnitTestKotlin -PwarningsAsErrors=true`
- `./gradlew spotlessCheck`

### Follow-up worth noting

`LoaderTestCase` is entirely unreferenced — no subclasses, no callers anywhere in the tree. I hardened it as reviewed, but deleting it would be the better outcome; happy to do that instead if you'd prefer.

Issue #2104, opened 2026-07-31
Directions vehicle filter mishandles multi-variant route polylines


bmander:
## Summary

When route focus is entered from a directions ETA pill, the selected-leg vehicle filter can remove a valid upstream vehicle if the route direction contains multiple polyline variants.

`boundedThrough()` currently treats `List<RoutePolyline>` as consecutive, travel-ordered segments. The OBA `stops-for-route` response can instead supply multiple valid shape variants for one direction. Bounding against only the polyline nearest the alight point can therefore exclude a vehicle traveling on another variant, even when that vehicle is actively serving the tapped trip and has a positive ETA at the boarding stop.

## Reproduction observed

- Directions leg: `3rd Ave & Virginia St` (`1_420`) → `3rd Ave & Pine St` (`1_430`)
- Interchangeable route: route 2 (`1_100089`), direction `0`
- ETA: approximately 14 minutes
- Arrival trip: `1_788758740`
- The arrival response reported:
  - `activeTripId == tripId`
  - `predicted == true`
  - a current GPS-backed location
- The route-vehicle poll for `1_100089` contained the same active trip and position.
- Nevertheless, tapping the pin-marked ETA pill displayed/focused no vehicle.

The route-2 direction group contained three polylines. The selected-leg geometry filter removed the active trip before `tryFocusVehicle()` could resolve and frame it.

## Expected behavior

A vehicle actively serving a trip with a positive ETA at the boarding stop must be considered upstream and remain eligible for directions route focus. Multiple route variants must not be interpreted as consecutive polyline fragments.

## Actual behavior

`List<RoutePolyline>.boundedThrough()` selects one nearest polyline and treats earlier list entries as upstream fragments. A legitimate vehicle on another route variant can be omitted from `focusedVehiclePathsByRoute`, causing a valid pin-marked ETA tap to do nothing.

## Relevant code

- `RouteSegmentHighlight.kt`: `boundedThrough()` / `containsRoutePoint()`
- `RouteMapController.kt`: `focusedVehiclePathsByRoute` and `filterToFocusedRide()`

## Current narrow mitigation

The issue-2099 working branch preserves the explicitly requested `pendingFocus` trip through the geometry filter. This restores the ETA-pill interaction using exact trip identity, but it does not correct the underlying multi-variant geometry model for other vehicles.

## Suggested direction

[...]

PR #2105, opened 2026-07-31
Interleave interchangeable routes in the directions ETA strip

### LINKED ISSUE #2099: Interleave interchangeable routes in the ETA strip in the directions drawer
Instead of showing an ETA strip for each interchangeable route, combine them all into one ETA strip. This will involve putting the route small badge inside of the ETA pill.

bmander:
## Summary

- combine the planned and interchangeable routes into one chronologically interleaved ETA strip in the directions drawer
- place each route’s compact badge inside its ETA pill
- preserve stable ordering for equal ETAs and unique Compose keys across route groups
- carry the planned route badge explicitly so same-named alternatives cannot erase its color
- preserve an explicitly tapped ETA trip through the selected-leg geometry filter so a valid pinned vehicle can be focused

Closes #2099.

## Why

Separate ETA strips make riders compare multiple timelines to determine which interchangeable route arrives first. One time-ordered strip makes that choice immediate while the badge inside every pill preserves route identity.

During device verification, a valid route-2 vehicle was present in both the arrivals response and route-vehicle poll but was removed by the directions selected-leg geometry filter before focus. The narrow safeguard in this PR trusts the explicitly requested active trip identity. The underlying multi-variant route-polyline modeling issue is tracked separately in #2104.

## User impact

Directions for interchangeable routes now present a single “take whichever comes first” ETA timeline. Tapping a pin-marked pill continues into route focus and can frame the requested live vehicle even when reconstructed route-wide geometry does not match that trip variant.

## Validation

- `./gradlew spotlessCheck`
- `./gradlew :onebusaway-android:testObaGoogleDebugUnitTest`
- `./gradlew :onebusaway-android:assembleObaGoogleDebugAndroidTest`
- `git diff --check`
- installed and manually exercised the debug APK on the shared physical phone

## Summary by CodeRabbit

* **New Features**
  * ETA arrival strips now combine planned and alternative routes in chronological order.
  * Individual ETA pills display route badges and colors for clearer route identification.
  * Focused vehicles remain visible even when temporarily outside planned route geometry.

* **Bug Fixes**
  * Improved handling of route-specific badges and vehicle filtering.
  * Prevented display conflicts when routes share names or arrival times.

* **Tests**
  * Added coverage for route interleaving, badge rendering, ordering, and focused vehicle visibility.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/arrivals/EtaStripRenderTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/arrivals/EtaStripRenderTest.kt
@@ -26,6 +26,7 @@ import org.onebusaway.android.SmokeTest
 import org.onebusaway.android.ui.arrivals.components.EtaStrip
 import org.onebusaway.android.ui.arrivals.components.previewArrival
 import org.onebusaway.android.ui.arrivals.components.previewRowCallbacks
+import org.onebusaway.android.ui.compose.components.RouteBadge
 import org.onebusaway.android.ui.compose.createUnconfinedComposeRule
 
 /**
@@ -54,7 +55,10 @@ class EtaStripRenderTest {
                         previewArrival("40", "Northgate", etaMinutes = 12, tripId = "trip_3")
                     ),
                     actionsFor = { null },
-                    callbacks = previewRowCallbacks()
+                    callbacks = previewRowCallbacks(),
+                    routeBadgeFor = {
+                        RouteBadge(it.shortName ?: error("preview route must have a name"), null)
+                    }
                 )
             }
         }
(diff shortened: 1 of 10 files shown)


bmander:
Addressed in 96c7a495e — the finding was valid on both counts.

`showDirectionPolylines()` ends in `publishVehicleSet()`, which reads both focus fields: `filterToFocusedRide` reads `focusExemptTripId` and `tryFocusVehicle` reads `pendingFocus`. Because that ran at the top of `reframe` while the fields were still updated at the bottom, the publish applied the **new** leg's geometry filter against the **old** focus state:

- the stale exemption admitted the previous leg's vehicle into the pushed marker set, and since that is the discrete reconcile, the marker stood until the next poll republished;
- a still-unresolved `pendingFocus` could fit the camera onto that same vehicle. This one predates the exemption — `reframe` never cleared `pendingFocus` — so thanks for catching it in passing.

Both fields are now cleared at the top of `reframe`, before the segment refresh. The separate `focusExemptTripId = request.focusTripId` assignment is gone: `requestFocus()` already sets both, so it stays the single place a focus is installed, and a request carrying no `focusTripId` leaves focus cleared — which matches the unconditional set `start()` does with the same field.

Not covered by a regression test: the invariant is controller state ordering rather than pure logic, and there is no `RouteMapController` test harness in the repo (the existing tests cover `focusedRideKeepsVehicle` as a pure function). Verified with `spotlessCheck`, `compileObaGoogleDebugKotlin -PwarningsAsErrors=true`, and the full `testObaGoogleDebugUnitTest` suite.

Issue #2102, opened 2026-07-31
Shrink directions-view route badges at high zoom


bmander:
Shrink directions-view route badges at high zoom, linearly from a low zoom scale to a high zoom scale. Start by using the shrink schedule from other shrinkable UI items, to be tuned.

PR #2106, opened 2026-07-31
Make the directions map's route badges focus their ride (#2101)

### LINKED ISSUE #2101: Make route badges clickable
Make route badges in the directions view clickable. Clicking on them focuses the corresponding leg.

bmander:
Closes #2101.

The route label drawn on an itinerary's transit line named the ride but led nowhere — it was built with no tap target at all, deliberately, so a stray tap could not trade the whole trip the rider was reading for one route's map. That left the most direct thing on screen inert: the rider sees the `[40]` sitting on the line and reaches for it.

A label now leads somewhere, and somewhere is **inside** the trip: it focuses the ride it names, exactly as tapping that ride's row in the directions drawer does. Both spend the same `focusTransit`, so one ride can't mean two things depending on where the rider reached for it — route focus with the ridden segment drawn over it, degrading to leg framing and then to the board stop just as the row does.

### How it fits together

- **`RouteBadge.tap` widens** from a `RouteDirectionKey?` to a sealed `RouteBadgeTap`. The two map views that label their lines now want different things of a tap: adjacency (#1827) opens a route the rider hasn't opened, directions (#2101) stays in the trip. Only the `ShowRoute` form carries the "a navigable label names exactly one route" invariant — an interchangeable ride is one ride however many routes it may be taken on.
- **`ObaMapCallbacks.onRouteBadgeClick` takes the badge whole**, as `onStopClick` already takes a `StopMarker`, so the `when` over that sealed type lives once in `MapFeature` rather than mirrored in both flavor adapters. Both adapters end up *simpler* than before this change.
- **A label is per route; a ride is what the rider boards once** — so the two don't always cover the same legs. A stay-aboard interline onto another route (#2000) is one ride wearing a label per route; an itinerary that rides one route twice is two rides under a single label. The label therefore carries the leg indices it knows, and `rideCoveringLegs` resolves them against the drawer's own entries — the side that has the rides.
- **The tap crosses map → drawer**, the opposite way from `mapDirectives`, and can't be a directive: resolving a ride needs the results VM's leg refs. `HomeViewModel` relays it as a `SharedFlow` the results sheet collects. A tap with no collector is a tap with no trip to focus into, so dropping it is right where queueing it would not be — notably, a badge tap while the map-point picker is up (itinerary still drawn, sheet not composed) correctly does nothing.

### Reviewer's attention, please

One judgment call worth a second opinion: where a single label covers **two separate rides of the same route**, it resolves to the first in travel order. One label is one tap target, so it can only lead somewhere once, and the earlier ride is the one the rider reaches first — but it is a tie-break, documented at `rideCoveringLegs` and pinned by a test, not a fact off the wire.

### Verification

- Compiles clean under `-PwarningsAsErrors=true` on `obaGoogle` and `obaMaplibre`, plus both test source sets.
- Full JVM unit suite passes. Updated `RouteBadgeInvariantsTest`, `ItineraryRouteBadgesTest`, `RouteViewGeometryTest`, `RouteMapPresentationPlanTest`; new `RideCoveringLegsTest` covers the interline, shared-label, empty and no-match cases.
- **Not yet exercised on a device.** Installed on a Pixel 7 Pro but not driven through a real plan, so the badge's hit area and the resulting focus are unconfirmed by eye. Worth a pass before merge — and worth doing alongside #2102, which changes badge sizing.

## Summary by CodeRabbit

* **New Features**
  * Route badges on itineraries are now tappable and focus the corresponding ride.
  * Route badges can open route details or focus specific itinerary legs, depending on their context.
  * Tapping a route badge now provides consistent behavior across supported map views.

* **Bug Fixes**
  * Improved handling of grouped, multi-leg, interchangeable, and non-transit itinerary routes.
  * Inert badges no longer trigger unintended navigation.

### DIFF
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/GoogleMapRenderer.kt
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/GoogleMapRenderer.kt
@@ -124,9 +124,9 @@ class GoogleMapRenderer(
     // the single selected vehicle), but kept as a map like the other *ByMarker lookups for symmetry.
     private val continuationBadgeByMarker = HashMap<Marker, ContinuationBadge>()
 
-    // Adjacency route badge tap targets (#1827), mirrored by the maplibre flavor (#1913). Their
-    // geographic anchors are laid out once upstream; these markers then move naturally with the map
-    // through pan and zoom.
+    // Route badge tap targets — adjacency (#1827) and a directions ride (#2101) — mirrored by the
+    // maplibre flavor (#1913). Their geographic anchors are laid out once upstream; these markers then
+    // move naturally with the map through pan and zoom.
     private val routeBadgeByMarker = HashMap<Marker, RouteBadge>()
 
     // The latest trips-for-route poll, published as it changes (after the markers are reconciled). The
--- a/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/compose/GoogleComposeAdapter.kt
+++ b/onebusaway-android/src/google/java/org/onebusaway/android/map/googlemapsv2/compose/GoogleComposeAdapter.kt
@@ -61,7 +61,6 @@ import org.onebusaway.android.map.compose.drivePings
 import org.onebusaway.android.map.googlemapsv2.GoogleMapRenderer
 import org.onebusaway.android.map.render.CameraSnapshot
 import org.onebusaway.android.map.render.MapProjector
-import org.onebusaway.android.map.render.RouteBadgeTap
 import org.onebusaway.android.map.render.ScreenOffset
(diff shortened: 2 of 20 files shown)


bmander:
Addressed the review comment in 74c8f8d5d.

The first-match-in-travel-order tie-break was documented at `rideCoveringLegs`, where it is implemented, but the four sites that depend on it each read as though a badge names exactly one ride. All four now say where a label covering the same route ridden twice leads, and defer to `rideCoveringLegs` for the reasoning rather than restating it.

Documentation only — no behaviour change, and the tie-break itself is unchanged and still pinned by `RideCoveringLegsTest`. It remains the judgment call flagged in the description and is still worth a second opinion on the merits.

Still not exercised on a device; that's the open item before merge.

Issue #2124, opened 2026-07-31
Route-focus vehicle eligibility re-projects every route variant at 20 Hz


bmander:
## Summary

`focusedRideKeepsVehicle` → `containsRoutePoint` runs on the **20 Hz per-frame** vehicle sampler, and each eligible path costs a full O(vertices) `Polyline.nearestProjection`. #2120 made the number of eligible paths larger and deterministic, so this pre-existing cost is now paid in full on every multi-variant route.

## Where

- `RouteMapController.setVehiclesSampler { nowMs -> sampleVehicles(WallTime(nowMs)) }` wires `sampleVehicles` as the render sampler.
- `sampleVehicles` → `filterToFocusedRide(routeId)` → `focusedRideKeepsVehicle` → `RouteSegmentHighlight.containsRoutePoint`.
- The same controller's `displayedRouteColor` KDoc already establishes the constraint: *"this sits on the **per-frame** sampler — `sampleVehicles` runs at 20 Hz on Google and the display rate on MapLibre, for every vehicle."* That method memoizes a hex-string parse because of it. A full geometric projection over every route variant is a much heavier thing to be doing on the same loop.

## Why #2120 changes the arithmetic

Before #2120, `boundedThrough` returned `matchIndex + 1` paths — order-dependent, averaging roughly `(N+1)/2` for `N` variants, and just 1 whenever the alight point happened to project onto the first entry. After #2120 it correctly returns *every* variant the alighting point sits on, which on a route whose variants share a downtown trunk is all of them (3 for the route 2 case in #2104). The off-every-variant fallback likewise returns all `N` unbounded where it previously returned `matchIndex + 1`.

That is the right answer — this issue is not asking to revert it — but it roughly doubles a per-frame cost that was never budgeted.

`containsRoutePoint`'s `any { }` short-circuits on the first matching path, so an **eligible** vehicle is cheap. A **rejected** vehicle pays the full `N × vertices` scan every frame — and rejecting downstream vehicles is precisely what this filter exists to do, so the expensive branch is the common one.

## Suggested fixes

1. **Bounding-box pre-reject per `BoundedRoutePath`.** Cache each path's lat/lon extent (padded by the tolerance) at construction in `boundedThrough`; a vehicle outside the box can't be within tolerance of the line, so it skips the projection entirely. Cheap, local to `RouteSegmentHighlight`, and kills most of the cost for the off-route rejects.
[...]

PR #2130, opened 2026-07-31
Pre-reject off-route vehicles with padded bounds (#2124)

bmander:
## What changed

- Cache a tolerance-padded geographic bounding box for each bounded route path.
- Reject clearly off-route vehicle samples before performing the full polyline projection.
- Preserve exact projection behavior for date-line-crossing shapes and callers using a non-default tolerance.
- Add coverage for every padded edge, outside-tolerance rejection, and custom tolerances.

## Why

The directions route-focus vehicle filter runs at 20 Hz for every vehicle. After the per-shape-variant eligibility work in #2120, rejected vehicles may require a full projection against every eligible path. The cached bounds make the common off-route rejection inexpensive while retaining the projection as the exact test.

## Validation

- `:onebusaway-android:testObaGoogleDebugUnitTest --tests org.onebusaway.android.map.RouteSegmentHighlightTest`
- `:onebusaway-android:testObaMaplibreDebugUnitTest --tests org.onebusaway.android.map.RouteSegmentHighlightTest`

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/map/RouteSegmentHighlight.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/map/RouteSegmentHighlight.kt
@@ -15,9 +15,12 @@
  */
 package org.onebusaway.android.map
 
+import kotlin.math.abs
+import kotlin.math.cos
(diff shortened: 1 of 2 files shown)


bmander:
Superseded by https://github.com/OneBusAway/onebusaway-android/pull/2132

Issue #2075, opened 2026-07-30
Tapping out of a walk or bike leg should return to route focus just like tapping out of a transit leg


bmander:
Currently in directions view, if a transit leg is highlighted, tapping off of it on the map will return focus to the whole directions itinerary. Tapping off of a bike or walk woute however will exit directions focus entirely. It should act like the transit leg, and refocus to the whole directions itinerary.

Issue #2097, opened 2026-07-30
Make stops unselectable in directions mode


bmander:
If the user is in directions mode and make him select a stop. This kicks them over into stop focus mode and if they swipe back into directions. Mode to stop remains selected. This is a confusing state and stop selection should be unrepresentable when in directions mode.

PR #2141, opened 2026-08-02
Confirm before backing out of a shown directions itinerary

### LINKED ISSUE #2140: Once an directions itinerary is shown on screen, confirm before backing out
It's far too easy to accidentally erase your directions state. Whenever a navigation event would result in the directions screen being closed, pop up a modal confirmation dialog.

bmander:
Fixes #2140.

A planned trip is several deliberate acts — two endpoints, a time, a mode, and the option picked out of the results — but the two gestures that discarded it were the two cheapest gestures on the screen:

- the system **Back** from the itinerary overview, and
- a stray **tap on the map background** (`unfocusMapOneLevel`, which from the plain overview drops straight out of directions).

Either one erased the whole plan, with no way back to it.

## What changed

Both gestures now stage a confirmation instead of leaving, whenever there is a trip actually **drawn on the map**. `HomeViewModel` owns the latch (`pendingDirectionsExit`) because the two gestures reach it from different places — `HomeScreen`'s `BackHandler` and the map's own click callback — and `HomeScreen` renders the modal off it. Confirming exits; cancelling, an outside tap, or Back all keep the trip.

The confirm button is the destructive one, so it reads **Discard** rather than "OK".

## What deliberately still leaves on the first gesture

- An **unplanned form** — nothing drawn, nothing to lose.
- A **drilled-into leg** (route or on-street) — that gesture steps back to the itinerary overview rather than out of directions, which #2075 established.

The gate is `shownItinerary`, which is now cleared when the trip leaves the map (`clearShownItineraryOnMap`) and on a fresh entry into directions, so it can't ask about a trip the rider can't see. It deliberately still survives the exit itself, since backing *into* the focus we just left has to put the trip back on the map.

`replaceFocus` drops a staged question whenever focus moves off directions, so an incoming deep link that takes the map doesn't leave a dialog behind answering for a focus that is gone. Externally-driven navigation (a notification/deep link) is not itself gated on a dialog — it shouldn't be interrupted by one.

## Tests

- `HomeViewModelTest` — both gestures ask before discarding a drawn trip and declining changes nothing; both leave outright with no trip drawn; a cleared trip stops guarding the way out; a fresh entry doesn't inherit the previous visit's trip; a deep link drops the staged question. The existing leg-drop tests now answer the confirmation on their last step.
- `DirectionsExitConfirmDialogTest` (new, instrumented) — both answers are offered and stay wired to opposite outcomes.

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation dialog when leaving directions with an active trip plan.
  * Users can cancel to keep directions or discard the plan to exit.
  * The prompt also supports Back presses and taps outside the dialog.
  * Improved handling when starting new trips, clearing itineraries, changing map focus, or selecting bike stations.

* **Tests**
  * Added coverage for confirmation, cancellation, dismissal, and immediate exits when no trip plan is present.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/home/directions/DirectionsExitConfirmDialogTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/home/directions/DirectionsExitConfirmDialogTest.kt
@@ -43,22 +43,15 @@ class DirectionsExitConfirmDialogTest {
     private val cancelLabel = context.getString(R.string.cancel)
 
     @Test
-    fun itNamesTheTripBeingDiscardedAndOffersBothAnswers() {
-        renderDialog()
+    fun itNamesTheTripAndOnlyDiscardConfirmsTheExit() {
+        var confirmed = 0
+        var dismissed = 0
+        renderDialog(onConfirm = { confirmed++ }, onDismiss = { dismissed++ })
 
         composeRule.onNodeWithText(context.getString(R.string.directions_exit_confirm_title))
             .assertIsDisplayed()
         composeRule.onNodeWithText(context.getString(R.string.directions_exit_confirm_message))
             .assertIsDisplayed()
-        composeRule.onNodeWithText(discardLabel).assertIsDisplayed()
-        composeRule.onNodeWithText(cancelLabel).assertIsDisplayed()
-    }
-
-    @Test
-    fun onlyDiscardConfirmsTheExit() {
-        var confirmed = 0
-        var dismissed = 0
-        renderDialog(onConfirm = { confirmed++ }, onDismiss = { dismissed++ })
 
         composeRule.onNodeWithText(cancelLabel).performClick()
         assertEquals(0, confirmed)
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/home/directions/DirectionsExitConfirmDialogTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/home/directions/DirectionsExitConfirmDialogTest.kt
@@ -1,0 +1,82 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


bmander:
Follow-up cleanup pass pushed (f775f735).

**The guard moved from the gesture to the transition.** `stageDirectionsExitConfirmation` now takes `(from, to)` and runs *after* each caller computes its target, instead of re-deriving "plain overview leaves directions" at the top of `unfocusMapOneLevel` — that rule was written twice, five lines apart, in two different functions, and would have drifted silently if a directions level were ever added.

**That enumeration found an unguarded exit.** `MapViewModel.showItinerary` starts the bike layer with `directions = true`, seeded from the itinerary's own dock ids — so the docks drawn during directions are *the trip's own*, and tapping one called `onBikeStationFocused` → `pushFocus(BikeStation)`, discarding the whole plan in a single tap on the exact screen this PR protects. It's now **refused** rather than confirmed, matching how a stop tap in directions is already refused (#2097): nothing is lost, so there is nothing to ask about. `MapFeature` leaves its own render state alone when the focus isn't taken, in the same order the stop tap uses.

Remaining ways `CurrentFocus.Directions` can be replaced, for the record: `revealStop` / `focusStandaloneRoute` from a pushed destination. Those are reachable only via chrome that's hidden in directions, and when they *do* fire they're externally driven (a notification or deep link), which shouldn't be interrupted by a modal — so they stay unguarded deliberately.

Smaller items from the same pass: `clearShownItineraryOnMap` now drops a staged question along with the trip it asked about, so the dialog can't outlive its reason; the `shownItinerary` comment no longer claims null means "no trip on the map" (it deliberately outlives the exit — the gate pairs it with the focus); the `confirmExitDirections` guard is documented as what keeps it from being a public door around `exitDirections` being private; `HomeScreen` collects the latch next to the dialog rather than ~200 lines above it; and the two instrumented dialog tests are fused into the one that asserts wiring.

Verified: full `testObaGoogleDebugUnitTest` green (one new case pinning the refused bike tap), both compile targets clean under `-PwarningsAsErrors=true`, and the instrumented dialog test re-run on device (1/1 OK).

PR #2146, opened 2026-08-02
Decode the no-prediction sentinel when the server adds slack to it (#2144)

### LINKED ISSUE #2144: Garbage ETA (-496029 hr) when the server sends the no-prediction sentinel plus slack time
## Symptom

The 1 Line at Capitol Hill northbound (`40_99603`) shows an ETA of **-496029 hr 21 min**.

## The record

```json
{
  "routeShortName": "1 Line",
  "predicted": true,
  "scheduledArrivalTime":   1785708570000,
  "predictedArrivalTime":           29999,
  "scheduledDepartureTime": 1785708600000,
  "predictedDepartureTime":         29999,
  "tripStatus": { "scheduleDeviation": -36 },
  "tripId": "40_LLR_2026-03-29_Jan29_Spring26_CLCrev_Sunday_100479_1058"
}
```

`29999 - 1785705908095 ≈ -496029 hours`. Every other field on the record is self-consistent — only
the two predicted instants are junk.

## Why the existing guard misses it

`ArrivalAdapters.predictedServerTimeOrNull` already decodes non-positive predicted instants to
"no prediction" — the #1687 / #46 fix for the `-1` and `0` sentinels. `29999` is positive, so it
passes straight through and is treated as an epoch-millis instant.

## Root cause

`29999` is not an error code. It is OBA's `-1` no-prediction sentinel **with the stop's slack time
added to it**, in `ArrivalAndDepartureServiceImpl` (~:726-742):

```java
long arrivalTime   = tpr.getTimepointPredictedArrivalTime();    // -1 = unset
long departureTime = tpr.getTimepointPredictedDepartureTime();  // -1 = unset
if (departureTime <= 0) {                       // fires on the sentinel...
    int slack = instance.getBlockStopTime().getStopTime().getSlackTime();   // 30 s
    departureTime = arrivalTime + slack * 1000; // ...then does arithmetic ON it: -1 + 30000 = 29999
}
setPredictedDepartureTimeForInstance(instance, departureTime);
if (arrivalTime == -1)
    setPredictedArrivalTimeForInstance(instance, departureTime);  // arrival := 29999 too
```

`TimepointPredictionRecord` defaults both predicted fields to `-1`. So the emitted value is exactly
`slackTime * 1000 - 1` — which is also why arrival and departure are identical, and why it ends in
`999`.

Filed upstream as OneBusAway/onebusaway-application-modules#PENDING.

## Client-side fix

`StopTimeEntryImpl.getSlackTime()` is defined as that stop time's own `departureTime - arrivalTime`
(seconds) — i.e. the scheduled dwell, which the client already has as
`scheduledDepartureTime - scheduledArrivalTime`. So the artifact can be **reconstructed exactly** and
compared for equality:

```
reject when  epochMs <= 0  ||  epochMs == (scheduledDepartureTime - scheduledArrivalTime) - 1
```

No threshold, no magnitude guess, and nothing involving the service date (a service-day guard was
considered and rejected — it misbehaves at day boundaries and for owl service). It cannot collide
with a real prediction: a dwell in milliseconds is ~1e4 while a genuine predicted instant is ~1.8e12.

## How widespread

Live sweep of Puget Sound — 79 stops, 1308 arrival rows, 2616 predicted fields. 29 rows match the
formula, and **nothing else absurd exists** in the sample:

| dwell at stop | artifact | rows | caught today? |
|---|---|---|---|
| 0 s | `-1` | 28 | yes, by the existing non-positive check |
| 30 s | `29999` | 1 | **no** |

So this has been firing constantly for years, but at zero-dwell stops it degenerates to `-1` and the
existing guard absorbs it. Only stops with a nonzero scheduled dwell leak — Link platforms dwell 30 s,
which is why it shows up there. A 60 s dwell would produce `59999`.

bmander:
Closes #2144.

The 1 Line at Capitol Hill northbound showed an ETA of **-496029 hr 21 min**.

## The record

```json
{
  "routeShortName": "1 Line",
  "predicted": true,
  "scheduledArrivalTime":   1785708570000,
  "predictedArrivalTime":           29999,
  "scheduledDepartureTime": 1785708600000,
  "predictedDepartureTime":         29999,
  "tripStatus": { "scheduleDeviation": -36 }
}
```

`29999 - 1785705908095 ≈ -496029 hours`. Every other field, `scheduleDeviation` included, is healthy —
only the two predicted instants are junk. The existing non-positive check (the #1687 fix for the bare
`-1` / `0` sentinels) can't see it, because `29999` is positive.

## Root cause

`29999` is not an error code. It is OBA's `-1` no-prediction sentinel **with the stop's slack time
added to it**, in `ArrivalAndDepartureServiceImpl`:

```java
long arrivalTime   = tpr.getTimepointPredictedArrivalTime();    // -1 = unset
long departureTime = tpr.getTimepointPredictedDepartureTime();  // -1 = unset
if (departureTime <= 0) {                       // fires on the sentinel...
    int slack = ...getStopTime().getSlackTime();                // 30 s
    departureTime = arrivalTime + slack * 1000; // ...then does arithmetic ON it: -1 + 30000 = 29999
}
setPredictedDepartureTimeForInstance(instance, departureTime);
if (arrivalTime == -1)
    setPredictedArrivalTimeForInstance(instance, departureTime);  // arrival := 29999 too
```

`TimepointPredictionRecord` defaults both fields to `-1`, so the emitted value is exactly
`slackTime * 1000 - 1` — which is also why arrival and departure are identical, and why it ends in
`999`. Filed upstream as OneBusAway/onebusaway-application-modules#474.

## The fix is exact reconstruction, not a magnitude test

This is worth stating plainly, because the obvious fix here would be a threshold and the codebase
forbids one.

`StopTimeEntryImpl.getSlackTime()` is defined as that stop time's own `departureTime - arrivalTime`,
so the server's slack **is** this arrival's scheduled dwell — a value already on the wire. We compute
what the server would have produced and compare for equality:

```kotlin
internal fun predictedServerTimeOrNull(epochMs: Long, scheduledDwellMs: Long): ServerTime? = epochMs
    .takeIf { it > 0L && it != scheduledDwellMs - 1L }
    ?.let { ServerTime(it) }
```

No threshold, no guess about how small a number is "too small", and nothing involving the service
date — a service-day guard was considered and **rejected**, because it misbehaves at day boundaries
and for owl service. It cannot collide with a real prediction: a dwell in milliseconds is ~1e4 while a
genuine predicted instant is ~1.8e12.

A test pins the distinction: `29999` at a stop with a **60 s** dwell is *kept*, because it isn't what
that stop's slack arithmetic would have produced. The check has no suspicious range.

## How widespread

Live sweep of Puget Sound — 79 stops, 1308 arrival rows, 2616 predicted fields. 29 rows match the
formula, and nothing else absurd exists in the sample:

| dwell at stop | artifact | rows | caught before this PR? |
|---|---|---|---|
| 0 s | `-1` | 28 | yes, by the non-positive check |
| 30 s | `29999` | 1 | **no** |

So this has been firing constantly for years, but at zero-dwell stops it degenerates to `-1` and the
existing guard absorbs it. Only stops with a nonzero scheduled dwell leak — Link platforms dwell 30 s,
which is why it surfaces there. A 60 s dwell would produce `59999`.

## Behaviour after the fix

The row falls back to its scheduled time and reads "Scheduled". That works through machinery that
already exists — `predicted` is derived from "has a usable instant", not the raw wire flag — so no new
UI states.

Deliberately **not** repairing the value from `scheduled + scheduleDeviation`, even though this
record's deviation is sane: earlier work on the WSF stale-departure bug showed that always-deriving
degrades healthy Path-A predictions, and inventing a number the server didn't send is a larger call
than reporting absence.

## Testing

Six new `ArrivalAdaptersTest` cases built from the live records — the bare sentinel, the slack-added
artifact at 30/60/90 s dwells, another stop's artifact correctly kept, and a real prediction from the
same response. Full unit suite green; compiles clean under `-PwarningsAsErrors=true`; spotless clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved arrival and departure prediction handling when no prediction is available.
  - Preserved valid predictions, including values resembling no-prediction markers for other stops.
  - Filtered out corrupt or invalid frequency arrivals instead of displaying them.
  - Corrected arrival availability reporting when all returned entries are invalid.

- **Tests**
  - Added coverage for standard and dwell-adjusted no-prediction values.
  - Verified valid predictions remain visible.
  - Added regression coverage for filtering corrupt frequency arrivals.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
@@ -38,16 +38,19 @@ internal fun ArrivalDeparture.asArrivalData(directionId: Int?): ArrivalData? = t
  * discard that row instead of inventing one from the frequency window.
  *
  * Requiring a frequency row and all four timestamps to agree keeps this tied to the server path that
- * collapses the schedule; the suffix check is an arithmetic signature, not a magnitude threshold.
+ * collapses the schedule. For the positive spelling, the arithmetic signature (`999` milliseconds)
+ * is not sufficient by itself because a real epoch can have that suffix; the candidate must also
+ * predate the frequency service window. We deliberately do not impose an upper bound because visits
+ * at downstream stops can occur after the trip-origin frequency window ends.
  */
 private val ArrivalDeparture.hasCollapsedFrequencySentinel: Boolean
     get() {
+        val frequencyStartTime = frequency?.startTime ?: return false
         val candidate = predictedArrivalTime
-        return frequency != null &&
-            predictedDepartureTime == candidate &&
+        return predictedDepartureTime == candidate &&
             scheduledArrivalTime == candidate &&
             scheduledDepartureTime == candidate &&
-            (candidate <= 0L || candidate % 1_000L == 999L)
+            (candidate <= 0L || (candidate < frequencyStartTime && candidate % 1_000L == 999L))
     }
 
 /**
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/adapters/ArrivalAdapters.kt
@@ -22,8 +22,33 @@ import org.onebusaway.android.models.Occupancy
 import org.onebusaway.android.models.Status
 import org.onebusaway.android.time.ServerTime
 
-/** Adapts a modernized [ArrivalDeparture] DTO (arrivals fetch) to the [ArrivalData] model. */
-internal fun ArrivalDeparture.asArrivalData(directionId: Int?): ArrivalData = DtoArrivalData(this, directionId)
+/**
+ * Adapts a modernized [ArrivalDeparture] DTO (arrivals fetch) to the [ArrivalData] model, or drops an
+ * unrecoverable frequency row whose scheduled and predicted times were all replaced by the server's
+ * no-prediction sentinel.
+ */
+internal fun ArrivalDeparture.asArrivalData(directionId: Int?): ArrivalData? = takeUnless { hasCollapsedFrequencySentinel }?.let { DtoArrivalData(it, directionId) }
+
+/**
+ * For a frequency instance, the server's predicted-time setters also overwrite the corresponding
+ * scheduled times. When both timepoint predictions are absent, that destroys the original scheduled
+ * dwell and leaves all four timestamps equal to either the bare sentinel (`-1` / `0`) or its
+ * `slack * 1000 - 1` spelling. The positive spelling always ends in `999`, because schedule slack is
+ * whole seconds. With neither a prediction nor a schedule left, there is no honest time to display;
+ * discard that row instead of inventing one from the frequency window.
+ *
+ * Requiring a frequency row and all four timestamps to agree keeps this tied to the server path that
+ * collapses the schedule; the suffix check is an arithmetic signature, not a magnitude threshold.
+ */
+private val ArrivalDeparture.hasCollapsedFrequencySentinel: Boolean
+    get() {
+        val candidate = predictedArrivalTime
+        return frequency != null &&
+            predictedDepartureTime == candidate &&
+            scheduledArrivalTime == candidate &&
+            scheduledDepartureTime == candidate &&
+            (candidate <= 0L || candidate % 1_000L == 999L)
+    }
 
 /**
  * Decodes a wire `predicted{Arrival,Departure}Time` into "the instant" or **`null`** ("no prediction"),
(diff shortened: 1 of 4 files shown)


bmander:
There is one remaining edge case in the client-side reconstruction: frequency-based arrivals can still leak the slack-adjusted sentinel.

`scheduledDwellMs` is derived from the scheduled timestamps on the wire, but `ArrivalAndDepartureServiceImpl.setPredicted{Arrival,Departure}TimeForInstance` overwrites the corresponding scheduled timestamps whenever `instance.getFrequency() != null`. When both timepoint predictions are unset, the server therefore writes `slack * 1000 - 1` into all four timestamps. The wire-level dwell becomes zero, so this decoder compares (for example) `29999` against `0 - 1` and retains the bad epoch.

An adapter-level test with a frequency payload whose four timestamps are all `29999` should reproduce it. The decoder needs a safe fallback for that server representation (or the PR should explicitly scope the workaround to non-frequency arrivals pending the upstream fix).

PR #2171, opened 2026-08-05
Resolve a planned leg's stop against its route's own stops (#2170)

### LINKED ISSUE #2170: Bug: ETA strip is absent for ST route 522
The ETA strip in a transit leg in the directions drawer is totally absent for sound transit route 522 in particular; it's not clear why.

bmander:
Fixes #2170.

## The bug

The ETA strip was **totally absent** on a Sound Transit 522 leg in the directions drawer — not even the "No upcoming arrivals" text, which is reserved for a stop that *was* looked up and had nothing.

The stop id the drawer asked OBA for did not exist, so arrivals came back as an error and `DirectionStopEtaStrip` returned early on anything that isn't `Content`.

## Root cause

`OtpObaIdResolver` mapped an OTP stop onto OBA by stamping the **leg route's** agency prefix onto the GTFS entity id, on the premise documented in its own KDoc: *"the stop is namespaced under the leg's route agency (that's the agency serving it here)"*.

That premise is false — **a route's agency does not own the stops it calls at.** Verified against the live Puget Sound deployments:

| | OTP | OBA |
|---|---|---|
| ST 522 route | `kcm:100232`, agency `kcm:40` | `40_100232` ✅ |
| its board stop | `kcm:23561` | `1_23561` — the app asked for `40_23561` → **404** |

Not unique to 522, just most visible there:

- Both Seattle streetcars have the same shape (OTP agency `kcm:23`, all `1_*` stops).
- One route can span **two** OBA agencies' stops — KCM 931 (`1_102558`) calls at both `1_*` and `29_*` — so no single prefix could ever have been right for the leg regardless. Its board stop even comes from a different OTP feed than its route (`CommTrans:2229` → `29_2229`).
- Wrong prefixes aren't always a clean 404: `40_620` and `1_620` are both real stops in different agencies, so the guess can silently name *someone else's* stop.

## Fix

Resolve the stop where its OBA agency is **stated** rather than guessed — the route's own OBA stop list.

- New `StopsForRouteRepository.routeStopIds` projection, off the **existing** cached/coalesced fetch (no new wire call shape).
- `OtpObaIdResolver.resolveLegs(legs)` returns each leg's route + end-stop ids in one pass, fetching the itinerary's distinct routes in one concurrent round and indexing each route's stops by entity id once. Matching uses OBA's id contract (split at the first `_`), so an entity id containing underscores survives.
- Route id resolution is **unchanged** — it was always correct.
- A stop that can't be resolved yields `null` and the caller degrades (stop still labelled and framed, just no arrivals board). Deliberately **no** fallback to the synthesized id: that fallback is the bug.

## ⚠️ Heuristic sign-off (CLAUDE.md)

Per *"No unsanctioned heuristics"*, reworking a pre-existing heuristic re-opens its sign-off. **Please review this paragraph specifically.**

This change **shrinks** the reverse-engineered rule's blast radius from two entity kinds to one: stops no longer go through it at all. What remains is `resolveAgency` for **routes** — derive the OBA agency from the OTP agency gtfsId suffix, verify it against agencies-with-coverage, else fall back to name matching.

The residual gap, now documented at the site: **membership is all the derived suffix is checked for, not identity.** A feed whose GTFS `agency_id` collided with a *different* OBA agency's id would resolve silently to the wrong agency. I simulated all 15 OTP agencies in the live Puget Sound graph — every derived hit's OBA name equals the OTP agency name, so it doesn't bite today, but nothing enforces it. Cheap hardening if wanted: require that name agreement before trusting the derived suffix. Left out as beyond this issue's scope.

The real fix remains an authoritative OTP-agency → OBA-agency map in the regions directory; when that lands, `resolveLegs` needs no change at all.

## Drive-by correction

The resolver's KDoc and its test both claimed OBA agency `97` is Skagit Transit, name-matched from OTP's UUID agency id — cited as the *justification* for the name-fallback rule. Live `agencies-with-coverage` says 97 is **Everett Transit**, and Skagit isn't a covered OBA agency at all. The test asserted `97_42` for a route that correctly resolves to `null`. Replaced with the real coverage list and the true behaviour (Skagit and Whatcom are in the OTP graph, not in OBA coverage → `null`).

## Cost of the change

This path is now network-bound where it used to be pure string work: one stops-for-route fetch per distinct transit route in the selected itinerary.

Worth being explicit that it is the **cold** fetch. The trip log builds before any leg can be tapped, so this warms the shared cache for the route focus that follows, never the reverse. That's why `resolveLegs` fans out concurrently — `SingleFlight` only coalesces *concurrent* callers, so a serial loop would have paid one round trip per leg.

The fetch includes polylines (the shared cache's superset). Measured gzipped, shapes are up to ~70% of the response — Link 1 Line is 9,965 B with vs 2,962 B without. Deliberately not split: a polyline-free variant would key a second cache entry per route and force a full re-fetch the instant the rider taps that leg, breaking the "each route fetched once, both projections from that one fetch" invariant the class is built around.

## Verification

- **Device-verified** on a Pixel 7 Pro: the 522 leg now shows its ETA strip.
- Data-level end-to-end: `arrivals-and-departures-for-stop/1_23561` lists route `40_100232` with headsigns "Bothell"/"Woodinville" — exactly what the strip's route-id + headsign matcher selects on.
- New JVM coverage for the 522 case, the two-agency route, cross-feed board stops, entity ids containing underscores, non-transit legs, each unresolvable path, and one-fetch-per-route.
- Compiles clean under `-PwarningsAsErrors=true`; unit tests and androidTest compile pass; Spotless applied.

## Notes for review

The third commit is a `/simplify` pass. It replaced a parallel `obaRouteIds` array (threaded into seven call sites, including cross-index pairings nothing checked) and a separate `prefetchRouteStops` — a public API whose result was discarded and which a caller could silently forget — with the single `resolveLegs` call. It also promotes the duplicated `StopsForRouteRepository` test stub to `testing/FakeStopsForRouteRepository`; the interface change had already forced a mechanical edit in the other copy, and the two had diverged on what an unstubbed route returns.

## Summary by CodeRabbit

* **Improvements**
  * Trip itinerary processing now resolves route stops in batches, reducing repeated lookups.
  * Improved stop matching across routes and agencies, including stop IDs containing underscores.
  * Trip results preserve stop display information even when an arrivals-board identifier cannot be resolved.
* **Bug Fixes**
  * Corrected route stop identification to use stops actually served by each route.
  * Improved handling of unresolved routes, stops, and non-transit legs.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/data/StopsForRouteRepository.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/data/StopsForRouteRepository.kt
@@ -72,12 +72,8 @@ interface StopsForRouteRepository {
     suspend fun routeMap(routeId: String): Result<RouteMapData?>
 
     /**
-     * Every OBA stop id the route serves — the whole references pool, not only the ids a direction
-     * group happens to list. This is the authoritative answer to "which OBA stop is this?" for a
-     * planned transit leg, whose stops OTP names in a different id space (see
-     * [org.onebusaway.android.directions.OtpObaIdResolver]). [Result.failure] on IO / HTTP / non-OK
-     * code **and** when there is no endpoint to contact — a caller that can't get the route's stops
-     * cannot name one of them, so there is nothing to degrade to.
+     * Every stop id the route serves — the whole references pool, not only the ids a direction group
+     * happens to list. Same failure contract as [routeStopGroups].
      */
     suspend fun routeStopIds(routeId: String): Result<List<String>>
 }
--- a/onebusaway-android/src/main/java/org/onebusaway/android/api/data/StopsForRouteRepository.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/api/data/StopsForRouteRepository.kt
@@ -51,6 +51,7 @@ import org.onebusaway.android.util.runCatchingCancellable
  * - [routeStopGroups] — the per-direction stop list (route info + a focused stop's trips).
  * - [routeMap] — the full route map: stops tagged with direction membership, selectable directions,
  *   and decoded shapes (the route overlay).
+ * - [routeStopIds] — just the ids of every stop the route serves.
  *
  * The wire DTO stays encapsulated here; callers only ever see the model projections.
  */
@@ -69,6 +70,16 @@ interface StopsForRouteRepository {
      * IO / HTTP / non-OK code, matching the legacy `MapDataSource.routeMap`.
      */
     suspend fun routeMap(routeId: String): Result<RouteMapData?>
+
+    /**
+     * Every OBA stop id the route serves — the whole references pool, not only the ids a direction
+     * group happens to list. This is the authoritative answer to "which OBA stop is this?" for a
+     * planned transit leg, whose stops OTP names in a different id space (see
+     * [org.onebusaway.android.directions.OtpObaIdResolver]). [Result.failure] on IO / HTTP / non-OK
+     * code **and** when there is no endpoint to contact — a caller that can't get the route's stops
+     * cannot name one of them, so there is nothing to degrade to.
+     */
+    suspend fun routeStopIds(routeId: String): Result<List<String>>
 }
 
 /**
@@ -113,6 +124,10 @@ class DefaultStopsForRouteRepository internal constructor(
         // runCatchingCancellable rethrows a CancellationException (raised by withContext if the caller is
         // cancelled mid-decode) rather than reporting the abandoned work as a failure.
         return runCatchingCancellable { withContext(Dispatchers.Default) { fetched.toRouteMapData(routeId) } }
+    }
+
+    override suspend fun routeStopIds(routeId: String): Result<List<String>> = entry(routeId).mapCatching { fetched ->
+        fetched?.references?.stops?.map { it.id } ?: throw noEndpoint()
     }
 
     /**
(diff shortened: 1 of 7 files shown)


bmander:
Addressed the remaining CodeRabbit item — the `stopIdCalls` thread-safety note from the second review, which landed as an "outside diff range" comment so it had no inline thread to resolve. Pushed as b5e7a700.

`stopIdCalls` is now a `CopyOnWriteArrayList`. One correction to the finding as written: it is **not** flaky today. `resolveLegs` does fan out with `async`, but those coroutines inherit `runTest`'s dispatcher, which confines them to the test thread — the appends are serialized in practice, and the two assertions on `stopIdCalls` pass deterministically. The fix is worth taking anyway because this is a *shared* fake: the first test that hands the resolver a genuinely multi-threaded dispatcher would lose entries and turn the fetch-count assertions flaky rather than fail loudly, and that failure would point at the resolver rather than at the recording. `CopyOnWriteArrayList` preserves call order, is safe to iterate while appends land (so `.toSet()` needs no external lock), and its copy-per-write cost is irrelevant at one entry per route.

Verified: `OtpObaIdResolverTest` (14), `StopsForRouteRepositoryTest` (5), `FocusedTripRepositoryTest` (4) all pass; compiled under `-PwarningsAsErrors=true`; `spotlessCheck` clean.

The two inline threads from the first review (ambiguous stop suffixes, the fake's unstubbed contract) were already fixed in 74349943 and are resolved. The **heuristic sign-off paragraph in the PR description still needs a human review** — that one isn't a CodeRabbit item and I can't close it myself.

PR #2178, opened 2026-08-06
Make the arrivals NOW pill reachable, and share its cutoff (#2177)

### LINKED ISSUE #2177: Arrivals ETA strip: the NOW pill is unreachable except at exactly 0
Noticed while reviewing the trip-tracking Live Update card (#2166 / PR #2173). Filing for later — not being worked now.

## What's wrong

The ETA strip's "NOW" pill is selected by an exact-zero test, so it is on screen for at most one minute per departure, and a bus that has just left renders a negative countdown instead.

`onebusaway-android/src/main/java/org/onebusaway/android/ui/arrivals/components/EtaStrip.kt:532`

```kotlin
val etaParts = if (eta != 0L) DisplayFormat.formatEtaParts(LocalContext.current, eta) else null
```

Anything below zero falls through to the numeric branch and prints `-1min`, `-2min`, and so on.

## Evidence

Pixel 7 Pro, Android 17, 2026-08-05. The D Line row at 15th Ave W & W Dravus St (stop `1_13800`), sampled repeatedly through `uiautomator dump` while buses came and went:

```
arrivals row    -5min   -2min   -3min   -6min      (never NOW)
tracking card   Now                                (same departures)
```

## The inconsistency that surfaced it

The tracking notification renders the same departures through `etaMinutes <= 0 -> "Now"` (`tracking/TripTrackingNotifications.kt`, `metric()`). So for a departure between -2 min and 0, the shade says "Now" while the arrivals row it was copied from says "-2min". The two surfaces already share `etaMinutes` for the number; they do not share the zero cutoff.

## To decide when picking this up

Whether NOW should cover `eta <= 0` — matching the notification, and matching `TRACKING_LINGER`, the 2-minute window the tracking policy already treats as "the rider is boarding right now" — or whether the strip genuinely wants to show how long ago a bus left. Either answer is defensible; what is not is the two surfaces disagreeing. Whichever is chosen, the cutoff should be shared the way `etaMinutes` is, not restated in both places.

## Not in scope

The pill's current look — 26sp bold white, no clock subline, `fillMaxHeight` to match its taller neighbours — is deliberate, from #1805 (render NOW at the same size as other pills) and then #1811 (shrink the label, drop the clock subline). This issue is about the pill not being reachable, not about how it is styled once it is.

bmander:
Fixes #2177.

## What was wrong

The ETA strip picked its "NOW" pill with an exact-zero test:

```kotlin
val etaParts = if (eta != 0L) DisplayFormat.formatEtaParts(LocalContext.current, eta) else null
```

so NOW was on screen for at most the one minute a departure spent at zero, and a bus that had just pulled out fell through to the numeric branch and printed `-1min`, `-2min`. The trip-tracking notification, rendering the same departures, said "Now" for all of them.

Chasing it turned up **four** restatements of the cutoff, with **three** different answers:

| surface | cutoff | effect |
|---|---|---|
| ETA strip pill | `eta == 0` | NOW unreachable except at exactly 0 |
| flat arrival row (`EtaContent`) | `eta == 0L` | same |
| trip-tracking notification | `<= 0` | correct, but only because `TRACKING_LINGER` bounds it at -2 |
| starred-stop badge (`MyListRepository`) | `<= 0`, unbounded | calls a bus that left five minutes ago "Now" |

## The decision the issue asked for

The issue left open whether NOW should be `eta <= 0` (matching the notification) or whether the strip genuinely wants to show how long ago a bus left. **Both, split at two minutes** — and the split is the notification's own window, not a new number:

- **NOW covers `NOW_WINDOW` = 2 minutes past a departure's own time.** That is the stretch in which the rider is boarding, and it is exactly the range the notification can ever render (its departures are dropped past `TRACKING_LINGER`), so the notification's behaviour is bit-identical to before.
- **Past the window it is a countdown again.** The client sends no `minutesBefore`, so the arrivals response keeps departed trips for the server's default few minutes — a `-5min` pill is a real state, and a bus that left five minutes ago is gone, not here. This is the one behaviour change beyond the fix: the starred-stop badge stops calling those "Now".

`TRACKING_LINGER` becomes `NOW_WINDOW` rather than its own `2.minutes`, by definition instead of by coincidence: the card has no rendering for a departure that has stopped being "Now", so the two windows being one value is what keeps that true.

The cutoff itself is now `isEtaNow`, living next to `etaMinutes` in `org.onebusaway.android.time` for the reason that function gives — it is the one place the rule lives, and all four surfaces call it.

⚠️ **Worth a human eye:** 2 minutes is a product judgement about when a bus stops being "here". It is not newly invented (it is `TRACKING_LINGER`, shipped in #2173, reused rather than duplicated), but this PR extends its reach from the notification to every ETA surface.

## Not in scope

The pill's look once it is on screen — 26sp bold white, no clock subline, `fillMaxHeight` — is deliberate, from #1805 and #1811. Untouched.

## Testing

- **JVM unit** (`ServerTimeTickerTest`): the window `[-2, 0]` and its edges; and that the far edge is the last minute `TRACKING_LINGER` holds a departure, so the shade and the arrivals row cannot drift apart again. `TrackingPolicyTest`'s boarding-moment test now asserts the departure still *reads* "Now", not merely that it is still listed.
- **Device** (`EtaStripNowPillRenderTest`, new): the whole just-left window renders NOW with no negative countdown surviving; a departure at -4 still says how long ago. Not `@SmokeTest` — the API-23 floor subset keeps its one strip test.
- `EtaStripRenderTest`'s recent-past pill moves -2 → -4, since -2 no longer has a number to match on.
- Verified: `compileObaGoogleDebugKotlin -PwarningsAsErrors=true`, `testObaGoogleDebugUnitTest`, `spotlessCheck`, and both strip render tests green on a Pixel 7 Pro (Android 17).

## Summary by CodeRabbit

* **Bug Fixes**
  * Recently departed arrivals now remain labeled “Now” for up to two minutes, instead of showing negative countdowns.
  * ETA displays, arrival badges, notifications, and trip tracking now use consistent “Now” behavior.
  * Departures beyond the two-minute window continue to show their negative countdown, such as “-4”.
* **Tests**
  * Added coverage for the “Now” window boundaries and negative ETA display behavior.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/arrivals/EtaStripNowPillRenderTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/arrivals/EtaStripNowPillRenderTest.kt
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.ui.arrivals
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.width
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.test.assertCountEquals
+import androidx.compose.ui.test.onAllNodesWithText
+import androidx.compose.ui.unit.dp
+import androidx.test.platform.app.InstrumentationRegistry
+import org.junit.Rule
+import org.junit.Test
+import org.onebusaway.android.R
+import org.onebusaway.android.ui.arrivals.components.EtaStrip
+import org.onebusaway.android.ui.arrivals.components.previewArrival
+import org.onebusaway.android.ui.arrivals.components.previewRowCallbacks
+import org.onebusaway.android.ui.compose.createUnconfinedComposeRule
+
+/**
+ * The NOW pill is reachable (issue #2177).
+ *
+ * It used to be selected by an exact-zero test, so it was on screen for at most the single minute a
+ * departure spent at zero, and a bus that had just pulled out fell through to the numeric branch and
+ * rendered "-1min" — while the trip-tracking notification, showing the same departure, said "Now".
+ * The cutoff is [org.onebusaway.android.time.isEtaNow] now, shared by both surfaces, so this pins
+ * the whole window rather than the one instant.
+ *
+ * Deliberately not `@SmokeTest`: the API-23 floor subset (#1818) keeps one strip render test, and
+ * that stays [EtaStripRenderTest].
+ */
+class EtaStripNowPillRenderTest {
+
+    // Unconfined composition — see createUnconfinedComposeRule (issue #1792).
+    @get:Rule
+    val composeRule = createUnconfinedComposeRule()
+
+    private val now = InstrumentationRegistry.getInstrumentation().targetContext
+        .getString(R.string.stop_info_eta_now)
+
+    /**
+     * Renders one pill per [etaMinutes]. The values are stable for the test's duration: previewArrival
+     * anchors the server "now" at 0 and liveEta floors each side to its own minute, so each pill holds
+     * its exact value for the first 60s.
+     */
+    private fun showStrip(vararg etaMinutes: Long) {
+        composeRule.setContent {
+            Box(Modifier.width(320.dp)) {
+                EtaStrip(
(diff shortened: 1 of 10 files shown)


bmander:
Abandoning — the premise doesn't hold.

Checking the history: the NOW pill's `eta == 0` test dates to at least `1d9bc9c5d` (2015-03-25), and the ETA arithmetic behind it is unchanged too. The pre-modernization `ArrivalInfo.java` computed

```java
final long predictedMins = predicted / ms_in_mins;
mEta = predictedMins - nowMins;
```

which is the same floor-each-instant-to-its-own-minute rule `etaMinutes` uses today. So `eta == 0` has always meant "the arrival falls in the current wall-clock minute", and NOW has always been on screen for exactly one minute per departure. Nothing regressed.

That makes widening NOW to a 2-minute window a **product change**, not a fix, and it isn't one to make on the back of a bug report. Closing unmerged.

Recorded for whoever picks this up: the reason NOW *feels* rarer than it used to isn't the cutoff, it's the loss of the featured-arrival slot. The old stop screen's `ArrivalsListHeader` pinned the soonest one or two arrivals at the top at 30sp with a dedicated `ETA_TEXT_NOW_SIZE_SP` styling for the NOW state, and picked them via `findPreferredArrivalIndexes` → `findFirstNonNegativeArrival`, so departed buses were skipped and the slot always held the imminent bus. Today NOW is one pill in a per-route horizontal strip, and `RouteRowGroup` sorts recent-past trips first, so departed pills lead each row and push the imminent one rightward. Same one-minute window, much less prominent placement.

Issue #2202, opened 2026-08-08
Extrapolation algorithm for trips tolerant of pronounced death spirals


bmander:
It is not uncommon for a bus to [bunch up](https://en.wikipedia.org/wiki/Bus_bunching); for example KC route 8 ("The late 8") serving Lower Queen Anne to Capitol Hill is generally oversubscribed and service can degrade significantly, reducing the speed to less than half its scheduled speed over the entire route. This is so reliable that surely we can account for it in some way in the trip extrapolator. Otherwise, the extrapolated vehicle constantly gets head of the actual coach, and then pops backwards when a new measurement comes in, making a frustrating trip even more frustrating.

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/3c8ab40e-55de-47e9-b138-149ca665b734" />

Issue #2190, opened 2026-08-07
Extrapolated position quantiles are O(1) in closed form, not a 20-step search


bmander:
`FirstPassageDistribution.quantile()` inverts its CDF by bisection — currently 20 steps, each an incomplete-gamma evaluation. It runs on the map's per-frame vehicle sampler (~20 Hz, main thread), once per vehicle for `median()`, plus ~15 more for the selected vehicle's uncertainty band and 160 for the trajectory debug view.

The search is unnecessary. The whole thing collapses to a lookup on a curve that doesn't depend on any of the model's parameters.

### Why

`reached(tau)` is `regularizedGammaP(m·tau/θ, dt/θ)`. Substituting `a = m·tau/θ` and `z = dt/θ`, solving `1 − reached(tau) = p` becomes

```
P(a, z) = 1 − p        →        a = A_p(z)
```

`A_p(z)` is a monotone curve in two variables that is **independent of `theta`, of `meanTravelMultiplier`, and of the schedule**. Recover the answer with `tau = A_p(z)·θ/m`, clamp to the horizon, and hand it to the same `interpolate(scheduleSeconds, distances, tau)` used today.

`z` is bounded: `MAX_HORIZON / θ_min = 900 / 22 ≈ 41`. Only a few `p` values are ever asked for on the hot path — `0.5` for the marker, `FAST_ESTIMATE_QUANTILE` (0.90) and `BAND_LOW_QUANTILE` (0.01) for the band — so a lazily-built table of a few hundred log-spaced `z` entries per `p` covers the entire domain.

### Verification

Checked numerically against the shipped bisection across θ ∈ {22, 50, 146}, m ∈ {1.02, 1.35}, dt ∈ [1, 600] s, p ∈ {0.01, 0.10, 0.50, 0.90, 0.99}: **maximum relative disagreement 1.1e-8**. As a sanity check on the shape, `A(0.5, z) ≈ z + 1/3` to four digits for `z ≳ 5`, which is the classic gamma-median result.

### Expected effect

Roughly three orders of magnitude off the per-frame cost, and zero allocations. For scale, before the mitigations in `d19b01529` the search was measured at 13.3 µs per `median()` on a desktop JVM (~0.38 µs per gamma evaluation); ART is 5–10× slower again. That commit halved the iteration count and removed a per-evaluation `GammaDistribution` allocation, which was enough to stop it being a frame-budget problem — this would remove it entirely.

### Notes

[...]

PR #2209, opened 2026-08-09
Widen the extrapolator's interval for a vehicle observed off its usual pace

bmander:
Addresses part of #2202. **Read the caveat before the numbers: this is a calibration fix, not a fix for the marker overshoot the issue describes.** The original version of this PR claimed otherwise; the measurements below retract that.

## What this does

A vehicle observed holding an unusual pace over a sustained recent window gets two adjustments to its position distribution: a **lump** of scheduled time added to the road ahead (+17s slow band, +16s fast band, 0 on pace) and a **widened dispersion** (`exp(c·|ln ρ|)`, bounded at ×1.76). The lump is baked into the passage profile once per snapshot as a pure translation, so the closed-form quantile path from #2190 is untouched and nothing new happens per frame.

`PaceModel` reads the window off the `TripState` history the app already keeps: ground covered in schedule terms over up to 600s ending at the anchor's GPS fix, via the new `ObaTripSchedule.scheduleTimeAt`.

## What it does *not* do

The issue describes an extrapolated marker running ahead of a bunched coach and snapping back. **This does not fix that, and no fit of this covariate will.** Measured over 400k held-out pairs at 5–15 minute horizons:

| covariate | Spearman vs forward pace |
|---|---|
| pace over the last 600 s | **−0.064** |
| pace over the whole trip | **−0.004** |

R² ≈ 0.01 either way, sign negative — mild mean reversion, not persistence. Even a bus that has run under 0.6 of schedule pace for thirty minutes has a median forward pace of **1.00**. Drawn against a real route-8 doom loop (on time → 18 minutes down, 5.0 km of an 11.4 km trip in 38 minutes), this model moves the projected median from +1564 m to +1526 m while the 80% band goes from 1145 m to 1557 m, and the coach falls inside at four of five anchors instead of three. That is the whole effect: **the interval becomes honest; the marker does not move.**

Aiming the marker needs the gap to the bus ahead — leading headway predicts forward pace monotonically (1.008 tailgating → 0.913 abandoned) and survives conditioning on deviation. Separate covariate, separate change.

## Why four parameters

A nested ladder, same split and objective, held out on 3.47M pairs:

| rung | params | cell score | worst cell |
|---|---|---|---|
| H37 (deviation only) | 0 | 0.004868 | 0.478 |
| θ, one scalar | 1 | 0.004611 | 0.310 |
| lump, one scalar | 1 | 0.004182 | 0.366 |
| **lump spline + θ scalar (this PR)** | **4** | **0.003608** | **0.344** |
| earlier nine-parameter version | 9 | 0.003164 | 0.344 |

Four parameters take 74% of the nine-parameter gain with an **identical worst cell**, and match it on slow-band coverage at every horizon. The dropped rate-multiplier spline had a fitted value *below 1* — speeding a slow bus up, partly cancelling its own lump — and removing it makes the profile warp a pure translation. Neither surviving term works alone: the lump alone makes the slow band's shape *worse*, θ alone recovers a seventh of the gain. What the extra five parameters still bought was distribution shape in the tails, not coverage.

## Held-out calibration (full test split, per horizon × deviation × pace cell)

| | H37 (today) | this PR |
|---|---|---|
| worst-cell coverage error | 0.478 | **0.344** |
| slow-band 80% coverage | 0.61–0.69 | **0.71–0.87** |
| no-lookback population | — | unchanged by construction |

## Guardrails (structural, learned from degenerate fits)

- **420 s minimum window**: below it the adjustment is identity — cold start reproduces the deviation-only model *to the bit*, asserted in the eval rather than assumed. Unconstrained, the optimiser turned short lookbacks into a stopped-right-now detector keyed to the dataset's 30 s polling cadence.
- **ρ clamped to the knot range**: unclamped, a bus that covered no ground at all has ρ = 0.005 and would multiply θ by 180.
- **Lookback mirrors the fit's own fix filters** (`phase == in_progress`, 200 m terminal margins). Without them a bus idling at its terminal then departing on time reads as deep-slow for ten minutes — a population the validation never scored.

Fit artifacts and measurements live in the companion repo (extrapolation-science): `fit_h39_g_shipped.py`, `fit_h39_ladder.py`, `measure_pace_effect.py`, `measure_chronic_pace.py`, `measure_headway_regime.py`, `plot_doom_loop.py`.

Unit tests: scipy-generated references pin the Kotlin curve; lookback selection, gating, the ρ clamp, the translation warp, and the cold-start-identity path are covered.

## Summary by CodeRabbit

* **New Features**
  * Improved arrival-time predictions using recent vehicle pace and travel history.
  * Sustained slow or fast travel now adjusts predicted arrival distributions.
  * Added more accurate schedule-time interpolation, including dwell periods between stops.
  * Standard predictions remain unchanged when insufficient or unavailable history exists.

* **Bug Fixes**
  * Improved handling of unsupported, out-of-range, or incomplete schedule data.

* **Tests**
  * Added coverage for pace adjustments, schedule interpolation, profile behavior, and historical tracking scenarios.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/FirstPassageExtrapolator.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/FirstPassageExtrapolator.kt
@@ -41,10 +41,12 @@ import org.onebusaway.android.time.WallTime
  * worst (horizon x deviation) cell's coverage error from 0.278 to 0.089, and the early-vehicle band
  * from 0.602 to 0.799 against a nominal 0.800.
  *
- * On top of that, a vehicle observed holding an unusual pace over a sustained recent window — the
- * death-spiral bus of issue #2202 — gets [PaceModel]'s adjustment: a lump of scheduled time and a
- * dispersion factor, baked into the profile and theta once per snapshot. Without a long enough
- * window the adjustment is identity, so this changes nothing for a freshly opened trip.
+ * On top of that, a vehicle observed holding an unusual pace over a sustained recent window gets
+ * [PaceModel]'s adjustment — a lump of scheduled time and a widened dispersion, baked into the
+ * profile and theta once per snapshot. That widens the interval rather than moving it, which is
+ * all the covariate supports: a vehicle's own pace history does not predict its future rate (see
+ * [PaceModel]). Without a long enough window the adjustment is identity, so this changes nothing
+ * for a freshly opened trip.
  *
  * Known residual, measured and left in deliberately: the left tail is a few points heavier than the
  * model at every horizon — trips that fall badly behind are more common than a single gamma process
@@ -60,8 +62,8 @@ class FirstPassageExtrapolator(state: TripState) : Extrapolator(state) {
      *
      * The profile starts from where the vehicle **actually is**, not its `scheduledDistanceAlongTrip`:
      * the model asks how long the road ahead should take, which is a property of that road, so a late
-     * bus is still governed by the stretch it currently occupies. The pace adjustment's mean effects
-     * are baked into the profile here ([warpedBy]); its dispersion factor rides on theta below.
+     * bus is still governed by the stretch it currently occupies. The pace adjustment's lump is baked
+     * into the profile here ([warpedBy]); its dispersion factor rides on theta below.
      */
     private val fit: Fit? by lazy(LazyThreadSafetyMode.NONE) {
         val anchor = state.anchor ?: return@lazy null
--- a/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/FirstPassageExtrapolator.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/extrapolation/FirstPassageExtrapolator.kt
@@ -41,6 +41,11 @@ import org.onebusaway.android.time.WallTime
  * worst (horizon x deviation) cell's coverage error from 0.278 to 0.089, and the early-vehicle band
  * from 0.602 to 0.799 against a nominal 0.800.
  *
+ * On top of that, a vehicle observed holding an unusual pace over a sustained recent window — the
+ * death-spiral bus of issue #2202 — gets [PaceModel]'s adjustment: a lump of scheduled time and a
+ * dispersion factor, baked into the profile and theta once per snapshot. Without a long enough
+ * window the adjustment is identity, so this changes nothing for a freshly opened trip.
+ *
  * Known residual, measured and left in deliberately: the left tail is a few points heavier than the
  * model at every horizon — trips that fall badly behind are more common than a single gamma process
  * allows. It is not simply stalled buses, whose frequency decays with the horizon as expected, and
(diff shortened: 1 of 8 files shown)


bmander:
Follow-up commit from a review pass: the runtime lookback now mirrors the research instrument's fix filters (phase == `in_progress`, 200m terminal margins). Without them, a bus idling at its terminal with AVL live and then departing on time would have read as deep-slow for its first ten minutes — a population the held-out validation never scored, since the fit pipeline drops those fixes before building pairs. An unreported phase disables the adjustment (identity) rather than guessing.

bmander:
## A worked example that partly retracts this PR's framing

I drew a real doom loop against the projections this model would make along it — KC route 8 to Mount Baker, vehicle 1_8292, on time at its first fix and 18 minutes down at its last, covering 5.0 km of an 11.4 km trip in 38 minutes. The extrapolator is stood up every five minutes and projected ten minutes forward, using a Python mirror of `DeviationModel.kt` and `PaceModel.kt` constant-for-constant.

**The cones ride above the trajectory in both panels, and the two panels are nearly identical.** At the first anchor the projected median lands **+1564 m** ahead of the coach without pace conditioning and **+1590 m** with it.

That is consistent with what was measured and accepted — the held-out numbers in this PR are *calibration* numbers, and H39-O earns them by widening the cone (θ ×1.69) and adding a 22 s lump, not by aiming it. But the issue asks for the marker to stop running ahead, and this does not do that. The PR description leads with that framing and overclaims; the numbers in it are honest.

### Why no refit will fix it

I added a trip-to-date (unbounded) lookback to rule out "a 600 s window can't see chronic slowness." It's ruled out. Over 400k held-out pairs at 5–15 minute horizons:

| covariate | Spearman vs forward pace |
|---|---|
| pace over the last 600 s | **−0.064** |
| pace over the whole trip | **−0.004** |

R² ≈ 0.01 either way, and the sign is *negative* — mild mean reversion, not persistence. Sliced rather than correlated, same story: the slowest ventile of recent pace still has a forward-pace mean of 0.92, and a bus that has run under 0.6 pace for thirty minutes has a median forward pace of **1.00**. Knowing a bus has crawled for ten minutes, the best estimate of its next ten is "about schedule pace."

The example bus makes the point by itself. Its successive (lookback → forward) pace pairs are 0.33→0.29, 0.31→0.59, 0.32→0.73, 0.59→0.79, 0.73→0.78 — it reverts from 0.29 toward 0.78 *while still shipping deviation the whole way*, because losing time and running slow are different things. At pace 0.78 you still bleed 22 s per 100.

### What I think this means for the PR

The change is a real calibration improvement and it is honest about intervals meaning what they say — the slow band's 80% band genuinely covers 80% now instead of 62%. But it should be described as that, not as a fix for the marker overshoot. Happy to rewrite the description, or to hold the PR if you'd rather land the aiming fix first.

**The live hypothesis for actually aiming the cone:** slowness is a property of the road and the hour, not the vehicle's past — a segment of route 8 whose timetable is optimistic at 10 pm. A uniform *route*-level bias is partly ruled out too (it would show as run-level autocorrelation, which measures ~0), so the effect, if real, is segment-local. That's learnable from the fleet, but it's a different covariate indexed by (position, time-of-day), and a bigger piece of work.

Figure and measurements: `extrapolation-science` commit `3537e41` — `plot_doom_loop.py`, `measure_chronic_pace.py`, `docs/doom_loop_projections.png`.

bmander:
## Two follow-ups: most of these parameters aren't earned, and there's a covariate that *does* move the mean

### 1. The nine parameters can be four

Fitting nested rungs on the same split and the same per-cell objective, held out on 3.47M pairs:

| rung | params | cell score | worst cell |
|---|---|---|---|
| A — H37 (deviation only) | 0 | 0.004855 | 0.414 |
| B — θ, one scalar | 1 | 0.004611 | 0.310 |
| E — lump, one scalar | 1 | 0.004182 | 0.366 |
| F — lump spline only | 3 | 0.004161 | 0.413 |
| **G — lump spline + θ scalar** | **4** | **0.003618** | **0.271** |
| D — H39-O (this PR) | 9 | 0.003099 | 0.300 |

**Rung G takes 70% of the win with four parameters and a *better* worst cell than the nine-parameter version** (0.271 vs 0.300), matching it on slow-band coverage (0.744 vs 0.751). It drops the rate-multiplier spline — the term whose fitted value *below 1* was speeding a slow bus *up*, partly cancelling its own lump. Without it the profile warp becomes a pure translation (`τ + Δ`) rather than scale-and-shift, which also removes the lifted-first-knot subtlety from `warpedBy` and the two KDoc caveats that came with it.

The non-obvious part: **neither term works alone.** The lump by itself makes the slow band's *shape* actively worse (CvM 3947 → 5222); θ by itself recovers 14%. They're only useful together. The remaining 30% that D buys is distribution shape, not coverage.

I'd suggest reducing to rung G unless the shape gain is wanted. Happy to do that.

### 2. Leading headway does what pace history couldn't

Bunching predicts the speed distribution is a *mixture* — front-of-bunch absorbs the demand and crawls, back-of-bunch finds emptied stops and sails — and the identifying variable is the gap to the bus ahead, not the vehicle's own past. That's testable without any geometry: for two runs on the same pattern at the same distance, `actual headway = scheduled headway + (dev_this − dev_prev)`.

Over 419,730 fixes with a measurable leading headway, median forward pace over the next 10 minutes:

| <0.6× (tailgating) | 0.9–1.15× | 1.5–2.2× | >2.2× (abandoned) |
|---|---|---|---|
| 1.008 | 0.990 | 0.948 | **0.913** |

Monotone, and it **holds inside every schedule-deviation band**, so it's incremental to what the model already conditions on. `P(pace < 0.7)` in the abandoned band is 0.197 against a 0.073 baseline.

And the doom-loop run from the figure is a textbook case: leading headway 1.6–2.4× scheduled (32–48 minutes actual against a 20-minute headway) with a follower at 0.03–0.69×.

Caveats, both measured: the effect is 5–9% in median pace where that run actually did 0.72, so headway closes roughly a quarter of the gap rather than all of it; and the *trailing* gap adds nothing once the leading gap is known (the ahead×behind grid is non-monotone across the trailing axis) — so there's no "which half of the pair" logic worth building.

This is the first covariate I've tested that moves the mean in the right direction and survives conditioning on deviation. It needs other vehicles' positions on the route, which the app already fetches for the vehicle map layer.

Measurements: `extrapolation-science` commit `539849e` — `fit_h39_ladder.py`, `measure_headway_regime.py`.

bmander:
This is a kludge; shelving it for now. Ultimately what we want is enough historical data on each trip to be able to make confident extrapolations.

Issue #2107, opened 2026-07-31
Improve the transit center user experience


bmander:
Presently the transit center user experience is very poor: a user has to click through every stop in a transit center to find the one they want. I propose two improvements:

(1) At a zoom level roughly corresponding to a transit center (about a block wide), show a list of transit routes right on the stop marker.
(2) At the same zoom, the stop drawer auto-engages as a drawer peek, showing arrivals for every stop in the view.

PR #2212, opened 2026-08-09
Tune the nearby-routes drawer, and de-duplicate what it borrowed (#2107)

bmander:
Follow-up to #2198, which landed the transit-centre drawer. Two things: the UX tuning that came out of using it on a device, and the cleanup pass the first cut deferred.

## UX

- **The peek is per-content.** A tapped stop is a deliberate request to read *that* stop, so it keeps its 30% peek and opens showing real rows. The nearby list is ambient — it appears on zoom without being asked for, over a map the rider is still reading — so it peeks at 15%, enough to say a list is there. The fraction is a property of `HomeSheetContent` rather than an `if` in `HomeScreen`.
- **The expanded sheet stops below the status bar / notch.** `BottomSheetScaffold` has no max-height parameter: it places the Expanded anchor at `containerHeight - measuredSheetHeight`, so a content list that measures the full window pins the top edge at y=0. Capping the content height is the only lever that moves that anchor. Uses `safeDrawing` rather than `statusBars`, since a cutout can exceed the status bar.
- **A centered "Nearby routes" title** at the head of the list, marked `semantics { heading() }`.
- **Rows drop the stop code.** It's the most prominent thing the feed publishes about a bay and the least useful thing here — the list is scanned down the route column, and a leading number on every row is noise. Still one tap away on the stop's own panel.
- **Tapping a row selects the stop on the map.** The marker comes from `focusedStopId`, which only gets set when `onArrivalsLoaded` consumes a pending-focus latch — and this path explicitly cleared it, so the marker never appeared. Arming it with `preserveViewport = true` sets the marker *and* disposes of a stale restore latch, which is what the clearing was there for, without the recenter that motivated the clearing.
- **Tapping a row pans to the bay instead of framing the line.** The drawer only exists at transit-centre zoom, so framing the whole route would zoom out to its full extent and throw away the view the rider is standing in.
- **Tapping an ETA pill drills to that trip** and fits the vehicle together with the bay — the same gesture the per-stop drawer's pill makes, through the same `focusTripId`. A blank tripId falls back to row-body behaviour.

## Cleanup

- **`ReportListContentHeight`** — both arrivals sheets made the identical Material3 content-height measurement, copy-pasted. Now one helper, which also reports once per content identity rather than every scroll frame (the old `derivedStateOf` republished `last.offset` per frame, invalidating `HomeScreen`'s outer body).
- **`showNearbyRouteOnMap` goes through `showStopRoute`**, so it can't be the path that breaks the `ShowRoute`/`SelectVehicle` pairing invariant that helper exists to hold. A row-body tap now clears a vehicle a previous pill left selected.
- **Nearby row actions resolve into a trip-keyed map once per response.** They were rebuilt on every call, including a hex-colour parse, and the callers are hot: the ETA strip recomposes once a second off the live clock, and the alert-glyph check walks every trip in every row.
- **Bay direction goes through `DisplayFormat.stopDirectionText`** like every other stop surface — riders read "Southeast bound" rather than the feed's raw `SE`, and read it translated. This was the only place in the app showing a raw GTFS direction code.
- Dropped the unreachable second bay null-check in `nearbyRouteRows`, the dead `NearbyArrivals.situations()` copy (no callers), and two KDoc links to the `shouldShowSheet` that #2198 deleted.
- The headsign/bay gap rides on the stop label instead of an unconditional spacer, which had shifted every existing stop-scoped arrivals row by 2dp.

## Testing

`spotlessCheck`, `testObaGoogleDebugUnitTest` (244 classes), and `assembleObaGoogleDebug -PwarningsAsErrors=true` all pass. Two new `HomeViewModelTest` cases pin the row-tap and pill-tap map behaviour. Every UX change above was checked on a physical device.

## Deferred

`rememberNearbyRouteRows` converts a whole viewport of arrivals to `ArrivalInfo` synchronously in composition, once per poll and once per pan. `ArrivalsDisplay` exists precisely so this can happen off the main thread (as `DefaultArrivalsRepository` does), which would mean moving row-building into `NearbyArrivalsViewModel` and having `Loaded` carry the rows. That's a restructure of the VM, its UI state, the callbacks, and its tests — left for its own PR rather than buried here.

## Summary by CodeRabbit

* **New Features**
  * Tapping a nearby route or ETA now focuses the relevant route, stop, or vehicle on the map.
  * Added a localized “Nearby routes” title and clearer direction-based bay labels.
* **Improvements**
  * Nearby arrivals sheets adjust collapsed and expanded heights to available screen space.
  * Improved spacing and layout consistency for arrival rows.
  * Nearby route lists handle arrivals and bay information more reliably.
* **Accessibility**
  * Added an accessible title for the nearby-arrivals list.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/compose/ReportListContentHeightTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/ui/compose/ReportListContentHeightTest.kt
@@ -1,0 +1,158 @@
+/*
+ * Copyright (C) 2026 Open Transit Software Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.onebusaway.android.ui.compose
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyListState
+import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.test.ComposeTimeoutException
+import androidx.compose.ui.unit.dp
+import java.util.Collections.synchronizedList
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * On-device tests for [ReportListContentHeight], the measurement both arrivals sheets use to fit their
+ * collapsed peek to short content.
+ *
+ * The case worth pinning is **replacing the content while reusing the same [LazyListState]**, which is
+ * what the drawers do on every poll and every pan. The effect restarts on the new content key before
+ * that frame's layout pass has run, so the first `layoutInfo` it can see still describes the previous
+ * list. A measurement that reported only that first reading would publish the outgoing list's height
+ * and never correct itself — so these tests assert the height the reporter *settles* on, not the first
+ * one it emits.
+ */
+class ReportListContentHeightTest {
+
+    // Unconfined composition — see createUnconfinedComposeRule (issue #1792).
+    @get:Rule
+    val composeRule = createUnconfinedComposeRule()
+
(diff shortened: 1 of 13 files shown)


bmander:
Addressed in 1bcbee36 — the finding was correct, and it was a regression the cleanup pass introduced.

**The bug.** `.first()` reported only the initial measurement per content key. As you noted, restarting the effect on a new key doesn't wait for that content's layout pass, so the restarted collector's first `layoutInfo` still describes the outgoing list — and nothing retried. Confirmed on device before fixing: swapping a 3-row list for an 8-row one on a shared `LazyListState` reported `[159, 159]`, the second entry being the stale re-read of the 3-row layout.

**The fix**, which is a bit deeper than the suggested post-layout gate. Rather than trying to identify a layout signal belonging to the new content key, I stopped restarting: one collector keyed only on the `LazyListState`, observing layout for as long as the composable is in composition. A content swap then needs no special handling — the new height is simply the next measurement. That removes the stale-read window structurally, and `contentKey` disappears from the API and both call sites.

Two things fell out of getting this right on hardware:

- **Only measure from the top.** Item offsets are viewport-relative, so a scrolled list's last-item bottom is its on-screen position, not the content height. Skipping those readings is a correctness fix — and since the per-frame-moving offsets are exactly the ones skipped, it also delivers the no-churn-while-scrolling property that `.first()` had been the wrong way to get.
- **Dropped the `last.index == totalItemsCount - 1` condition.** For content past the viewport the same expression gives the straddling item's bottom edge — a floor rather than a measurement, which is all a caller clamping to a peek cap can use, and better than the silence that previously left a shorter list's height standing.

**Test.** Added `ReportListContentHeightTest` with the swap case you asked for, in both directions on a shared `LazyListState`, plus a fits case and an overflow-floor case. Two of the row counts fit the viewport, so the swap assertions are about the exact value rather than a floor. All four pass on device; they fail against the previous implementation.

Issue #2194, opened 2026-08-07
Move vehicle fullness status directly into marker design and remove marker info bubble


bmander:
Presently every piece of information in the vehicle marker info bubble is redundant except the vehicle fullness, and its interaction plays poorly with the app interaction model. Two steps:

(1) Move the occupancy indicator into the marker design. This is an open design problem.
(2) Delete the info bubble.

PR #2215, opened 2026-08-09
Flip the vehicle marker's rim to white in dark mode (#2055)

### LINKED ISSUE #2055: Tune vehicle marker
In the vehicle map marker tune the visuals slightly. The marker outline is black always; it should be light when in dark mode to show up.

bmander:
Closes #2055.

The vehicle marker is rimmed with a hairline so its silhouette holds apart from what's behind it — and what's behind it is a base map the app restyles by mode (`R.raw.light_map` / `R.raw.dark_map`). The rim was black either way, so on the dark base map it stopped separating anything, and on a route color dark enough to sit near the night basemap it took the disc's edge with it.

The rim now resolves from a qualified resource, `vehicle_marker_outline` — black in light mode, white in dark. That's the same flip `route_stop_outline` already makes for the route-stop circles on the same map, made the same way (qualified resource rather than a `ThemeUtils.isInDarkMode` branch) so the two stay comparable.

### What deliberately did *not* change

Only the **body's** rim follows the mode. The mode glyph's black dilate holds the glyph off the disc's wire-supplied route color, which no theme moves; the occupancy pips keep their fixed white-empty / black-full polarity for the reason already documented at the pip loop. Both are unchanged, and now say why at the site.

### Cache keys

Because the rim varies, it joins `VehicleBitmaps.iconKey` and the bitmap LRU key. Both caches outlive a light/dark switch — the bitmap LRU is a process-wide object and the Google flavor's descriptor cache lives as long as the renderer — so a key that ignored the rim would hand every already-drawn vehicle its previous mode's rim indefinitely. It's constant across every vehicle in a frame, so this costs one token in the key and nothing else.

### Tests

- **`VehicleMarkerOutlineTest`** (new, instrumented) pins both halves: the color resolved per mode, and the rim as actually drawn — read as the marker's topmost non-transparent pixel and asserted structurally against the disc (strictly darker in light, strictly lighter in dark), so no antialiasing tolerance has to be invented. Either half can pass while the other is broken, which is why both are here.
- **`VehicleIconAllocationTest`** gains the key/bitmap staleness guard for a mode switch.
- **`VehicleMarkerPipTest`** now renders against an explicitly light-mode context. Its pip counts sample the tab band, where a white rim would answer to the "empty pip" predicate — the assertions would have become a reading of the device's theme rather than of the pip row.

All 7 pass on a Pixel 7 Pro (API 36).

### Preview

`VehicleMarkerGrid` gains a dark-mode `@Preview` and stands its markers on a mode-following ground, since on a fixed white ground the dark-mode rim would be invisible in the tuning tool and legible in the app.

### Verification still wanted

The pixel-level behavior is covered; what a rider actually sees on a live dark map is not. Worth an eyeball on the route/trip map in dark mode against a few route colors.

## Summary by CodeRabbit

* **New Features**
  * Map vehicle markers now use theme-appropriate rim colors, improving visibility in light and dark modes.
  * Marker rims are consistently sized across vehicle and trip markers.
  * Vehicle glyphs and occupancy indicators render without black outlines for a cleaner appearance.
  * Marker previews adapt backgrounds, labels, and artwork to the selected system theme.

* **Bug Fixes**
  * Marker artwork refreshes correctly when switching between light and dark map modes.
  * Improved contrast and consistency for map marker silhouettes and glyphs.


bmander:
Pushed a second commit: **the mode glyph now draws flat**, with no rim of its own.

It carried the same black 8-way dilate the body gets, but it had nothing to separate itself from — it sits alone on a disc whose color `MarkerRendering.legibleOn` already picked to contrast with it, and on the light route colors that make `legibleOn` choose a *black* glyph the rim was black-on-black, so it only thickened the artwork into a blot.

Only the body is rimmed now. It's the one part with a base map behind it, which is what a rim is for. The occupancy pips shed theirs back in #2194 for this same reason; the glyph now matches them. The consequence is that `legibleOn` is the sole thing keeping the glyph off the disc, which the KDoc now states.

`VehicleMarkerOutlineTest` gains the check — no black anywhere inside the disc, read over a square inscribed in it, paired with an assertion that the glyph is still drawn there, since "no black in the disc" is equally what an empty disc looks like. 21 tests across the three marker classes pass on a Pixel 7 Pro (API 36).

bmander:
Third commit: **the mode glyph grows 1.5x**, 10.8 grid units to 16.2.

At the old size it read as a small mark floating in a large disc — its ink spanned about 38% of the disc's diameter, where an icon-in-circle normally runs to half or better, and losing the rim in the previous commit took away some ink too.

16.2 is where this stops without the disc growing with it. It's the largest box that is *itself* inscribed in the disc, rather than merely having its ink land there: half-diagonal `16.2/2 × √2 ≈ 11.46` against the 11.625 fill radius. That's the bound worth holding, because it doesn't depend on how much of its box any one mode's artwork happens to use — the five drawables run 70–85%, and a future glyph drawn edge to edge would still land on the disc rather than over its rim.

The glyph stays centered on the disc, so it clears the occupancy tab's pip row by a wide margin. Nothing else in the geometry moves.

`VehicleMarkerOutlineTest`'s sampled square widens 7.2 → 7.5 so it keeps holding the glyph's *edges*, which is where a dilate would put its black — a square holding only the interior could pass with the halo sitting just outside it. Its corners still clear the rim's inner edge. 21 tests pass on the Pixel.

bmander:
Addressed both review items.

**1. `remember` key in the preview grid — fixed.** Valid catch. `previewBitmap` resolves the rim from the `Context`'s mode, but `remember(vehicleType, argb, occupancy)` didn't include the mode, so a composition surviving a mode change instead of being disposed (an interactive preview toggling the theme) would keep showing the rim it was first drawn with. `dark` is now passed into `MarkerRow` and joins the key, with a note saying why it's a parameter rather than read inside.

**2. `VisibleForTests` suppressions — partly fixed.** The narrowing half was right: `MarkerPixelFixtures` had a `@file:Suppress` to reach one constant, which is wider than the use, and its comment still named `MARKER_SIZE_DP` — a constant the file stopped using when `markerScale` moved to a width-based derivation. It's now scoped to the `markerScale` getter alone.

I did **not** add tracking-issue links to either suppression. The CLAUDE.md rule requiring one ("always add a one-line rationale *and* a tracking-issue link") sits under "CI is strict — warnings fail the build" and governs suppressions that defer work — a deprecation that needs a higher `minSdk`, a schema change — where the link is what stops the deferral from being forgotten. Nothing is deferred here: these are tests calling `@VisibleForTesting` members, which is the annotation working as designed, and there is no future change for an issue to track. Both sites carry the one-line rationale, matching the pre-existing `@Suppress("VisibleForTests")` in `VehicleMarkerPipTest` on `main`.

33 map instrumented tests pass on a Pixel 7 Pro; compile clean under `-PwarningsAsErrors=true`, spotless clean.

Issue #2042, opened 2026-07-27
UX refinements for interleaved legs


bmander:
Interleaved transit legs on the map need two refinements; each could be a PR on its own.

(1) When tapped, show _all_ routes that interleave into the leg on the map, not just the first/preferred/OTP-returned route.
(2) Interleave all transit routes into a merged ETA row. Because multiple routes are sharing the same ETA row/vehicle queue, each pill in the ETA row will need to indicate which route it's on. This is an open design question.

Issue #2003, opened 2026-07-23
Trip status page should show trip id, links to previous and next trips in the block


bmander:
Trip status page should show trip id, links to previous and next trips in the block

bmander:
This was an exploration into how we can help the user make sense of the difference between "trip is being tracked" and "block is being tracked" but this particular feature would mostly just be noise to the user. Closing.

PR #2221, opened 2026-08-13
Resolve each vehicle marker against the poll it came from

bmander:
On a "whichever comes first" Link leg, every 2 Line train on the map draws with the **bus** glyph, while the 1 Line beside it draws correctly as a tram. The same vehicles announce themselves to a screen reader as "unidentified".

### It isn't the API

Puget Sound reports `40_2LINE` as `type: 0` (tram), and a live `trips-for-route/40_2LINE` poll resolves cleanly — all 16 entries' `status.activeTripId` are present in `references.trips`, each mapping to a route with `type: 0`.

### What was wrong

A vehicle **set** is not always one poll's worth of vehicles. A route-focus session with interchangeable routes (#2042) or a stay-aboard interline (#2000) merges the leader poll with an extra poll per extra route — `RouteMapController.sampleVehicles` builds `leaderVehicles + extraVehicles`, pairing each vehicle with the poll it came out of.

`MapVehicles` then dropped that pairing and published `response = poll.response`, the leader's alone. Both renderers resolved *every* marker's `trip` → `route` against it:

```
vehicleIcon(vehicle, response) → VehicleBitmaps.routeTypeFor(response, activeTripId)
vehicleTitle(context, vehicle, response)
```

A poll's references answer for its **own** vehicles. Measured on the live feed: the 1 Line's poll carries 18 of its own trips and only 2 of the 2 Line's 16 — so nearly every 2 Line vehicle missed both hops and took `DEFAULT_VEHICLE_TYPE` (bus) and the `vehicle_marker_unidentified` fallback. Both fallbacks are correct for their real cases (#2020, #2003); this made a routine outcome of them.

#2043 fixed this exact mismatch for the disc **colour** by pairing each vehicle with its poll and resolving in the controller. The type/name half was never carried the rest of the way.

### The fix

`VehicleMarker` gains `source: RouteTrips` — the poll the vehicle came out of — beside the already-per-vehicle `routeColor`/`bandColor`. `sampleVehicles` keeps the pairing it already had instead of dropping it at the last step.

Because the marker now holds a non-null pool, `vehicleBitmap` / `iconKey` / `vehicleTitle` read `vehicle.source` rather than taking a pool alongside the vehicle it has to agree with. Net effect is a **deletion**:

- `MapVehicles` is back to `data class MapVehicles(val markers: List<VehicleMarker>)`
- both renderers lose `lastVehicleResponse`, its assignment, and the parameter threading — including MapLibre's reset in `clearAll`, which the Google flavor had no counterpart for
- passing the wrong pool (or none) is no longer expressible at the render layer

141 insertions, 109 deletions across 9 files.

### Tests

- `VehicleBitmapsTest.aPollResolvesOnlyItsOwnVehicles` — two Link polls; the 2 Line trip resolves tram against its own pool and falls back to bus against the 1 Line's. The bug, pinned at the pure seam.
- Folded the file's id-*insensitive* `routeTrips` fake into the id-keyed `pool`. The two differed in exactly the property this change is about, so a later test reaching for the older helper would have been structurally unable to reproduce the bug.

`./gradlew spotlessApply` applied; both flavours compile clean under `-PwarningsAsErrors=true`; full JVM unit suite passes.

### Not in this PR

`RouteMapController` already resolves `routeColor` per vehicle and puts the answer on the marker. Route type, display name, and headsign could ride the same way, and then no `RouteTrips` would reach the render layer at all. That needs `routeColorMemo` broadened into an identity memo (otherwise three more reference hops land on the 20 Hz sampler), which is a design change to the identity pipeline rather than a bug fix — worth doing as a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vehicle marker rendering by associating each vehicle with its own trip information.
  * Vehicle icons and titles now remain accurate when vehicles come from multiple routes or polling responses.
  * Improved refresh behavior when map scale changes or vehicle data updates.
  * Improved icon caching for more consistent map performance.
  * Preserved fallback handling for missing trip, route, and vehicle details.

### DIFF
--- a/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
+++ b/onebusaway-android/src/androidTest/java/org/onebusaway/android/map/VehicleIconAllocationTest.kt
@@ -93,6 +93,7 @@ class VehicleIconAllocationTest {
                     point = v.point,
                     isRealtime = v.status.isLocationRealtime,
                     status = v.status,
+                    source = response,
                     fixTimeMs = v.fixTimeMs
                 )
             }
@@ -129,11 +130,11 @@ class VehicleIconAllocationTest {
             for ((index, vehicle) in vehicles.withIndex()) {
                 val filling = withOccupancy(vehicle, OCCUPANCY_CYCLE[(poll + index) % OCCUPANCY_CYCLE.size])
                 counts.requests++
-                val key = VehicleBitmaps.iconKey(context, filling, response)
+                val key = VehicleBitmaps.iconKey(context, filling)
                 counts.keys.add(key)
                 cache.get(key) {
                     counts.bitmapDecodes++
-                    VehicleBitmaps.vehicleBitmap(context, filling, response)
+                    VehicleBitmaps.vehicleBitmap(context, filling)
                 }
             }
         }
@@ -203,8 +204,8 @@ class VehicleIconAllocationTest {
         val samples = vehicles.flatMap { vehicle ->
             OCCUPANCY_CYCLE.map { occupancy ->
                 val reporting = withOccupancy(vehicle, occupancy)
-                VehicleBitmaps.iconKey(context, reporting, response) to
-                    VehicleBitmaps.vehicleBitmap(context, reporting, response)
+                VehicleBitmaps.iconKey(context, reporting) to
+                    VehicleBitmaps.vehicleBitmap(context, reporting)
             }
         }
         val byKey = samples.groupBy({ it.first }, { it.second })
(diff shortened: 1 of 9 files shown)


bmander:
Valid, and fixed in 3414b11c2.

Confirmed the mechanism against this renderer rather than the SDK docs alone: every *other* marker kind here already caches its `Icon` for exactly this reason — `routeBadgeIcons`, `tripCircleIcons`, `rentalIcons`, each carrying a comment that `fromBitmap` mints a fresh sprite id and that a replaced icon is never released. The Google flavour caches vehicle icons through `descriptorCache` on the same path, under the same `VehicleBitmaps.iconKey`. Vehicles on maplibre were the one holdout.

Worth noting what made it easy to miss: `VehicleBitmaps` has its own bitmap LRU, and its KDoc claimed that cache was where the bound lived *because* maplibre had no second level. But that cache hands back the same `Bitmap`, and re-wrapping it registers another sprite — so it bounded the bitmaps and not the sprites. Added `vehicleIcons` (`LruCache<String, Icon>`, keyed by `iconKey`, evicted in `dispose`), and corrected that KDoc so it stops asserting the gap is deliberate.

Pre-existing rather than introduced by this PR — the allocation frequency is unchanged from `main` — but this PR reworked `vehicleIcon`, and the fix is the file's own idiom, so it belongs here.

---

On the 🔴 Critical mergeability score citing "a compilation-blocking duplicate declaration": I can't reproduce that. Both flavours compile clean under `-PwarningsAsErrors=true`, the full JVM unit suite passes, and CI's `test (33)` job passed on d14eb32. If it refers to something specific, please point at the declaration and I'll take another look.

Issue #2127, opened 2026-07-31
For interlined routes show a line across the route at the cutover point


bmander:
Presently there is no indication on the of where an interlined route change over. Put a line through the route to indicate the cutover point. This should be visually distinct from the bulb/bulb line ending of two routes meeting at a stop that require an alight and board.

Similar to:

```
----------------------
   [45]   \   [75]
----------------------
```

bmander:
Done in #2133 (merged as bc45df75).

A stay-aboard interline now draws a slash cut across the corridor at the join where the vehicle changes route — deliberately unlike the endpoint bulbs, since a bulb-to-bulb join means "get off and board again", which is the one reading a stay-aboard route change must not get. The cut goes exactly where the bulbs are withheld.

- Itinerary legs on the directions map (`DirectionsMapController`), and the drilled-into ride, which is now split into one span per route it is ridden as so each takes its own route's colour and the seam is an end again (`RouteSegmentHighlight`).
- Both renderers: a `CustomCap` on gms, a rotated symbol layer on maplibre (its classic annotation has no configurable cap).
- Which joins are cut comes from `Interlines.chains`' own transitions, so the map cuts in the same places the drawer tells the rider to stay on board and nowhere else — a self-interline (a route reversing onto itself) is left unmarked.

Not covered here: the route-map interline continuation (#1691), where the cutover reads as "solid line ends, dashed tail in the next route's colour begins". That's a different design rather than a gap in this issue; worth its own issue if the slash should appear there too.

Issue #2084, opened 2026-07-30
Draw circular bulbs at end of transit legs


bmander:
Draw circular bulbs at end of transit legs. This will help visually separate multiple transit legs with the same GTFS color.

Issue #2048, opened 2026-07-27
When focused on a directions leg, continue to show the rest of the route deemphasized


bmander:
When focused on a directions leg, continue to show the rest of the route deemphasized

Issue #2100, opened 2026-07-31
Stripe different-color routes in interchangeable segments in the directions page


bmander:
Presently when multiple interchangeable routes share a transit leg, the badge might show multiple colors but the route only draws one color. Figure out a way to show all the colors on the route. One option is side-by-side stripes; another is sequential stripes. It's not clear which will look best.

PR #2245, opened 2026-08-18
Draw no leg-end bulbs on a drilled-into ride (#2241)

### LINKED ISSUE #2241: Bug: route end bulbs display on interchangeable-route directions legs
In the map rendering of transit legs, the leg-end bulbs do not render when a transit leg is selected, except for some reason with interchangeable-route legs. That shouldn't happen.

bmander:
Fixes #2241.

A directions transit leg is bulbed at each end of the ride, so consecutive rides stay segmented and a bulb pair reads as "alight here, board there" (#2084). Drilling into one leaves that vocabulary behind: the route view draws the ridden span with no marks at all and names the boarding with the stop the rider gets on at, and the cutover slash is the only end mark it keeps (#2127).

The rider's journey is retained around that ride as context (#2048), and each context leg is the itinerary's own line reduced — half weight, no chevrons, no stripes — but it kept its end marks. Including the leg being drilled into, whose context copy the ride is redrawn over: a mark the view had already decided not to draw, held out of sight only by what happened to land on top of the point it is drawn about. On an interchangeable ride it was not covered, and the bulbs showed.

Bulbs have one producer in the app (`DirectionsMapController`), and the route view's whole drawn set is the approach lines, that context, and the ridden span — so the context was the only thing that could still be carrying one.

Reduce the marks with everything else the context reduces. Same rule `asPinnedTripGhost` already applies for the same reason — the marks belong to a trip being followed at full weight, not to the faint trace saying where one leg of it sits — so the drilled-into ride now draws no bulbs whichever routes it may be boarded on, and the journey around it none either.

## Testing

Two new JVM tests:

- `RouteViewGeometryTest.itinerary context carries no end marks`
- `RouteSegmentHighlightTest.routePolylinesWithSegment_drawsNoLegEndBulbs_whateverTheRideIsSharedWith` — asserted over the whole drawn set, since "no bulbs here" is a property of the view and not of the one line that happens to be selected.

`testObaGoogleDebugUnitTest` for both classes passes with `-PwarningsAsErrors=true` (30 tests, 0 failures); `spotlessCheck` clean.

Not device-verified: the precise reason an interchangeable ride's bulbs were uncovered while an ordinary ride's were not is a rendering-order/geometry detail, so a look at a 1 Line/2 Line leg before merge is worthwhile.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed route map itinerary lines displaying endpoint markers intended only for actively followed trips.
  * Ensured all rendered itinerary polylines consistently remove start and end bulb markers.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/map/RouteViewGeometry.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/map/RouteViewGeometry.kt
@@ -88,12 +88,25 @@ internal fun List<RoutePolyline>.asSelectedRouteApproach(): List<RoutePolyline> 
  * question about a ride the rider is choosing; a context leg is here to say where the leg being read sits
  * in the journey, and is drawn at half the width the stripes were cut against — so the same rhythm arrives
  * as a fleck of noise beside the leg that is actually being read.
+ *
+ * And it drops both end marks (#2241), for the same reason and with the same force as
+ * [asPinnedTripGhost]: a bulb pair means "alight here, board there" and a cut means "the route changes
+ * under you", which are readings of a trip being *followed*, not of the faint trace saying where one leg
+ * of it sits. The leg they could still be read on is the ride the rider drilled into — and this view
+ * redraws that ride at full weight over its context copy, deliberately marking where the rider boards
+ * with the boarding *stop* rather than a bulb (see [routePolylinesWithSegment]). A mark surviving into
+ * the context is therefore one this view has already decided not to draw, held under the redraw only by
+ * the redraw happening to cover it: a mark is drawn about the line's endpoint, not inside its stroke, so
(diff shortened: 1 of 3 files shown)


bmander:
Closing unmerged: the direction here is wrong. The bulbs were never the problem — the inconsistency was, and deleting them to reach consistency also leaves the real regression standing. Drilling into a ride re-derives it in route-mode vocabulary, which drops the shared-ride striping (#2100) as well as the marks: tapping an interchangeable ride to look closer is exactly when the map stops saying it is two routes.

Both losses have one cause, and it is what #2246 is about — the drill-in re-derives the ride instead of reducing the leg line that already exists. Refiling as one change on top of #2246's builder unification: the drilled-into ride is drawn from the leg's own line, striped when shared, with the ride's own end marks, plus the selection case.

PR #2306, opened 2026-09-08
Degrade stop selection state after 4h of inactivity (#2294)

### LINKED ISSUE #2294: App reverts to no stop selection after a while
From user feedback:

> Also the app does not reset itself. You can open it hours later and it is still tracking the last stop you chose the day before. Is this wasteful of energy in some way? There used to be a setting where you could turn this behavior off. Plus it requires extra clicks to get out of the last stop in order to search again.

This is just default Android behavior; the state is preserved faithfully now whereas in the past the app tended to get nuked by the OS and lose state. But this did have a useful function: the OS forgetting kept pace with humans losing state on the travel task. It's not clear if there really was an option to reset state after a while. But they're on to something; it may be useful to degrade state at a similar schedule to the underlying task expiration.

This is an open design question. My inclination: use a settings param for focus timeout, defaulting to 4 hours. The settings text could be "Remember stop focus for": [30 min, 1 hour, 2 hours, 4 hours, 8 hours, Always]

prince-0408:
## Summary

Fixes #2294

The app was restoring the last selected stop indefinitely, requiring extra taps to dismiss it before searching again.

## What was changed and why

Added a 4-hour expiry to the saved focus state in `CurrentFocusPersistence.kt`.

- `write()` stamps the current time to `KEY_FOCUS_TIMESTAMP` on every focus change
- `read()` returns `CurrentFocus.None` if the saved timestamp is older than `FOCUS_EXPIRY_MS` (4 hours)

In-session use resets the clock continuously, so the focus only expires when the app hasn't been interacted with for 4+ hours. Saves written before this change (no timestamp) are left untouched.

**Note on the 4-hour value:** This is a judgment call based on the typical transit task lifespan — long enough to cover a multi-hour journey without false positives, short enough to reliably reset between commutes. Happy to adjust if maintainers prefer a different threshold (e.g. 2 hours or 8 hours). The constant `FOCUS_EXPIRY_MS` is in one place and easy to change.

## Testing

- Focus a stop, then background the app for 4+ hours — on next open the map should show no selected stop
- Focus a stop and reopen within 4 hours — the stop should still be selected
- To test quickly: set `FOCUS_EXPIRY_MS` temporarily to a small value (e.g. `10_000L`), background and reopen after 10 seconds

## Summary by CodeRabbit

- **Bug Fixes**
  - Persisted home-screen focus now expires after four hours, preventing outdated focus states from being restored.
  - Focus state records now include the time they were saved for more accurate restoration.

### DIFF
--- a/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/CurrentFocusPersistence.kt
+++ b/onebusaway-android/src/main/java/org/onebusaway/android/ui/home/CurrentFocusPersistence.kt
@@ -38,6 +38,7 @@ internal object CurrentFocusPersistence {
     private const val KEY_ROUTE_LEG_NAMES = "home.currentFocus.route.legNames"
     private const val KEY_ROUTE_LEG_DIRECTIONS = "home.currentFocus.route.legDirections"
     private const val KEY_ROUTE_SELECTED_TRIP = "home.currentFocus.route.selectedTripId"
+    private const val KEY_FOCUS_TIMESTAMP = "home.currentFocus.timestamp"
 
     private const val FOCUS_NONE = "none"
     private const val FOCUS_STOP = "stop"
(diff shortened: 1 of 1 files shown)


bmander:
Thank you for your contribution but it ended up being a fairly large change.

Superseded by #2308

Issue #1075, opened 2021-06-25
Add a home screen widget


barbeau:
**Summary:** 

From https://onebusaway.ideascale.com/a/dtd/568461-8715:

>Today View Home Screen Widget
>Please implement the Today View widget found on the iOS versions into Android! It's exactly what I need!

@aaronbrethorst @ualch9 Could you add some screenshots to this issue of what the latest version of this on iOS looks like?

**Steps to reproduce:** 

Try to add an OBA widget

**Expected behavior:** 

Give me an OBA widget like OBA iOS

**Observed behavior:** 

No widget exists

aaronbrethorst:
![C1F5702D-0B1B-4531-AABB-028EFCC908DF](https://user-images.githubusercontent.com/2254/123473718-94177580-d5be-11eb-8deb-899f25f1ac76.png)

barbeau:
@aaronbrethorst Thanks! Does this automatically refresh, and if so how often? And are the route/directions shown here ones that the user has starred in the app? Curious on the user experience for setting this up.

Also, is there an expanded view (I see the chevron...)?

aaronbrethorst:
Good questions!

* The widget refreshes, I believe once per minute. Or the user can tap on the Refresh button 
* The ArrivalDeparture objects shown here are application bookmarks that have a second flag set on them that says 'show me in the Today View widget'
* This is actually the expanded view. the collapsed version just shows fewer rows of content

barbeau:
@aaronbrethorst Thanks! Sorry, one last question - are the arrivals that appear in the widget from the stop closest to the user's real-time location?

aaronbrethorst:
I don’t recall offhand how they are sorted, but that does seem like the ideal way to sort, imo

barbeau:
Ah, gotcha - so the listing is from multiple stops.

aloksharma060:
Hi am a Btech student and i also want to contribute for this project ?? @barbeau

robertgodfrey:
Thanks! Yeah actually I had initially planned to include the logo somewhere like in the photos below and then hide the logo when the widget was at minimum size (2x1) because there wasn't enough room. 

I ended up removing it to keep the look consistent between the min size widget and the other sizes. Open to adding it back!

<img width="544" height="307" alt="Image" src="https://github.com/user-attachments/assets/bd2b2aa7-fe0c-43f0-b4cf-93a2c21bb926" />

[...]

PR #2325, opened 2026-09-16
Add home screen widget (#1075)

robertgodfrey:
Continuation of https://github.com/OneBusAway/onebusaway-android/pull/1538, rewritten in Kotlin and updated with all the changes that were merged since that PR was originally opened back in March.

### Design decisions

* Widgets are per-stop, rather than like in the iOS version where there is one widget that displays the "today view"
* Widgets can display a maximum of three routes
* All selected routes will always be displayed on a widget - if there are no upcoming ETAs, the route will display `N/A` instead of a time
* ETAs older than -2 minutes are hidden so the widget won't be filled with buses that have already deparated
* ETAs refresh every five minutes via API call, or immediately when the user taps the refresh icon
* Relative time labels ("5 min" → "4 min") update every minute without additional API calls

### Notes

* Widgets have an adaptive layout that shows/hides ETA columns and rows as the widget is resized
* Tapping the widget opens the arrivals list for the configured stop
* I tested on two physical devices, one running Android 16 and one running Android 8

Android 8 relaunches the config screen after the widget is placed via `requestPinAppWidget`. I initially wanted to close this silently, but I couldn't find a reliable way to distinguish between the initial relaunch and a user-initiated edit (available Android 12+).  The current behavior on Android 8 shows the config screen pre-filled and the user taps Save or Back to dismiss it. 

### Main Files

`StopTimesWidget`: AppWidgetProvider managing widget lifecycle, alarm scheduling, and RemoteViews rendering
`WidgetArrivalWorker`: WorkManager worker that fetches arrivals from the OBA API
`StopTimesWidgetConfigActivity`: dialog configuration screen
`StopPickerActivity`: stop selection screen showing starred and recent stops from the local database
`WidgetPrefs`, `WidgetConfig`, `WidgetArrivalSnapshot`: persistence layer using SharedPreferences + Gson

### Screenshots (Updated)

image

### Video (Updated) 

https://birdseedbob.com/recording-kt.mp4 (too big to attach here and I don't feel like editing it, so it's in a CloudFlare bucket)

- [x] Format your changes with `./gradlew spotlessApply`.

## Summary by CodeRabbit

- **New Features**
  - Added a home-screen Stop Times widget showing upcoming arrivals for selected stops and routes.
  - Added setup options for choosing a stop, naming the widget, and selecting up to three routes.
  - Added stop selection from starred and recently viewed stops.
  - Added “Add widget” actions from stop menus and arrival boards.
  - Widgets support manual refresh, automatic updates, resizing, loading states, and schedule-status colors.
  - Added widget reconfiguration and support for preserving selected settings.

- **Tests**
  - Expanded coverage for the “Add widget” stop-menu option.


onebusaway-android/src/main/res/xml-v31/stop_times_widget_info.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:minHeight="40dp"
+        android:updatePeriodMillis="0"
+        android:initialLayout="@layout/stop_times_widget"
+        android:description="@string/widget_description"
+        android:resizeMode="horizontal|vertical"
+        android:widgetCategory="home_screen"
+        android:previewLayout="@layout/stop_times_widget_preview"
+        android:targetCellHeight="1"
+        android:targetCellWidth="2"
+        android:maxResizeHeight="110dp"
```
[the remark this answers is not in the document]
robertgodfrey: 110dp resolves to 2 cells, so all three rows are visible. Verified on device & in emulator.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/StopTimesWidgetConfigActivity.kt
```diff
@@ -0,0 +1,328 @@
...
+                )
+            }
+        }
+    }
+
+    // Built from the stop's own route metadata + the response's reference pool, not the arrivals list
+    // itself — a stop with no arrival in the fetch window (off-hours for a weekend/holiday-only route,
+    // or a stale stop whose service was later restructured) would otherwise offer no routes to pick.
+    private suspend fun loadRoutesForStop(stopId: String): List<RouteOption> {
+        val response = stopArrivalsDataSource.arrivals(stopId, DefaultArrivalsRepository.MINUTES_AFTER_MAX).getOrNull()
```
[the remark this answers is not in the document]
robertgodfrey: Leaning toward not fixing this one. It's real but self-recovering: if the fetch fails, backing out and reopening the config screen is a fresh fetch attempt and just works. Building out a full loading/success/failure+retry state machine for a transient network blip during a few-second, user-initiated setup screen is more infrastructure than the failure mode warrants, especially next to the actual data-integrity fixes already landed in this PR (stale backoff bypass, pin-config race, dropped route on reconfigure, post-reboot ETA math). Leaving this as-is unless others feel it's worth adding.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/WidgetArrivalWorker.kt
```diff
@@ -0,0 +1,178 @@
...
+     */
+    private fun ArrivalData.toSnapshotArrival(): WidgetArrivalSnapshot.Arrival {
+        val scheduled = if (stopSequence != 0) scheduledArrivalTime else scheduledDepartureTime
+        val predictedTime = if (stopSequence != 0) predictedArrivalTime else predictedDepartureTime
+        val hasPrediction = predicted && predictedTime != null
+        val displayTime = if (hasPrediction) predictedTime else scheduled
+        return WidgetArrivalSnapshot.Arrival(
+            scheduledTimeMs = scheduled.epochMs,
+            displayTimeMs = displayTime.epochMs,
+            isPredicted = hasPrediction
```
bmander, line 156: [P2] Preserve cancellation status in the widget snapshot

A response containing `tripStatus.status = CANCELED` still reaches this conversion, but the snapshot retains only the times and prediction flag. The renderer therefore shows an ordinary countdown for a bus that will not run, with no cancellation indication. A JVM probe against `buildSnapshot` confirms that otherwise identical canceled and non-canceled +5 minute trips produce equal snapshots. Please either exclude canceled trips or preserve and render their cancellation state, as the arrivals board does.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/WidgetArrivalWorker.kt
```diff
@@ -0,0 +1,178 @@
...
+
+        val routesWithArrivals = mutableListOf<WidgetArrivalSnapshot.Route>()
+        val routesWithoutArrivals = mutableListOf<WidgetArrivalSnapshot.Route>()
+
+        for ((routeId, shortName) in routeShortNames) {
+            val arrivals = arrivalsByRoute[routeId]
+            if (arrivals != null) {
+                routesWithArrivals += WidgetArrivalSnapshot.Route(
+                    shortName = shortName,
+                    arrivals = arrivals.take(MAX_ARRIVALS_PER_ROUTE).map { it.toSnapshotArrival() }
```
bmander, line 126: [P2] Filter departed arrivals before truncating the route cache

The four-item limit runs before the renderer's stale-arrival filter. With one route's response at -5, -4, -3, -2, +1 and +3 minutes, `buildSnapshot` retains only the four past trips; the widget renders just -2 and then N/A on its next minute tick, even though the API supplied upcoming buses. I reproduced this against the real method in a JVM test. Filter expired trips using the response's server time before limiting, and retain enough upcoming arrivals for the interval between refreshes.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/StopTimesWidgetConfigActivity.kt
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    private fun onSave(stopId: String, stopName: String, widgetName: String, selectedRoutes: Map<String, String>) {
+        if (appWidgetId == AppWidgetManager.INVALID_APPWIDGET_ID) {
+            pinWidgetWithConfig(stopId, stopName, widgetName, selectedRoutes)
+            return
+        }
+
+        WidgetPrefs.saveConfig(this, appWidgetId, WidgetConfig(stopId, stopName, widgetName, selectedRoutes))
+        StopTimesWidget.refreshWidget(this, AppWidgetManager.getInstance(this), appWidgetId)
```
bmander, line 143: [P2] Invalidate cached arrivals when the widget configuration changes

Reconfiguring an existing widget from stop A to stop B replaces only its config; the snapshot is still keyed solely by widget ID and contains no stop/config identity. If the next arrivals fetch fails, `WidgetArrivalWorker` calls `updateArrivalsFromCache(..., allowStaleRefresh = false)`, which renders A's routes and times beneath B's title and binds the tap to B. The relative-time alarm can also render that old snapshot while the new request is pending. Invalidate snapshots when the stop or route selection changes, and prevent an in-flight request for the old config from publishing into the new config.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/WidgetConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param routeShortNames route id -> short name (e.g. "1_100" -> "44"), for the up-to-3 routes the
+ * widget shows.
+ */
+@Serializable
+data class WidgetConfig(
+    val stopId: String,
+    val stopName: String,
+    val widgetName: String,
+    val routeShortNames: Map<String, String> = emptyMap()
```
bmander, line 32: [P2] Keep the widget bound to its configured region

The saved config has no region/deployment identity, while `WidgetArrivalWorker` fetches through `DefaultStopArrivalsDataSource` -> `ObaApiProvider`, which resolves the app's currently selected region on every call. Create a widget in region A, then switch the app to region B: the existing widget now queries B with A's stop and route IDs. It will continually fail for an unknown stop, or show another deployment's arrivals if IDs overlap. Persist the widget's region/deployment and use a matching fetch/open path (or explicitly suspend it on a mismatch) instead of silently rebinding an existing widget.

onebusaway-android/src/main/java/org/onebusaway/android/ui/widget/StopTimesWidget.kt
```diff
@@ -0,0 +1,526 @@
...
+
+            val triggerAtElapsed = ElapsedTime.now() + delay
+            val alarmType = if (action == ACTION_UPDATE_RELATIVE_TIMES) {
+                AlarmManager.ELAPSED_REALTIME
+            } else {
+                AlarmManager.ELAPSED_REALTIME_WAKEUP
+            }
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S && !alarmManager.canScheduleExactAlarms()) {
+                alarmManager.setWindow(alarmType, triggerAtElapsed.ms, 30.seconds.inWholeMilliseconds, pendingIntent)
```
bmander, line 424: [P2] Avoid frozen relative ETAs when exact alarms are unavailable

This fallback is reachable on fresh installs without exact-alarm access. On Android 12+, a 30-second `setWindow` window can be expanded to ten minutes, so neither the one-minute label update nor the five-minute fetch is guaranteed to run near its intended time. A rendered '1 min' and 'Updated just now' can consequently remain visible after the bus has departed; both are static strings, so the host cannot advance them itself. Use a display that remains accurate under delayed delivery (for example absolute arrival/update times in this mode), or another supported countdown mechanism. Android documents both the window expansion and the lack of a default exact-alarm grant: https://developer.android.com/develop/background-work/services/alarms#inexact-during-window

bmander (changes requested):
Reviewed 813a5d4156fa06feb0796fb334f7f03844fcd12b. I found five correctness issues affecting displayed arrivals, widget configuration identity, and refresh timing; details are inline.

Validation: ran `:onebusaway-android:testObaGoogleDebugUnitTest` with `DefaultArrivalsRepositoryTest` (9 passing tests) and two temporary widget reproduction probes. Both probes passed by confirming the defects: canceled/normal trips produce identical snapshots, and departed trips can exhaust the four-arrival cache before upcoming trips are retained. Production code was unchanged. The configuration and region findings are based on tracing the current code; the alarm finding is checked against Android's documentation. Existing PR CI is green. I did not rerun launcher/device UI tests.

robertgodfrey:
Thank you for taking a look @bmander! I will get fixes out sometime this week

